PR #6, opened 2024-02-04
Implement some remaining endpoints, enforce middleware, and add switch to flowsheet

JacksonMeade:
This PR features a number of improvements and new endpoints.

**Change Order Service**
Flowsheet entries can now be 'swapped' with others. As you'll see, this uses Drizzle's [transactions](https://orm.drizzle.team/docs/transactions) interface to make sure that if swapping fails, the play orders are reverted.

**Add Track and Remove Track now update the 'plays' entries in the catalog**
When a new item is added or removed from the flowsheet, entries that match have their play counts incremented/decremented.

**Adds Cognito Middleware to DJs and Library endpoints**
Now that we have confirmed the middleware works via the `testAuth` endpoint, this PR secures the card catalog against outside requests.

**AddGenres, AddFormat, and GetGenres**
Unimplemented endpoints/those marked to-do from the top half of the guide spreadsheet have been added.

**Get Entries by Show**
In place of the 'playlists' endpoints, this is the endpoint that the app will use to retrieve a DJ's show entries.

**Upgrades Drizzle**
In order to use the dynamic query builder while updating the plays of each show, we need Drizzle version 0.29. This deprecates the previous 'InferModel' interfaces, which will require later work.

### DIFF
--- a/src/controllers/flowsheet.controller.ts
+++ b/src/controllers/flowsheet.controller.ts
@@ -1,5 +1,5 @@
 import { Request, RequestHandler } from 'express';
-import { NewFSEntry, FSEntry, Show, ShowDJ } from '../db/schema';
+import { FSEntry, NewFSEntry, Show, ShowDJ } from '../db/schema';
 import * as flowsheet_service from '../services/flowsheet.service';
 
 type QueryParams = {
--- a/src/controllers/flowsheet.controller.ts
+++ b/src/controllers/flowsheet.controller.ts
@@ -282,3 +282,20 @@ export const getOnAir: RequestHandler = async (req, res, next) => {
     next(e);
   }
 };
+
+export const changeOrder: RequestHandler<object, unknown, { entry_id: number; new_order: number }> = async (req, res, next) => {
+  const { entry_id, new_order } = req.body;
+
+  if (entry_id === undefined || new_order === undefined) {
+    res.status(400).json({ message: 'Bad Request: entry_id and new_order are required' });
+  } else {
+    try {
+      const updatedEntry = await flowsheet_service.changeOrder(entry_id, new_order);
+      res.status(200).json(updatedEntry);
+    } catch (e) {
+      console.error('Error: Failed to change order');
+      console.error(e);
+      next(e);
+    }
+  }
+};
--- a/src/services/flowsheet.service.ts
+++ b/src/services/flowsheet.service.ts
@@ -103,11 +103,13 @@ export const addTrack = async (entry: NewFSEntry): Promise<FSEntry> => {
   ));
 
   if (matching_albums.length > 0) {
-    for (let album of matching_albums) {
-      await db.update(library)
-      .set({ last_modified: new Date(), plays: sql`${library.plays} + 1` })
-      .where(eq(library.id, album.id));
-    }
+    await Promise.all(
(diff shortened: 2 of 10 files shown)


src/controllers/flowsheet.controller.ts
```diff
@@ -282,3 +282,20 @@ export const getOnAir: RequestHandler = async (req, res, next) => {
...
+
+  if (entry_id === undefined || new_order === undefined) {
+    res.status(400).json({ message: 'Bad Request: entry_id and new_order are required' });
+  } else {
+    try {
+      const updatedEntry = await flowsheet_service.changeOrder(entry_id, new_order);
+      res.status(200).json(updatedEntry);
+    } catch (e) {
+      console.error('Error: Failed to change order');
+      console.error(e);
```
jakebromberg, line 297: Do we get some sort of performance and error monitoring system with the runtime that we're using? We should think about that if we don't.

src/services/flowsheet.service.ts
```diff
@@ -66,12 +68,80 @@ export const getEntriesByRange = async (startId: number, endId: number) => {
...
+};
+
 export const addTrack = async (entry: NewFSEntry): Promise<FSEntry> => {
+
+  const matching_albums = await db.select()
+  .from(library_artist_view)
+  .where(and(
+    eq(library_artist_view.artist_name, entry.artist_name ?? ""),
+    eq(library_artist_view.album_title, entry.album_title ?? ""),
+    eq(library_artist_view.label, entry.record_label ?? "")
```
jakebromberg, line 102: Is there a runtime cost to leaving in equality checks for empty strings?

src/services/flowsheet.service.ts
```diff
@@ -234,6 +304,11 @@ export const getLatestShow = async (): Promise<Show> => {
   return latest_show;
 };
 
+export const getMostRecentTwoShows = async (): Promise<Show[]> => {
```
jakebromberg, line 307: Any reason not to parameterize this for last n shows?

JacksonMeade: Nope - updated!

src/controllers/flowsheet.controller.ts
JacksonMeade: We do - the AWS/Docker backend reports things to a log file, I believe, but we'll want to confirm with Adrian on whether it is saved and for how long.

AyBruno: I haven't implemented anything for managing persisted logs/ clearing logs, but logs should be stored in a file /var/lib/docker/containers/<container_id>. I'd like to set something more sophisticated up, but haven't had a chance.

src/services/flowsheet.service.ts
JacksonMeade: There is. This was a bad way to do it; I've now updated drizzle to version 0.29 and used their [dynamic query builder interface](https://orm.drizzle.team/docs/dynamic-query-building) to solve this problem. Will probably want to consult as to whether this upgrade is desired; while I'm willing to make the proper adjustments (they're small) it does mean we're using some deprecated, though not non-functional, interfaces.

JacksonMeade:
Closing this PR as it contains too many issues at once. Alternate PRs have been opened in its place.

PR #13, opened 2024-02-08
fix: Allow messages to have greater length

JacksonMeade:
Right now, messages are limited to 64 characters. While this was sensible on the old database, allowing for more characters to expand the DJ name length and maintain both the datetime and the phrase 'started show' or 'ended show' is vital. This updates the schema to make messages up to 250 characters in length, for safety.

### DIFF
--- a/src/db/schema.ts
+++ b/src/db/schema.ts
@@ -160,7 +160,7 @@ export const flowsheet = wxyc_schema.table('flowsheet', {
   record_label: varchar('record_label', { length: 128 }),
   play_order: serial('play_order').notNull(),
   request_flag: boolean('request_flag').default(false).notNull(),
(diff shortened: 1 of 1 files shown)


src/db/schema.ts
```diff
@@ -160,7 +160,7 @@ export const flowsheet = wxyc_schema.table('flowsheet', {
   record_label: varchar('record_label', { length: 128 }),
   play_order: serial('play_order').notNull(),
   request_flag: boolean('request_flag').default(false).notNull(),
-  message: varchar('message', { length: 64 }),
+  message: varchar('message', { length: 250 }),
```
jakebromberg, line 163: Remind me what the message field is for?

jakebromberg (approved):

PR #14, opened 2024-02-08
chore: Update sql query ending show into consistent drizzle notation

JacksonMeade:
For maintainability and consistency, this PR converts the pure SQL query used on the endshow endpoint to drizzle notation. It also adds a 'started show' notification when a DJ starts a new show, with the timestamp included.

### DIFF
--- a/src/services/flowsheet.service.ts
+++ b/src/services/flowsheet.service.ts
@@ -1,21 +1,21 @@
-import { sql, desc, eq, and, lte, gte } from 'drizzle-orm';
+import { and, desc, eq, gte, lte } from 'drizzle-orm';
+import { IFSEntry, UpdateRequestBody } from '../controllers/flowsheet.controller';
 import { db } from '../db/drizzle_client';
 import {
+  DJ,
+  FSEntry,
   NewFSEntry,
-  flowsheet,
-  shows,
   Show,
+  ShowDJ,
+  artists,
+  djs,
+  flowsheet,
+  library,
   // NewShow,
   rotation,
-  library,
-  artists,
-  DJ,
-  djs,
-  FSEntry,
   show_djs,
-  ShowDJ,
+  shows
 } from '../db/schema';
-import { IFSEntry, UpdateRequestBody } from '../controllers/flowsheet.controller';
 
 export const getEntriesByPage = async (offset: number, limit: number) => {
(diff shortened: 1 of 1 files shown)


src/services/flowsheet.service.ts
```diff
@@ -1,21 +1,21 @@
...
-  rotation,
-  library,
+  ShowDJ,
   artists,
-  DJ,
   djs,
-  FSEntry,
+  flowsheet,
+  library,
+  // NewShow,
```
jakebromberg, line 14: Is this intentionally left here? If so, could you add a todo comment for what needs to happen in order to uncomment this?

src/services/flowsheet.service.ts
```diff
@@ -164,26 +171,29 @@ const createJoinNotification = async (id: number): Promise<FSEntry> => {
...
+  await Promise.all(remaining_djs.map(async (dj) => {
+    await db.update(show_djs).set({ active: false }).where(eq(show_djs.dj_id, dj.dj_id));
+    if (dj.dj_id === primary_dj_id) return;
+    await createLeaveNotification(dj.dj_id);
+  }));
+
+  const dj_information = (await db.select().from(djs).where(eq(djs.id, primary_dj_id)).limit(1))[0];
+
+  await db.insert(flowsheet).values({
+    message: `End of Show: DJ ${dj_information.dj_name} left the set at ${new Date().toLocaleString('en-US', { timeZone: 'America/New_York' })}`
```
jakebromberg, line 192: Is there any way to handle the timezone of the client? If I were to do a remote set from Seattle, would it make sense for me to see my own timezone?

JacksonMeade: I can write that into the frontend, but for flowsheet-as-legal-FCC-document purposes, what do we want written to the backend?

jakebromberg: That's up to you. Most systems default to UTC.

AyBruno (changes requested):
Generally I've been opting to use the 'CURRENT_TIMESTAMP' Postgres function/variable/magic. Let's stick to that when inserting dates to stay consistent.

Additionally the sql query that was being used only requires one back and forth trip to the database. I know the ORM is nice, but for more complex queries like this I've been going with prepared sql queries  to offload work and network traffic to the db.

PR #17, opened 2024-05-29
Fix hanging requests cognito

AyBruno:
Cognito middleware usesd to passed like so:
```js
my_route('/route', cognitoMiddleware, (req, res) => { /*logic */ })
```
But in order to allow multiple levels (mgmt only) endpoints I had to wrap the middleware function that is eventually returned in another function. So it should now be called like this:
```js
my_route('/route', cognitoMiddleware(), (req, res) => { /*logic */ })
```
or
```js
my_route('/route', cognitoMiddleware('station-management'), (req, res) => { /*logic */ })
```
Having routes with the middleware, but without the parens needed to call the function just stops the middleware pipeline in its tracks causing the program to hang. I've added a http request timeout of 5s as well to stop future requests from hanging indefinitely and taking up resources.

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -52,7 +52,9 @@ jwtVerifier
     console.error(`Failed to hydrate JWT verifier : ${e}`);
   })
   .then(() => {
-    app.listen(port, () => {
+    const server = app.listen(port, () => {
       console.log(`listening on port: ${port}!`);
     });
+
+    server.setTimeout(5000);
   });
--- a/src/routes/library.route.ts
+++ b/src/routes/library.route.ts
@@ -4,14 +4,13 @@ import { cognitoMiddleware } from '../middleware/cognito.auth';
 
 export const library_route = Router();
 
-//secure: discuss, should we lock down the catalog?
-library_route.get('/', cognitoMiddleware, libraryController.searchForAlbum);
+library_route.get('/', cognitoMiddleware(), libraryController.searchForAlbum);
 
-//secure: mgmt
+// secure: mgmt
 library_route.post('/', cognitoMiddleware('station-management'), libraryController.addAlbum);
 
 //secure: mgmt
-library_route.get('/rotation', cognitoMiddleware, libraryController.getRotation);
+library_route.get('/rotation', cognitoMiddleware(), libraryController.getRotation);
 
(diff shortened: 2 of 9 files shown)


src/app.ts
```diff
@@ -52,7 +52,9 @@ jwtVerifier
     console.error(`Failed to hydrate JWT verifier : ${e}`);
   })
   .then(() => {
-    app.listen(port, () => {
+    const server = app.listen(port, () => {
       console.log(`listening on port: ${port}!`);
     });
+
+    server.setTimeout(5000);
```
dvdokkum, line 59: I would maybe consider bumping this to 10 just to give ourselves some more breathing room in high load scenarios, but I don't know the system as well so I trust your gut here.

src/routes/library.route.ts
```diff
@@ -4,14 +4,13 @@ import { cognitoMiddleware } from '../middleware/cognito.auth';
...
 
-//secure: discuss, should we lock down the catalog?
-library_route.get('/', cognitoMiddleware, libraryController.searchForAlbum);
+library_route.get('/', cognitoMiddleware(), libraryController.searchForAlbum);
 
-//secure: mgmt
+// secure: mgmt
 library_route.post('/', cognitoMiddleware('station-management'), libraryController.addAlbum);
 
 //secure: mgmt
```
dvdokkum, line 12: just noting here that the comment says secure but it doesn't include the `station-management` arg

dvdokkum (approved):
just noting that there's a bunch of stuff in here unrelated to the cognito fix, just in case that isn't intentional.

src/routes/library.route.ts
dvdokkum: in case it isn't clear, i'm talking about the GET for `/rotation` which i think correctly doesn't include the arg, but the comment indicates otherwise.

AyBruno:
Yeah I accidentally branched off of the previous fix/feature branch for each of these branches I submitted pull requests for. It should be sufficient to just merge the last one if they all look good.

dvdokkum:
@AyBruno they all look good to me. all approved, do what you wanna do!

src/routes/library.route.ts
AyBruno: Yes, it shouldn't be restricted to station management.

PR #18, opened 2024-05-29
Fix flowsheet join end

AyBruno:
The flowsheet /end and /join endpoints were not updating their flowsheet entries with the current show_id. I updated the createJoinNotification and createLeaveNotification functions to account for that as well as the flowsheet insertion statements that account for the start and end of a show.


dvdokkum (approved):
just noting that there's a bunch of stuff in here covered by other PRs as well, in case that isn't intentional. thank you for doing this one, it makes my script much easier to have the show_id available in the flowsheet in addition to the join/leave respone

PR #20, opened 2024-05-30
swapped less than sign to greater than sign smh

AyBruno:
Was only getting rotation entries that had already expired. 🤦


dvdokkum (approved):
lgtm

PR #21, opened 2024-06-06
Fixes for the postgres error: column play_freq does not exist

AyBruno:
Drizzle orm doesn't handle column name aliasing in views correctly, so the column can't be named 'rotation' instead it must be named 'play_freq'. I updated this and pushed the migration to the prod db.

I have also commented out logic that was making use of the view for now just to unblock the rest of the team as the $dynamic() queries do not work the we thought they did when we wrote the album plays feature.


JacksonMeade (approved):
Leaving an approving review because this appears to solve the issue if removing the view is assumed to do so, and its blocking. Let me know if you’d like support in developing that add to album backend code over slack.

dvdokkum (approved):
ship it

jakebromberg:
+1

AyBruno:
Sounds good! The view still exists, I just renamed the column to not use an alias. I'm not using the view within the flowsheet service though because of that separate $dynamic() issue.

PR #25, opened 2024-08-26
Feat test automation

AyBruno:
### Testing automation env + flowsheet test suite

This pull request contains the configuration and implementation of a `jest`+`supertest` testing env for the backend service.
Built on that base is a first pass at full flowsheet integration tests and a few helper utilities for test setup and cleanup. Also included are configuration flags for `.env` to bypass cognito auth for local testing as well as a method to include cognito in the test flow when desired.

I've also added a library called `jest-html-reporters` that will generate a nice ui to display the results of any test suites added to `tests/specs` directory. It is configured to output to the `tests/report` directory.

Next steps will be to automate the triggering of these tests inside of our CI and publish the report to github pages.


dvdokkum (approved):
lgtm

PR #26, opened 2024-08-26
Added DJ not in show check & DJ info cleanup

AyBruno:
Added logic to the flowsheet controller to ensure the requesting dj is a member of the show before entering the leaveShow() logic. This will be preceded by the new flowsheet middleware from another pull request.

Also included in this pull request is some data cleanup for the getDJList() controller. It filters out any identifying info about the DJ such as their real names, emails, etc as to maintain their privacy.


dvdokkum (approved):
lgtm

PR #39, opened 2025-02-03
Change flowsheet default limit to 30, add previous_shows filter

dvdokkum:
This changes the default number of returned items in the flowsheet from 5 to 30.

I also added a new param `?previous_shows` which takes a number `2` and returns all entries from the current show + the last 2 shows.

Didn't bother testing locally... lemme know if you see anything amiss.

### DIFF
--- a/tests/specs/flowsheet.spec.js
+++ b/tests/specs/flowsheet.spec.js
@@ -2,6 +2,7 @@ require('dotenv').config({ path: '../../.env' });
 const request = require('supertest')(`${process.env.TEST_HOST}:${process.env.PORT}`);
 const { after } = require('node:test');
 const fls_util = require('../utils/flowsheet_util');
+const { db } = require('../../src/db/drizzle_client');
 
 /*
  * Start Show (Primary dj hits /flowsheet/join)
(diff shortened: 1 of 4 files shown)


tests/specs/flowsheet.spec.js
```diff
@@ -2,6 +2,7 @@ require('dotenv').config({ path: '../../.env' });
 const request = require('supertest')(`${process.env.TEST_HOST}:${process.env.PORT}`);
 const { after } = require('node:test');
 const fls_util = require('../utils/flowsheet_util');
+const { db } = require('../../src/db/drizzle_client');
```
AyBruno, line 5: db should not be needed here. Causes an error as our tests are js using supertest and not directly hooked into the app. Type definition in drizzle_client file causes parsing error.

dvdokkum: ahh thanks for catching that! fixed.

dvdokkum:
@AyBruno thanks for improving the work! good to merge from your perspective? if so, feel free to approve and i'll take it from there

AyBruno (approved):

PR #52, opened 2025-06-19
Replace tsc with tsup and add openapi3.0 docs

AyBruno:
Replace tsc with tsup for smaller bundles and enhanced build times
 * Alterations to tsconfig to support this
 * Add .js on import paths
 * use json jest config rather than ts
 * update npm scripts

Add Openapi 3.0 docs defined by in src/app.yaml
* New endpoint /api-docs
* Updated some types

Minor: added use of cors library rather than manually setting headers


jakebromberg:
Could you add some notes to the commit message about why we're doing this?

PR #60, opened 2025-08-07
Feature/server sent events

AyBruno:
Adds logic for managing and using sse to push real time event to clients. Integrates with updates to auth middleware in order to filter out events that user's may request but aren't authorized for.

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -46,8 +46,8 @@ app.use('/schedule', schedule_route);
 app.use(
   '/events',
   (req, res, next) => {
-    // no global timeout on these long lived connection.
-    // SSE logic handles this itself
+    // no global timeout on these long lived connections
+    // SSE logic handles timeouts itself
     res.setTimeout(0);
     next();
   },
--- a/src/app.ts
+++ b/src/app.ts
@@ -9,6 +9,7 @@ import { dj_route } from './routes/djs.route.js';
 import { flowsheet_route } from './routes/flowsheet.route.js';
 import { library_route } from './routes/library.route.js';
 import { schedule_route } from './routes/schedule.route.js';
+import { events_route } from './routes/events.route.js';
 import { jwtVerifier, cognitoMiddleware } from './middleware/cognito.auth.js';
 import { showMemberMiddleware } from './middleware/checkShowMember.js';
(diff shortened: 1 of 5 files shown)


src/app.ts
```diff
@@ -42,6 +43,17 @@ app.use('/djs', dj_route);
 
 app.use('/schedule', schedule_route);
 
+app.use(
+  '/events',
+  (req, res, next) => {
+    // no global timeout on these long lived connection.
```
jakebromberg, line 49: Nit: connections

src/middleware/cognito.auth.ts
```diff
@@ -56,7 +56,14 @@ function NewJwtVerifier(testing: boolean): JwtVerifier {
 
 export const jwtVerifier = NewJwtVerifier(process.env.AUTH_BYPASS === 'true');
 
-export const cognitoMiddleware = (permissionsLevel: string | null = null) => {
+export const Roles = {
+  mgmt: 'station-management',
+  musicDirector: 'music-management',
+  stationMgr: 'station-manager',
+  dj: 'dj',
```
jakebromberg, line 63: Do we need an admin role?

src/utils/serverEvents.ts
```diff
@@ -0,0 +1,241 @@
+import { Response } from 'express';
+
+export const Topics = {
+  test: 'test-topic',
+  primaryDj: 'prim-dj-topic',
+  showDj: 'show-dj-topic',
+  liveFs: 'live-fs-topic',
+} as const;
```
jakebromberg, line 8: Could you add comments to differentiate these values?

src/middleware/cognito.auth.ts
AyBruno: I don't think we necessarily need an admin role. We can just add ourselves to the group that we need permissions for.

AyBruno: For example if we ever need access to the music director endpoints, we have access to aws and can add ourselves to the music-management group. From then on we have all the permissions of the md.

jakebromberg: Got it. Sounds good.

PR #66, opened 2025-09-15
Task: Enhance error handling and handle sse errors

AyBruno:
- Making use of custom error handler middleware and extended WxycError class to allow us to pass client error (400-499) as throwable errors rather than handling them directly in controllers. There will be a separate effort to refactor to this pattern.
- Handling server sent events (sse) errors properly to avoid crashes when invalid requests are passed.


AyBruno:
> Could you provide a more detailed summary of the changes in this commit? How is the error handling enhanced? What is an SSE error?

@jakebromberg I've updated PR description

jakebromberg (approved):

PR #116, opened 2026-01-16
Add trailing whitespace check for .env files

jakebromberg:
## Summary
- Adds a cross-platform Node.js script (`scripts/lint-env.js`) that checks `.env` and `.env.example` for trailing whitespace
- Adds `npm run lint:env` script for local development
- Adds lint step to CI workflow (runs before tests)
- Adds `.env.example` template for new developers

## Test plan
- [x] Verified script catches trailing whitespace (shows line numbers and visualizes with `·`)
- [x] Verified script passes when no trailing whitespace exists
- [x] Verified script works on macOS (Darwin)
- [x] CI workflow runs lint step before tests

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,0 +1,46 @@
+### Backend Service Port
+PORT=8080
+CI_PORT=8081
+AUTH_PORT=8082
+
+### Better Auth Configuration
+BETTER_AUTH_URL=http://localhost:8082/auth
+BETTER_AUTH_JWKS_URL=http://localhost:8082/auth/jwks
+BETTER_AUTH_ISSUER=http://localhost:8082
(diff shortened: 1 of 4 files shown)


AyBruno (commented):
@jakebromberg We may also want to add a step to the deploy workflow where we run this linting script on our ec2 instance otherwise issues in that env would still fall through the cracks.

JacksonMeade (approved):

jakebromberg:
> @jakebromberg We may also want to add a step to the deploy workflow where we run this linting script on our ec2 instance otherwise issues in that env would still fall through the cracks.

@AyBruno 
You'll have to go over with me how that would work. I can follow up with that in an additional PR.

AyBruno (approved):

PR #126, opened 2026-01-20
perf: optimize test infrastructure for faster CI

jakebromberg:
## Summary
- Add Jest performance settings (maxWorkers 50%, bail on first failure, forceExit)
- Reduce health check timeout from 60s to 15s for faster failure detection
- Add shared database connection pool for tests (eliminates per-test connection overhead)
- Add Docker layer caching to CI workflow
- Add npm caching for faster dependency installation
- Update GitHub Actions to v4

## Test plan
- [ ] CI pipeline passes
- [ ] Compare CI run time with previous runs

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -32,6 +32,7 @@ jobs:
       db-init: ${{ steps.changes.outputs.db-init }}
       docs-only: ${{ steps.changes.outputs.docs-only }}
       src: ${{ steps.changes.outputs.src }}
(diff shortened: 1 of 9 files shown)


AyBruno (changes requested):
Refer here for an example of how we are detecting what needs to be built in the deploy workflow:
https://github.com/WXYC/Backend-Service/blob/main/.github/workflows/deploy-base.yml#L78

This does not take into account if the `shared` package has changed and we're working to resolve that in the deploy workflow

PR #145, opened 2026-02-06
fix: always use dj.wxyc.org for email links

JacksonMeade:
**This is absolutely critical for getting login flow working**

Rewrites email verification and password reset URLs to use the frontend domain instead of the backend API domain, while preserving path and query parameters.

#### What Changed
- Added `rewriteUrlForFrontend` helper that replaces only the host and protocol, leaving path and query params intact
- Applied URL rewriting to verification email links
- Applied URL rewriting to password reset email links (via `buildResetUrl`)

#### Why
Email links now point to `dj.wxyc.org` instead of `api.wxyc.org`, so users land on the frontend. The `callbackURL` query parameter remains a relative path (e.g., `/onboarding`) that the frontend resolves correctly. This removes the need for frontend URL sanitization helpers.

### DIFF
--- a/shared/authentication/src/auth.definition.ts
+++ b/shared/authentication/src/auth.definition.ts
@@ -24,17 +24,38 @@ import { eq, sql } from 'drizzle-orm';
 import { WXYCRoles } from './auth.roles';
 import { sendResetPasswordEmail, sendVerificationEmailMessage } from './email';
 
-const buildResetUrl = (url: string, redirectTo?: string) => {
-  if (!redirectTo) {
+/**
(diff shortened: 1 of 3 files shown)


JacksonMeade:
> I'm wondering if you would mind implementing tests to cover this?

I've added a unit test for the behavior of the rewrite function, which is the most 'breakable' part of what I've done here. I will not be writing end to end testing for this until we have new users; this is a zero-level kind of feature that should have been in the original implementation.

jakebromberg (approved):

PR #144, opened 2026-02-02
feat: add automated migrations to deployment pipeline

jakebromberg:
## Summary

- Add database indexes for performance optimization on frequently-queried foreign key columns
- Automate database migrations as part of the deployment pipeline
- Fix orphaned migration file that was not tracked in Drizzle's journal
- **NEW: Add comprehensive migration testing and analysis tools**

### Migration Testing Tools (New)

Added infrastructure to catch migration issues before production deployment:

| Tool | Purpose |
|------|---------|
| `lint-migrations.js` | Static analysis for dangerous patterns |
| `generate-rollback.mjs` | Auto-generate rollback SQL scripts |
| `estimate-migration.mjs` | Estimate migration duration based on table sizes |
| `create-schema-snapshot.mjs` | Export anonymized production schema |
| `load-schema-snapshot.mjs` | Load snapshot into test database |
| `test-migrations-snapshot.mjs` | Test migrations against production-like data |

**Lint Rules:**
- `concurrent-index`: Warns when CREATE INDEX misses CONCURRENTLY
- `not-null-no-default`: Blocks ADD COLUMN NOT NULL without DEFAULT
- `missing-if-not-exists`: Warns about missing idempotency guards
- `dangerous-ops`: Blocks TRUNCATE, DROP TABLE without IF EXISTS

**CI Integration:**
- Migration linting runs on PRs that modify `shared/`
- Weekly snapshot testing workflow (Sundays 6 AM)

**npm Scripts:**
```
npm run lint:migrations          # Lint all migrations
npm run lint:migrations:changed  # Lint changed only (for CI)
npm run rollback:generate        # Generate rollback templates
npm run migration:estimate       # Estimate durations
npm run snapshot:test            # Test against snapshot
```

### Database Indexes

Added B-tree indexes to improve query performance:

| Table | Index | Columns |
|-------|-------|---------|
| `flowsheet` | `flowsheet_show_id_idx` | `show_id` |
| `flowsheet` | `flowsheet_album_id_idx` | `album_id` |
| `flowsheet` | `flowsheet_rotation_id_idx` | `rotation_id` |
| `show_djs` | `show_djs_show_id_dj_id_idx` | `(show_id, dj_id)` |
| `show_djs` | `show_djs_dj_id_idx` | `dj_id` |
| `bins` | `bins_dj_id_idx` | `dj_id` |
| `bins` | `bins_album_id_idx` | `album_id` |

### Deployment Pipeline

Added migration jobs to the deployment workflow:

```mermaid
[...]

PR #149, opened 2026-02-12
feat: add admin role with roster permissions

jakebromberg:
## Summary
- Add admin role with roster management permissions

Split from #144 for easier review.

## Test plan
- [ ] Verify admin role grants roster permissions
- [ ] Verify non-admin users cannot access roster endpoints


jakebromberg:
Closing: the admin role is identical to stationManager, and the roster resource isn't used anywhere across any branch. This should ship alongside the feature that actually introduces roster endpoints.

PR #181, opened 2026-02-21
fix: correct migrations folder path in migration_script.mjs

jakebromberg:
## Summary
- Fixed `migration_script.mjs` which referenced `src/db/migrations` — a path that doesn't exist
- Updated to `shared/database/src/migrations` to match the path used in `drizzle.config.ts`
- Added unit test that verifies the migrations folder path is correct and resolves to an existing directory

## Test plan
- [x] New unit test `tests/unit/database/migration-script-path.test.ts` passes
- [x] Test verified to fail before the fix and pass after

### DIFF
--- a/tests/unit/database/migration-script-path.test.ts
+++ b/tests/unit/database/migration-script-path.test.ts
@@ -1,10 +1,7 @@
 import * as fs from 'fs';
 import * as path from 'path';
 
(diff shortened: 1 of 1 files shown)


AyBruno (changes requested):
This file is no longer used and is deleted in `main` thanks to improvements the `drizzle` team made to `drizzle-kit`. This PR can be closed

PR #186, opened 2026-02-21
fix: use singleton PostHog client instead of per-request instantiation

jakebromberg:
## Summary

- **Bug**: `createBackendMirrorMiddleware` instantiated a new `PostHog` client (with its own connection pool and flush interval) inside the `res.once('finish', ...)` callback on every request, then called `shutdown()` on each one.
- **Fix**: Replaced with a module-level lazy singleton via `getPostHogClient()`. The client is created once on first use and reused for all subsequent requests. Removed the per-request `shutdown()` call since the singleton persists for the process lifetime.
- **Test**: Added `mirror.posthog.test.ts` that mocks `posthog-node`, fires two request finish events, and asserts the `PostHog` constructor is called exactly once.

## Test plan

- [x] New unit test fails before fix (constructor called 2x)
- [x] New unit test passes after fix (constructor called 1x)
- [x] All existing legacy mirror tests still pass (57/57)

### DIFF
--- a/apps/backend/middleware/legacy/mirror.middleware.ts
+++ b/apps/backend/middleware/legacy/mirror.middleware.ts
@@ -1,17 +1,6 @@
 import { NextFunction, Request, Response } from 'express';
 import { MirrorCommandQueue } from './commandqueue.mirror';
-
-import { PostHog } from 'posthog-node';
-
-let postHogClient: PostHog | null = null;
-function getPostHogClient(): PostHog {
-  if (!postHogClient) {
-    postHogClient = new PostHog(process.env.POSTHOG_API_KEY ?? '', {
-      host: 'https://us.i.posthog.com',
-    });
-  }
-  return postHogClient;
-}
+import { getPostHogClient } from '../../utils/posthog.js';
 
 export const createBackendMirrorMiddleware =
   <T>(createCommand: (req: Request, data: T) => Promise<string[]>) =>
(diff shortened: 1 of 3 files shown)


apps/backend/middleware/legacy/mirror.middleware.ts
```diff
@@ -3,6 +3,16 @@ import { MirrorCommandQueue } from './commandqueue.mirror';
 
 import { PostHog } from 'posthog-node';
 
+let postHogClient: PostHog | null = null;
```
AyBruno: Rather than keep the posHogClient in this mirror.middleware.ts file, I think it would be good future proofing to move it into its own file in the utils directory as a common module level singleton. This way, in the future as we implement features we have a dedicated place to reach out to posthog for feature flags.

apps/backend/middleware/legacy/mirror.middleware.ts
```diff
@@ -12,9 +11,7 @@ export const createBackendMirrorMiddleware =
     res.once('finish', () => {
       void (async () => {
         try {
-          const postHogClient = new PostHog(process.env.POSTHOG_API_KEY ?? '', {
-            host: 'https://us.i.posthog.com',
-          });
+          const client = getPostHogClient();
```
AyBruno: This can be inlined `getPostHogClient().isFeatureEnabled()`

PR #187, opened 2026-02-21
fix: apply showMemberMiddleware to flowsheet write routes

jakebromberg:
## Summary
- **Bug**: `showMemberMiddleware` was imported in `apps/backend/app.ts` but never applied to any route, allowing any authenticated DJ to modify flowsheet entries even without joining the active show.
- **Fix**: Applied `showMemberMiddleware` after `requirePermissions` on the POST, PATCH, and DELETE flowsheet routes in `flowsheet.route.ts`. Removed the unused import from `app.ts`.
- **Hardening**: Added try/catch error handling to the middleware (previously had none) and simplified the DJ lookup from `.filter().length` to `.some()`.

## Test plan
- [x] Unit tests for `showMemberMiddleware` covering:
  - DJ not in show → returns 400
  - DJ in show → calls `next()`
  - No DJs in show → returns 400
  - DB error → returns 500 (new error handling)
- [x] Full unit test suite passes (124 tests, 9 suites)

Fixes #10

### DIFF
--- a/apps/backend/app.ts
+++ b/apps/backend/app.ts
@@ -12,7 +12,6 @@ import { library_route } from './routes/library.route.js';
 import { schedule_route } from './routes/schedule.route.js';
 import { events_route } from './routes/events.route.js';
 import { request_line_route } from './routes/requestLine.route.js';
-import { showMemberMiddleware } from './middleware/checkShowMember.js';
 import { activeShow } from './middleware/checkActiveShow.js';
 import errorHandler from './middleware/errorHandler.js';
 import { requestIdMiddleware } from './middleware/requestId.js';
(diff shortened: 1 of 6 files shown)


AyBruno (changes requested):
Looks good! This actually wasn't a bug that we weren't using this yet. I wrote it around the time that we were deciding to leave aws cognito and as I has implemented it for our cognito auth specifically, I never hooked it up. Now that we've settled on better-auth, this is a good improvement to clear out some code duplication.

That being said! This PR should include removal of showMember checks within the flowsheet controllers that are preexisting. 
Example: https://github.com/WXYC/Backend-Service/blob/main/apps/backend/controllers/flowsheet.controller.ts#L357-L361

AyBruno (approved):

PR #189, opened 2026-02-21
fix: prevent SSH timeout stacking in MirrorSQL

jakebromberg:
## Summary
- Each `sshInstance()` call in `MirrorSQL` created a new 5-minute `setTimeout` without clearing the previous one, causing dangling timeouts to stack on rapid repeated calls.
- Stores the timeout handle as a static class property and calls `clearTimeout` on the existing handle before setting a new one.
- Adds unit tests verifying only one timeout is active after multiple calls, and that superseded timeouts don't trigger premature disposal.

## Test plan
- [x] New test: `tests/unit/middleware/legacy/ssh-timeout.test.ts` — confirms `clearTimeout` is called on prior handles and that only the latest timeout fires `dispose()`.
- [x] Tests fail before fix, pass after fix.
- [x] Existing unit tests still pass (`npm run test:unit`).

Fixes #36

### DIFF
--- a/tests/unit/middleware/legacy/ssh-timeout.test.ts
+++ b/tests/unit/middleware/legacy/ssh-timeout.test.ts
@@ -7,7 +7,7 @@ jest.mock('node-ssh', () => {
   return { NodeSSH: jest.fn(() => mockSSH) };
 });
 
-import { MirrorSQL } from '../../../../apps/backend/middleware/legacy/sql.mirror';
+import { MirrorSQL } from '../../../../shared/database/src/legacy/sql.mirror';
 
 describe('MirrorSQL SSH timeout stacking', () => {
   beforeEach(() => {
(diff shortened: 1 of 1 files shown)


apps/backend/middleware/legacy/sql.mirror.ts
```diff
@@ -3,6 +3,7 @@ import { Config, NodeSSH } from 'node-ssh';
 export class MirrorSQL {
   private static _instance: MirrorSQL | null = null;
   private static _ssh: NodeSSH | null = null;
+  private static _timeoutHandle: ReturnType<typeof setTimeout> | null = null;
```
AyBruno: This file has been moved here and has similar changes: https://github.com/WXYC/Backend-Service/blob/main/shared/database/src/legacy/sql.mirror.ts

This PR should be converted to a cleanup to remove this file from living in the `backend` service.

PR #196, opened 2026-02-21
fix: add withTimezone to all wxyc_schema timestamp columns

jakebromberg:
## Summary
- All 14 `timestamp()` columns in `wxyc_schema` tables were missing `{ withTimezone: true }`, making them ambiguous during DST transitions. Auth tables (public schema) already used `withTimezone`. This aligns the two.
- Adds a unit test (`tests/unit/database/schema.timestamps.test.ts`) that reads the schema source and asserts every `timestamp()` call includes `withTimezone: true`, preventing future regressions.
- No migration generated — that will be a separate step.

## Affected tables
`shift_covers`, `artists`, `library`, `flowsheet`, `genres`, `reviews`, `shows`, `specialty_shows`, `album_metadata`, `artist_metadata`

## Test plan
- [x] Unit test fails before fix (14 columns detected)
- [x] Unit test passes after fix
- [ ] Verify existing integration tests still pass in CI
- [ ] Generate and review migration separately before deploying

Closes #39

> **Note:** This fix is consolidated in #199 (coordinated schema migration batch). Consider closing in favor of #199.

### DIFF
--- a/shared/database/src/schema.ts
+++ b/shared/database/src/schema.ts
@@ -189,7 +189,7 @@ export const shift_covers = wxyc_schema.table('shift_covers', {
   schedule_id: serial('schedule_id')
     .references(() => schedule.id)
     .notNull(),
-  shift_timestamp: timestamp('shift_timestamp').notNull(), //Timestamp to expire cover requests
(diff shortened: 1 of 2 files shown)


AyBruno (approved):
This is good for safety I suppose, but any inserts/updates we perform on timestamp fields already default to UTC and shouldn't technically require this.

PR #234, opened 2026-03-11
fix: rename album_title query param to album_name to match API spec

### LINKED ISSUE #233: GET /library search ignores album_name query parameter
The \`GET /library\` search endpoint accepts \`album_title\` as a query parameter, but the OpenAPI spec (\`CatalogSearchParams\` in \`api.yaml\`) and the dj-site frontend both send \`album_name\`. This means album searches are silently ignored — the backend only matches against \`artist_name\`, so searching for an album title like \"Confield\" returns artist-name fuzzy matches (Scofield, Sinfield) instead of the album itself.

### Root cause

\`AlbumQueryParams\` in \`library.controller.ts\` defines the field as \`album_title\`, while \`CatalogSearchParams\` in \`api.yaml\` defines it as \`album_name\`. The frontend sends \`album_name\`, which Express ignores since it doesn't match any declared query param.

### Fix

Rename \`album_title\` to \`album_name\` in \`AlbumQueryParams\` and update all references in the controller, integration tests, and test utilities to match the spec.

jakebromberg:
## Summary

- Rename `album_title` to `album_name` in `AlbumQueryParams` for `GET /library` to match `CatalogSearchParams` in `api.yaml` and the dj-site frontend
- Update integration tests and test utility to use the corrected parameter name

The frontend sends `album_name` but the backend expected `album_title`, so album searches were silently ignored — only artist name matching worked.

Closes #233

## Test plan

- [ ] `GET /library?album_name=Confield` returns Autechre's "Confield"
- [ ] `GET /library?artist_name=Built+to+Spill&album_name=Keep+it` returns matching results
- [ ] `GET /library` with no params still returns 400
- [ ] Integration tests pass (`searches by album title`, `searches by both artist and album`)

### DIFF
--- a/apps/backend/controllers/library.controller.ts
+++ b/apps/backend/controllers/library.controller.ts
@@ -79,7 +79,7 @@ export const addAlbum: RequestHandler = async (req: Request<object, object, NewA
 
 type AlbumQueryParams = {
   artist_name?: string;
-  album_title?: string;
+  album_name?: string;
   code_letters?: string;
   code_artist_number?: string;
   code_number?: number;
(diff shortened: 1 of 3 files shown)


AyBruno (changes requested):
Why not update the api spec to match what the API expects?

P.S. Use this api spec for reference that predates the one in shared repo: https://github.com/WXYC/Backend-Service/blob/main/apps/backend/app.yaml#L28

PR #306, opened 2026-04-05
Fix playlist proxy to use flowsheet table instead of deleted album_metadata

jakebromberg:
## Summary

- Rewrite playlist proxy enrichment to query the `flowsheet` table's inline `artwork_url` instead of the deleted `album_metadata` table
- Uses the same normalized key-based matching pattern but queries flowsheet directly
- Required because #305 removes the `album_metadata` table

Depends on #305 (must merge first).

## Test plan

- [ ] `npm run test:unit` — playlist proxy tests pass (22 tests, verified locally)
- [ ] CI passes after #305 merges


jakebromberg:
Changes folded into #305 — the playlist proxy rewrite needs to ship with the metadata table removal to avoid breaking the build.

PR #376, opened 2026-04-22
fix: consolidate auth middleware to JWT-only verification

### LINKED ISSUE #374: fix: consolidate auth middleware to JWT-only verification
## Context

The backend has two auth middleware with different verification methods:

- **`requirePermissions`** (`@wxyc/authentication`) verifies JWTs via `jose.jwtVerify()` + JWKS. Works correctly.
- **`requireAnonymousAuth`** (`apps/backend/middleware/`) uses `auth.api.getSession()` via better-auth's bearer plugin. The bearer plugin only understands HMAC-signed session tokens, not JWTs, so any client sending a JWT gets silently rejected (401).

The dj-site sends JWTs to all backend APIs via `backendBaseQuery`. Routes behind `requirePermissions` work; routes behind `requireAnonymousAuth` (proxy/metadata, library/search, config/secrets, requestLine) return 401.

## Root cause

better-auth's bearer plugin converts Bearer tokens to session cookies by verifying an HMAC signature. Session tokens have 2 dot-separated parts (`token.hmac`), so the HMAC check passes. JWTs have 3 parts (`header.payload.signature`), so the HMAC check fails silently and `getSession()` returns null.

## Changes

- Make `requirePermissions` support empty permissions: `requirePermissions({})` verifies the JWT without requiring a role or checking permissions, suitable for anonymous/public endpoints
- Add ban check to `requirePermissions` using the `banned` field already present in JWT payloads
- Replace all `requireAnonymousAuth` usages with `requirePermissions({})`
- Extract activity tracking into a composable `trackActivity` middleware
- Delete `requireAnonymousAuth` and its tests
- Update rate limiting and request line controller to use `req.auth` instead of `req.user`

## Affected routes

| Route | Before | After |
|-------|--------|-------|
| `proxy/*` | `requireAnonymousAuth` (broken) | `requirePermissions({})` + `trackActivity` |
| `library/search` | `requireAnonymousAuth` (broken) | `requirePermissions({})` + `trackActivity` |
| `config/secrets` | `requireAnonymousAuth` (broken) | `requirePermissions({})` + `trackActivity` |
| `requestLine` POST/parse | `requireAnonymousAuth` (broken) | `requirePermissions({})` + `trackActivity` |

## Deployment note

The iOS app (`wxyc-ios-64`) currently sends session tokens (from anonymous sign-in) as Bearer tokens. These pass the bearer plugin's HMAC check but will fail `jose.jwtVerify()`. The iOS app needs to be updated to exchange session tokens for JWTs via `/auth/token` before this PR is deployed. See WXYC/wxyc-ios-64 for the companion PR.

## Test plan

- [x] All 793 unit tests pass (72 suites)
- [x] No lint errors
- [ ] Verify proxy/metadata, library/search, config/secrets, requestLine routes accept JWT Bearer tokens in staging
- [ ] Verify banned user JWT returns 403
- [ ] Verify `requirePermissions({ catalog: ['read'] })` still requires role

jakebromberg:
## Summary

- Consolidate two diverging auth middleware (`requireAnonymousAuth` + `requirePermissions`) into one (`requirePermissions`) with JWT-only verification
- Add ban checking to `requirePermissions` (reads `banned` from JWT payload)
- Support `requirePermissions({})` for routes that need JWT verification without role/permission checks
- Extract activity tracking into composable `trackActivity` middleware
- Delete `requireAnonymousAuth` and its tests

Closes #374

## Test plan

- [x] All 793 unit tests pass (72 suites)
- [x] No lint errors
- [ ] Verify proxy/metadata, library/search, config/secrets, requestLine routes accept JWT Bearer tokens in staging
- [ ] Verify banned user JWT returns 403
- [ ] Deploy iOS update (session token → JWT exchange) before merging


jakebromberg:
## CI fix: `AUTH_BYPASS` skips token presence checks

The 4 failing tests (`should return 401 without Authorization header`, `malformed`, `empty Bearer`, `invalid Bearer`) all get 200 because the `AUTH_BYPASS` block in `requirePermissions` calls `next()` unconditionally — even with no auth header. The old `requireAnonymousAuth` had no bypass path, so it always validated via `auth.api.getSession()`.

The bypass should skip *signature verification*, not *requiring a token entirely*. Moving the header/token presence checks above the early return fixes it:

```typescript
if (process.env.AUTH_BYPASS === 'true' && process.env.NODE_ENV !== 'production') {
  const authHeader = req.headers.authorization;
  if (!authHeader) {
    return res.status(401).json({ error: 'Unauthorized: Missing Authorization header.' });
  }
  const token = authHeader.startsWith('Bearer ') ? authHeader.slice(7).trim() : authHeader.trim();
  if (!token) {
    return res.status(401).json({ error: 'Unauthorized: Missing token in Authorization header.' });
  }
  try {
    const payload = decodeJwt(token) as WXYCAuthJwtPayload;
    const userId = payload.id || payload.sub;
    if (userId) {
      req.auth = { ...payload, id: userId } as WXYCAuthJwtPayload;
    }
  } catch {
    req.auth = { id: token } as WXYCAuthJwtPayload;
  }
  return next();
}
```

This preserves the bypass behavior (no JWKS verification) while still rejecting unauthenticated requests, matching what the integration tests expect.

The "invalid Bearer token" test (`Bearer invalid-token`) will also pass because the `catch` block treats non-JWT tokens as raw user IDs and calls `next()` — which matches the old `requireAnonymousAuth` behavior where any valid session token (not just JWTs) was accepted. If you want bypass mode to also reject non-JWT tokens, the catch block would need to return 401 instead.

PR #456, opened 2026-04-25
Retry LML search with track title when album title returns empty

### LINKED ISSUE #453: Metadata service should retry with track title when album title search returns empty
## Bug

When a flowsheet entry has an album title that doesn't exactly match any Discogs release, the metadata service returns only fallback search URLs instead of retrying the LML search with the track title. This means entries with slightly misspelled or variant album titles get no Discogs metadata, artwork, or streaming links — even when the track title alone would have successfully matched.

## Root Cause

In `fetchAlbumMetadata` (`apps/backend/services/metadata/metadata.service.ts`, line 85), the search term is constructed as:

```ts
const searchTerm = albumTitle || trackTitle || '';
```

This uses short-circuit evaluation: if `albumTitle` is present (even if slightly wrong), `trackTitle` is never considered. When the LML search returns 0 results, the function immediately falls through (line 94) to return bare search URLs with no retry:

```ts
if (!lmlResults || lmlResults.results.length === 0) {
  const urls = searchUrls.getAllSearchUrls(artistName, albumTitle, trackTitle);
  return {
    albumMetadata: { youtubeMusicUrl: urls.youtubeMusicUrl, ... },
    artistId: null,
    searchResult: null,
  };
}
```

## Real-World Example

Flowsheet entry #5198490 — Cocteau Twins:
- `album_title`: "cocteau twins singles collections" (incorrect — the actual release is "Singles Collection")
- `track_title`: "crushed"
- LML search with `artist="Cocteau Twins" album="cocteau twins singles collections"` returned **0 results**
- LML search with `artist="Cocteau Twins" album="crushed"` **successfully found the release**

The entry ended up with no artwork, no Discogs link, no Spotify/Apple Music URLs — only generic search URLs.

## Suggested Fix

Add a fallback retry in `fetchAlbumMetadata`: when both `albumTitle` and `trackTitle` are present and the initial search returns 0 results, retry the LML search using the track title as the search term:

```ts
// Primary search: prefer album title
let lmlResults = await searchDiscogs(artistName, searchTerm || undefined);

// Fallback: if album title search returned nothing and we have a track title, retry with it
if ((!lmlResults || lmlResults.results.length === 0) && albumTitle && trackTitle) {
  lmlResults = await searchDiscogs(artistName, trackTitle);
}
```

## Test Plan

- [ ] Add unit test: album title present but returns 0 results, track title succeeds on retry — verify `searchDiscogs` is called twice and metadata is populated
- [ ] Add unit test: album title returns 0 results, track title also returns 0 results — verify no infinite retry, just fallback search URLs
- [ ] Add unit test: album title returns results on first try — verify no retry (single `searchDiscogs` call)
- [ ] Verify existing tests still pass (no retry when `albumTitle` is null and `trackTitle` is used as primary search term)

jakebromberg:
## Summary

- When `fetchAlbumMetadata` gets no results from the primary LML search (using album title), it now retries with the track title before falling back to bare search URLs
- Fixes cases like flowsheet entry #5198490 where a misspelled album title ("cocteau twins singles collections" vs "Singles Collection") prevented metadata from being found, even though the track title ("crushed") would have matched

Closes #453

## Test plan

- [x] Album title returns empty, track title retry succeeds — verify `searchDiscogs` called twice, metadata populated
- [x] Both album and track title searches return empty — verify no infinite retry, falls back to search URLs
- [x] Album title returns results on first try — verify no retry (single `searchDiscogs` call)
- [x] All existing tests pass unchanged

### DIFF
--- a/apps/backend/services/metadata/metadata.service.ts
+++ b/apps/backend/services/metadata/metadata.service.ts
@@ -91,6 +91,15 @@ async function fetchAlbumMetadata(request: MetadataRequest): Promise<{
     // Fall through to construct search URLs
   }
 
+  // Fallback: if album title search returned nothing and we have a track title, retry with it
+  if ((!lmlResults || lmlResults.results.length === 0) && albumTitle && trackTitle) {
+    try {
+      lmlResults = await searchDiscogs(artistName, trackTitle);
+    } catch (error) {
+      console.warn('[MetadataService] LML track title fallback failed:', error);
+    }
+  }
+
   if (!lmlResults || lmlResults.results.length === 0) {
     // No LML results — construct search URLs as bare minimum
     const urls = searchUrls.getAllSearchUrls(artistName, albumTitle, trackTitle);
--- a/tests/unit/services/metadata.service.test.ts
+++ b/tests/unit/services/metadata.service.test.ts
@@ -225,4 +225,57 @@ describe('metadata.service', () => {
     // Year from search result, not release
     expect(result?.album?.releaseYear).toBe(2001);
   });
+
+  describe('track title fallback', () => {
+    it('retries with track title when album title search returns empty', async () => {
+      // First call (album title) returns nothing; second call (track title) finds the release
+      mockSearchDiscogs
+        .mockResolvedValueOnce({ results: [], total: 0, cached: false })
+        .mockResolvedValueOnce({ results: [lmlSearchResult], total: 1, cached: false });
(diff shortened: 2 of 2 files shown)


jakebromberg:
Closing in favor of fixing this at the LML layer. The album title should be sanitized (artist name stripped) before searching Discogs, rather than retrying with a different search term.

PR #512, opened 2026-04-26
Library schema: denormalize artist_name + add search_doc tsvector

### LINKED ISSUE #486: [A.1] Library schema: denormalize artist_name + add search_doc tsvector
Part of [Epic A](https://github.com/WXYC/Backend-Service/issues/483).

## Why

The catalog search's cross-table OR predicate prevents the planner from using either trigram GIN index, leading to 100-200ms median per query. Denormalizing `artist_name` from `artists` onto `library` lets both searchable columns live on a single table; a STORED generated `search_doc` tsvector built from those two columns becomes the index that powers the new search path.

## Scope

- `shared/database/src/schema.ts`: add `library.artist_name varchar(...)` (nullable initially) and `library.search_doc tsvector GENERATED ALWAYS AS (setweight(to_tsvector('simple', coalesce(artist_name, '')), 'A') || setweight(to_tsvector('simple', coalesce(album_title, '')), 'B')) STORED`.
- New migration: add the columns + a GIN index on `search_doc` + a trigram index on `library.artist_name` (the existing `artist_name_trgm_idx` is on `artists.artist_name`, not `library`).
- Tests: schema-source assertions confirming the column declarations and the migration shape (mirroring `tests/unit/database/schema.flowsheet-dj-name.test.ts`).

## Production note

Adding nullable columns is metadata-only. The STORED tsvector column can only be populated for rows with `artist_name` set, so it will be NULL for every existing row until A.2 backfills `artist_name`. That's intended: A.2 + A.3 will populate; A.5 will read.

## Out of scope

- Backfilling `artist_name` from the join (A.2).
- Wiring the live write path / cascade on artist rename (A.3).
- Service-side reads (A.5).

jakebromberg:
## Summary

- Adds `library.artist_name varchar(128)` (nullable) and a STORED generated `library.search_doc tsvector` (artist=A, album=B weight bands).
- Adds two GIN indexes: `library_search_doc_idx` on `search_doc` and `library_artist_name_trgm_idx` on `artist_name`.
- Schema-only step. No backfill, no live-write changes, no service consumption (those are A.2 / A.3 / A.5).

## Why

The catalog search currently OR's a trigram predicate across `artists.artist_name` and `library.album_title`. Because the OR spans two tables, the planner can't use either trigram GIN index and the predicate is evaluated as a join filter after a merge join, costing 100-200ms per query on the production-clone dataset. Putting both searchable columns on a single table -- and indexing them together as a STORED tsvector -- collapses the hot path to sub-1ms.

## Production note

`ALTER TABLE ADD COLUMN` of a nullable varchar is metadata-only and instant. The STORED tsvector column also adds without a row rewrite because every existing row has `artist_name = NULL`, so the generated value is a constant empty tsvector. The two GIN index builds scan the full table; on the 64K-row staging library that is under a second. Apply during a low-traffic window if the production library has grown materially.

## Test plan

- [x] `npm run typecheck`
- [x] `npm run lint` (0 errors)
- [x] `npm run format:check`
- [x] `npm run test:unit` (1060 passing)
- [x] `node scripts/validate-migrations.mjs`
- [x] `npx drizzle-kit generate` reports no further drift after the migration is applied
- [ ] Apply locally and verify both indexes exist on the library table after migration

Closes #486

### DIFF
--- a/shared/database/src/migrations/0056_library-search-doc.sql
+++ b/shared/database/src/migrations/0056_library-search-doc.sql
@@ -1,0 +1,29 @@
+-- Denormalize artist_name onto library + add search_doc tsvector (A.1).
+--
+-- The catalog search currently OR's a trigram predicate across artists.artist_name
+-- and library.album_title. Because the OR spans two tables, the planner can't
+-- use either trigram GIN index and the predicate is evaluated as a join filter
+-- after a merge join, costing 100-200ms per query on the production-clone
+-- dataset. Putting both searchable columns on a single table -- and indexing
+-- them together as a STORED tsvector -- collapses the hot path to sub-1ms.
+--
+-- Order of operations across Epic A:
+--   A.1 (this migration): add the columns nullable + add indexes.
+--   A.2: backfill library.artist_name from the artists join.
+--   A.3: live writes -- addAlbum writes artist_name; cascade on artists rename.
+--   A.4: album_plays materialized view + scheduled refresh.
+--   A.5: service rewrite -- Both mode reads search_doc + ts_rank * ln(plays+1).
+--
+-- Production note: ALTER TABLE ADD COLUMN of a nullable varchar is metadata-only
+-- and instant. The STORED tsvector column also adds without a row rewrite
+-- because every existing row has artist_name = NULL, so the generated value is
+-- a constant empty tsvector. The column will only be populated for rows that
+-- A.2 backfills (and rows the live write path in A.3 inserts). The two GIN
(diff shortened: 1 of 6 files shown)


jakebromberg:
Superseded — main has moved forward (migrations 0056 + 0057 already shipped) and the orchestrator now coordinates migration numbering via [claimSequentialNumber](https://github.com/jakebromberg/claude-orchestrator/issues/20). Re-running the issue (#486) under the new system to produce a fresh PR on top of current main.

Issue #523, opened 2026-04-26
Migration 0056 hanging in prod blocks all auto-deploys


jakebromberg:
## Symptom

Every auto-deploy run since PR #510 merged (commit \`b6ebc38\`, 2026-04-26 04:37 UTC) has failed at the \`migrate\` step with \`Run Command Timeout\` after exactly 10 minutes. The job logs show the migration container hangs immediately after \`🔄 Running Drizzle migrations…\` with no progress and no error. \`deploy\` is gated on \`migrate\`, so every change merged to \`main\` since that point is built and pushed to ECR but never deployed to the EC2 host — including #513's artist-identity-etl image, which is what surfaced this.

Failing run example: https://github.com/WXYC/Backend-Service/actions/runs/24961925864

## Diagnosis

Migration \`0056_project-reconciled-identity-on-library-artist-view.sql\` does:

\`\`\`sql
DROP VIEW "wxyc_schema"."library_artist_view";
CREATE VIEW "wxyc_schema"."library_artist_view" AS (...18 columns + 6 new...);
\`\`\`

\`DROP VIEW\` requires \`AccessExclusiveLock\` on the view. If any session in production holds an \`AccessShareLock\` (an active \`SELECT\` against the view, an idle-in-transaction connection that touched it, the logical-replication tunnel session) the migration waits forever. Drizzle's migrator has no \`lock_timeout\` set, so it sits until the GitHub Actions runner kills it at 10 minutes.

## Fix

Two parts, both needed:

1. **Code (this issue)** — Rewrite migration 0056 to use \`CREATE OR REPLACE VIEW\` (no \`DROP\`, atomic catalog update) wrapped with \`SET LOCAL lock_timeout = '30s'\`. The new view's columns are an append-only superset of the old one, so \`CREATE OR REPLACE VIEW\` works (Postgres only requires existing columns to remain in the same order with the same types). With \`lock_timeout\`, future view migrations that still hit contention will fail loudly within 30s instead of hanging for 10 minutes.

2. **Operational (separate)** — Identify and terminate the session currently blocking 0056 in prod, so the next deploy can apply the migration. Tracked as a follow-up if the rewrite alone doesn't unblock it.

## Definition of Done

- [ ] Migration 0056 SQL replaced with \`SET LOCAL lock_timeout='30s';\` + \`CREATE OR REPLACE VIEW\` (no DROP).
[...]

Issue #521, opened 2026-04-26
Normalize names in artist-identity ETL match (NFC, case folding, diacritics)


jakebromberg:
## Context

\`jobs/artist-identity-etl/job.ts\` matches LML's \`entity.identity.library_name\` to Backend-Service's \`artists.artist_name\` via exact case-sensitive equality:

\`\`\`ts
.where(eq(artists.artist_name, lml.library_name))
\`\`\`

PR #513's design rationale: both sides treat \`library_name\` as the canonical artist key, so exact equality should cover most rows. Mismatches fall through silently as "unmatched" (the row appears in \`identities\` but not in \`runIncremental\`'s \`matched\` count).

Whether normalization is actually needed depends on real-world coverage. After WXYC/Backend-Service#518 wires the ETL into production and the first run logs land, we'll know.

## Action

1. **First**, after the ETL runs once in production, look at the run log: \`Scanned X LML rows; matched Y; updated Z\`. Compute \`(X - Y) / X\` -- the unmatched rate.
2. **If unmatched rate is meaningful** (>5% suggests a real coverage problem; <1% means exact match is fine and this issue can close as won't-fix):
   - Add NFC unicode normalization on both sides before comparison.
   - Optionally add case folding -- but this risks false matches (e.g., "Wire" the band vs "WIRE" the magazine, both in WXYC's catalog as distinct entities). Revisit only if unmatched is high and Unicode-only wasn't enough.
   - Diacritic stripping is the most aggressive option and should be opt-in. WXYC's canonical example data deliberately includes diacritic-bearing names (Nilüfer Yanya, Csillagrablók, Hermanos Gutiérrez), so any normalization rule has to be tested against those.
3. Compare on a normalized expression (e.g., a generated column or a functional index) rather than rewriting both stored values -- this keeps the source data unchanged and makes the rule easy to revert.

## Why "first measure, then act"

Premature normalization risks false matches. The \`artists\` table doesn't currently have a unique constraint on \`artist_name\`, so a too-aggressive normalization could fold "Wire" + "WIRE" into one match and then the COALESCE-in-SET (#513) would propagate the same identity to both -- which is wrong.

Leave this open until #518 has run and we have data.

## Out of scope

- Adding a unique constraint on \`artist_name\` -- separate concern, broader implications.
[...]

Issue #520, opened 2026-04-26
Add orchestration tests for artist-identity ETL job loop


jakebromberg:
## Context

\`jobs/artist-identity-etl/transform.ts\` is well-covered by unit tests (\`columnsToFill\` and \`columnsInConflict\` across all the relevant cases). The orchestration logic in \`job.ts\` -- \`runIncremental\`, \`loadExisting\`, \`applyFill\` -- is not.

PR #513 review flagged this as LOW (deferred) because:
- The \`COALESCE(<col>, $value)\` SET expression in PR #513's review-feedback amend (945ae18) moved the "never overwrite" guarantee into the database, where it holds regardless of how many rows match an artist name.
- The remaining bug surface is mostly in the loop's bookkeeping (matched/updated/conflicts counts), the LML fetch path, and the \`last_run\` timestamp boundary.

## Action

Add a job-level integration test (under \`tests/integration/\` or a new \`tests/e2e/etl-artist-identity.test.ts\`) that:

1. Seeds the test \`artists\` table with two rows sharing the same \`artist_name\` -- one with a populated \`discogs_artist_id\`, one with all-null IDs.
2. Mocks \`fetchLmlIdentities\` to return a single LML row with that \`library_name\` and a different \`discogs_artist_id\`.
3. Runs \`runIncremental\` once.
4. Asserts:
   - The pre-populated row keeps its original \`discogs_artist_id\` (proves COALESCE preserves staff edits).
   - The all-null row gets the LML-supplied \`discogs_artist_id\`.
   - The conflict counter logged in the run output reflects the divergence on the populated row.
   - \`getLastRunTimestamp\` returns the run's start time afterward (so the next run filters from there).

A second case worth covering: an LML row with all-null external IDs is filtered out at the SQL level (the \`WHERE … IS NOT NULL\` clauses in \`fetch-lml.ts\`), so the loop should never see it. A unit test that spies on \`loadExisting\` confirms zero invocations for an empty LML response.

## Out of scope

Continuing the deferred items from #513's review: query batching (#519) and name normalization (separate follow-up).

Part of WXYC/wxyc-shared#82 follow-up. Part of [Shared Types Consolidation](https://github.com/orgs/WXYC/projects/16).

Issue #519, opened 2026-04-26
Batch artist-identity ETL queries to remove N+1 pattern


jakebromberg:
## Context

`jobs/artist-identity-etl/job.ts:60-90` (the `runIncremental` loop) issues two round-trips per LML row: one `loadExisting` SELECT, then one `applyFill` UPDATE. With LML's `entity.identity` at ~thousands of rows today and an hourly cron, that's tolerable -- but the job's runtime grows linearly with reconciled-identity count, and a future polling-mode cadence (or a backfill against a much larger LML table once normalization runs catch more matches) will start mattering.

Flagged in WXYC/Backend-Service#513 review as MEDIUM, deferred for scale reasons.

## Action

Replace the per-row SELECT + UPDATE pair with a single bulk statement. Two viable shapes:

1. **Bulk SELECT keyed by `library_name`**, then per-row UPDATE only for rows with non-empty `toFill`. Halves round-trips.
2. **Single SQL UPDATE with VALUES**:
   ```sql
   UPDATE artists a
   SET discogs_artist_id   = COALESCE(a.discogs_artist_id, v.discogs_artist_id),
       musicbrainz_artist_id = COALESCE(a.musicbrainz_artist_id, v.musicbrainz_artist_id),
       ...
   FROM (VALUES (\$1, \$2, ...), ...) AS v(library_name, discogs_artist_id, ...)
   WHERE a.artist_name = v.library_name
   RETURNING a.id, ...;
   ```
   This collapses fill + null-preservation + multi-row-per-name handling into one statement, and the `RETURNING` clause gives an actual write count for logging. Conflict logging (existing non-null differs from LML's value) becomes a separate read-only diagnostic SELECT before the UPDATE, or just dropped in favor of the post-update RETURNING + a follow-up audit.

Option 2 is preferred -- it's one statement, idempotent, and aligned with how the COALESCE-in-SET fix in PR #513 already moved correctness into the database.

## Out of scope

- Orchestration tests -- separate issue WXYC/Backend-Service#520.
- Name normalization -- separate issue WXYC/Backend-Service#521.

Part of WXYC/wxyc-shared#82 follow-up. Part of [Shared Types Consolidation](https://github.com/orgs/WXYC/projects/16).

PR #560, opened 2026-04-27
Unblock deploy chain: replay 0064 mojibake + verifier allowlist

### LINKED ISSUE #550: drizzle migrator silently skips migrations when journal timestamps regress
## Symptom

Five migrations were silently never applied to production despite being merged to \`main\` weeks ago:

- 0054_flowsheet-search-doc-with-dj-name
- 0055_add-reconciled-identity-to-artists
- 0056_project-reconciled-identity-on-library-artist-view
- 0062_flowsheet-linkage-audit-columns
- 0063_flowsheet-legacy-link-attempted-at

Verified directly against prod RDS: \`artists\` has zero reconciled-identity columns, \`library_artist_view\` is back to 17 columns (no reconciled-identity), and \`flowsheet\` has neither \`linkage_source\` nor \`legacy_link_attempted_at\`.

## Root cause

\`drizzle-orm\`'s postgres-js migrator decides what to apply by comparing each journal entry's \`when\` against the *largest* \`created_at\` already in \`drizzle.__drizzle_migrations\`. Anything \`when <= latest_created_at\` is treated as already done.

\`0061_library-canonical-entity\` was committed with \`when=1779424000000\` (~2026-05-22), 13 days ahead of its sibling timestamps. As soon as 0061 applied, the cursor jumped forward and every entry with a smaller \`when\` that hadn't yet been applied was permanently skipped — including 0054, 0055, 0056 (which had been hanging in prod due to #523 / I/O saturation in #525) and 0062, 0063 (which were authored after 0061 with smaller, natural timestamps).

The validator at \`scripts/validate-migrations.mjs\` does enforce monotonically increasing \`when\` values, but the rule was not in place when 0061 was first committed, so the regression slipped through.

## Action

Two parts:

1. **Recover the lost schema** — PR #526 adds \`0064_recover-skipped-migrations.sql\` which idempotently replays the union of all five skipped migrations (ADD COLUMN IF NOT EXISTS, CREATE OR REPLACE VIEW, CREATE INDEX IF NOT EXISTS, DROP IF EXISTS), with \`when=1779600000000\` so it is greater than the prod cursor.
2. **Prevent recurrence** — Same PR extends the validator with explicit allowlists for the existing 0062/0063 out-of-order entries and the orphan 0046 SQL file. Going forward the validator catches any new out-of-order \`when\` before it can be merged.

## Definition of Done

- [ ] PR #526 merged.
- [ ] Production \`artists\` has all six reconciled-identity columns.
- [ ] Production \`library_artist_view\` has 23 columns.
- [ ] Production \`flowsheet\` has \`linkage_source\`, \`linkage_confidence\`, \`linked_at\`, \`legacy_link_attempted_at\`.
- [ ] Validator catches any future out-of-order \`when\` in CI.

Part of [Shared Types Consolidation](https://github.com/orgs/WXYC/projects/16).

jakebromberg:
## Summary

After #551 merged, the auto-deploy migrate step exited 1 because the post-migration verifier in \`dev_env/init-db.mjs\` reported two journal entries with no matching \`__drizzle_migrations\` row. drizzle's actual migration ran fine — 0065 applied and \`flowsheet.search_doc\` now has \`dj_name\` in production — but the verifier tipped the job to failure, which made \`deploy\` skip. Backend is still on \`v0.1.192\`.

The two flagged entries:

- **0054_flowsheet-search-doc-with-dj-name** — drizzle skipped this one in the cursor regression (#550). 0065 carries its effect; the original 0054 hash will never be inserted.
- **0064_propagate-v012-mojibake** — commit \`7f35fc39\` lowered 0064's \`when\` from 1779683200000 to 1778683200000 to match a hand-inserted prod row. That put it below the current prod cursor (1779424000000), so drizzle silently skipped it. The mojibake fixes never reached production.

This PR does two things:

1. **\`shared/database/src/migrations/0066_replay-mojibake-fixes.sql\`** — a verbatim copy of 0064's UPDATE block at \`when=1779683200002\` (above the new cursor of \`1779683200001\` set by 0065). Every UPDATE is value-keyed (\`WHERE col = 'corrupted_form'\`) so a re-run leaves the corrected rows alone — same idempotency 0064 was already designed for. Same shape we used for 0054 → 0065 in #551.
2. **\`dev_env/init-db.mjs\` grows \`EXPECTED_ABSENT_TAGS\`** — an allowlist for journal entries that will never have a matching hash because their effect lives under a different replay file. Verifier reports allowlisted absent entries as a note and only fails on hashes missing outside the allowlist. Each entry's comment ties it back to its replay so the link is greppable.

Closes #550 (the original cursor-skip bug, now fully recovered). Unblocks #517, #518, #519, #520, #521, #523, and the WXYC/wxyc-shared#82 epic — all of which need a successful deploy run.

## Why a replay instead of bumping 0064's \`when\`

Bumping 0064's \`when\` above the cursor would put it after 0065 in journal-array order while keeping idx=64. The validator's monotonic-\`when\` rule iterates the array in order, so any later contributor's natural-timestamp migration would trip on a now-out-of-order entry. Replay-as-0066 keeps the journal monotonic and idx-ordered, which is the same invariant #551's 0065 maintained.

## Test plan

- [x] \`scripts/validate-migrations.mjs\` passes (1 pre-existing allowlisted warning, 0 errors).
- [x] \`npm run typecheck\` clean.
- [x] \`npm run lint\` clean (0 errors).
- [x] **Local DB end-to-end:** \`npm run db:start\` applies all 66 journal entries to a fresh PostgreSQL — no skips, all hashes recorded.
- [ ] CI integration tests pass.
- [ ] After merge, watch \`Auto Build & Deploy\` for the merge commit; \`migrate\` runs 0066 idempotently (most rows already match the corrected form on local; on prod the 60 rows from the audit get fixed), the verifier passes (0054 + 0064 noted as allowlisted absent, all other hashes present), then \`deploy\` runs.
- [ ] After deploy, verify \`SELECT count(*) FROM wxyc_schema.flowsheet WHERE artist_name = 'μ-Ziq'\` is non-zero in prod (one of the corrected forms from the audit).

### DIFF
--- a/dev_env/init-db.mjs
+++ b/dev_env/init-db.mjs
@@ -119,6 +119,30 @@ async function runMigrations() {
  * independent check by comparing SHA-256 hashes of migration SQL files against
  * the hashes recorded in drizzle.__drizzle_migrations.
  */
+// Tags whose hash is expected NOT to appear in __drizzle_migrations.
+//
+// drizzle's migrator skipped these in production after a journal `when`
+// regression (see #550), and the cursor has since advanced past them.
+// They will never be applied — a follow-up "replay" migration carries
+// their effects under a fresh hash that does land. The verifier needs
+// to know which originals are intentionally absent so it doesn't fail
+// the migrate job for a state we cannot fix retroactively.
+//
+// Each entry is a journal `tag`; both the original and its replay are
+// listed in the comment so future contributors can find the connection.
+const EXPECTED_ABSENT_TAGS = new Set([
+  // 0054 was permanently skipped by drizzle after 0061's far-future
+  // `when` jumped the cursor. PR #551 added 0065 to replay the
+  // search_doc rewrite under a fresh hash.
+  '0054_flowsheet-search-doc-with-dj-name',
+  // 0064 was permanently skipped after commit 7f35fc39 lowered its
+  // `when` below the prod cursor (1779424000000) to match the prod
+  // __drizzle_migrations row that was inserted before the cursor
+  // landing. 0066 replays its mojibake UPDATE block (already idempotent
+  // by design — every UPDATE is keyed on the corrupted value).
+  '0064_propagate-v012-mojibake',
+]);
+
 async function verifyMigrations() {
   console.log(styleText(['bold'], '🔍 Verifying migration completeness...\n'));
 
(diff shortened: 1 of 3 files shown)


jakebromberg:
Superseded by fe95168 in main, which lands the same recovery (0066 mojibake replay + verifier allowlist) plus a more thorough version of the verifier change (paired-replay guard, HISTORICAL_REPLACED_TAGS map, six unit tests pinning allowlist shape). Closing this PR.

PR #627, opened 2026-04-28
Trigger LML metadata enrichment for tubafrenzy webhook upserts

### LINKED ISSUE #626: Tubafrenzy webhook upserts skip LML metadata enrichment
## Problem

iOS app shows missing album art and empty playcut detail screens for recent tracks. Investigation traced the cause to a structural gap in metadata enrichment.

The `/internal/flowsheet-webhook` handler at `apps/backend/routes/internal.route.ts` is the source of essentially all production flowsheet inserts today (tubafrenzy DJ tool → tubafrenzy webhook → BS upsert). The handler upserts the row but never calls `fireAndForgetMetadata`. As a result, every track row arrives with all 10 metadata columns NULL:

- `artwork_url`, `discogs_url`, `release_year`
- `spotify_url`, `apple_music_url`, `youtube_music_url`, `bandcamp_url`, `soundcloud_url`
- `artist_bio`, `artist_wikipedia_url`

The dj-site `addEntry` controller path *does* enrich, but the volume of inserts via that path is effectively zero today (DJs are still on the legacy tubafrenzy DJ tool). The `playlist-proxy` service that serves iOS reads `flowsheet.artwork_url` to decorate playcuts, so it has nothing to read.

## Production confirmation

Coverage by day on prod (queried 2026-04-28):

| Day | Tracks | with `artwork_url` | with `spotify_url` | with `artist_bio` |
|---|---|---|---|---|
| Apr 28 | 50 | 0 | 0 | 0 |
| Apr 27 | 305 | 0 | 0 | 0 |
| Apr 26 | 163 | 0 | 0 | 0 |
| Apr 25 | 261 | 12 | 0 | 0 |
| Apr 21–24 | ~1130 | ~80% | 0 | 0 |
| ≤Apr 20 | 0 | 0 | 0 | 0 |

`spotify_url` and `artist_bio` are zero across every day, including the Apr 21–24 window where `artwork_url` was 80%. That window's coverage came from a one-shot backfill (`flowsheet-lml-link-backfill` / `scripts/backfill-metadata.ts`) — `fireAndForgetMetadata` itself has effectively never written in production, because the webhook is the only insert path with non-trivial volume.

A 10-sample dry run of LML's `/api/v1/lookup` against recent prod nulls returned full enrichment (artwork + spotify + bio + year) for 7/10 samples with 0 errors. The 3 misses are obscure artists not in Discogs' cache — expected, and the existing code handles null fields gracefully.

LML's `/health` is fully green; `LML_API_KEY` and `LIBRARY_METADATA_URL` on EC2 are correct; the Apr 27 raw_message regression and the Discogs-token issue are both already resolved.

## Fix

Wire `fireAndForgetMetadata` into the webhook handler. To avoid duplicating the enrichment logic across two call sites, extract it into a new module `apps/backend/services/metadata/enrichment.service.ts` exporting `fireAndForgetMetadataForRow`. The controller's local `fireAndForgetMetadata` becomes a thin adapter; the webhook handler calls the new function after `.returning({ id })` on the upsert.

The promise floats — the webhook 200 is never blocked on LML, and any LML / DB error is captured to Sentry under `subsystem='metadata'` rather than propagated.

Gate: only fires for `entry_type === 'track'` with a non-null `artist_name`, only on create/update actions (delete is unaffected).

## Out of scope (follow-ups)

- Backfilling historical NULL rows for ~3000 tracks since Apr 25 — separate run of `scripts/backfill-metadata.ts` against prod, with a row-count limit.
- The `flowsheet-lml-link-backfill` SIGTERM mid-run on Apr 28 01:30 UTC — separate concern, that job writes catalog-linkage columns (not metadata) and is unrelated to iOS rendering.
- The `REFRESH MATERIALIZED VIEW CONCURRENTLY album_plays` 5s statement timeouts in BS logs — separate, unrelated to iOS rendering.

jakebromberg:
Closes #626.

## Summary

- `/internal/flowsheet-webhook` now fires `fireAndForgetMetadataForRow` after each track upsert, so every row arriving from tubafrenzy gets the 10-column metadata payload (artwork_url, streaming URLs, artist_bio, etc.) that the iOS app reads.
- Extracts the fire-and-forget enrichment from `flowsheet.controller.ts` into a shared `apps/backend/services/metadata/enrichment.service.ts`, so the controller and the webhook share one implementation. The promise floats — the webhook response is never blocked on LML.
- Adds the upsert's `.returning({ id })` so the new path has a flowsheet id to enrich.

## Test plan

- [x] `npm run typecheck` — passes
- [x] `npm run lint` — 0 errors (pre-existing warnings only)
- [x] `npm run format:check` — passes
- [x] `npm run test:unit` — 1377/1377 passing, including:
  - new `tests/unit/services/metadata.enrichment.test.ts` covering fetch → update, null short-circuit, and Sentry capture on fetch failure
  - new cases in `tests/unit/routes/internal.route.test.ts` asserting the trigger fires on track upserts and skips talksets, empty `artist_name`, and delete actions
  - existing controller tests adapted to verify delegation; the Sentry side-effect assertions move to the enrichment module's tests
- [ ] Smoke after merge: tail backend logs on EC2 for `[Flowsheet] Metadata fetch failed` after the next live show; confirm new tracks land with `artwork_url IS NOT NULL` via a `SELECT … WHERE add_time > now() - interval '1 hour'`.

## Out of scope

- Backfill of ~3000 historical NULL rows since Apr 25 — handled separately with `scripts/backfill-metadata.ts` against prod.
- `flowsheet-lml-link-backfill` SIGTERM mid-run on Apr 28 — unrelated job (writes `album_id`, not metadata).
- `REFRESH MATERIALIZED VIEW CONCURRENTLY album_plays` statement-timeout errors in logs — unrelated to iOS rendering.

## Merge

Use rebase-and-merge per Backend-Service convention.

### DIFF
--- a/apps/backend/controllers/flowsheet.controller.ts
+++ b/apps/backend/controllers/flowsheet.controller.ts
@@ -1,13 +1,11 @@
 import { Request, RequestHandler } from 'express';
 import { Mutex } from 'async-mutex';
-import * as Sentry from '@sentry/node';
-import { eq } from 'drizzle-orm';
-import { db, NewFSEntry as FullNewFSEntry, FSEntry, Show, ShowDJ, flowsheet } from '@wxyc/database';
+import { NewFSEntry as FullNewFSEntry, FSEntry, Show, ShowDJ } from '@wxyc/database';
 
 // play_order is computed by the service layer, not provided by controllers
 type NewFSEntry = Omit<FullNewFSEntry, 'play_order'>;
 import * as flowsheet_service from '../services/flowsheet.service.js';
-import { fetchMetadata } from '../services/metadata/index.js';
+import { fireAndForgetMetadataForRow } from '../services/metadata/index.js';
 import { runLmlLinkage } from '../services/flowsheet-linkage.service.js';
 import { reportLinkageError } from '../services/linkage-metrics.service.js';
 import WxycError from '../utils/error.js';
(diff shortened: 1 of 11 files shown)


jakebromberg:
Addressed review feedback in 38994c3:

- **HIGH** — Enrichment now gated on `xmax = 0` (true INSERT only). Benign retries / no-op updates from tubafrenzy no longer trigger re-fetch + 10-column UPDATE + CDC/index churn. Added test coverage for the update-branch skip case.
- **LOW** — Widened `EnrichmentInput.albumTitle` / `trackTitle` to `string | null`; the `?? undefined` conversion now lives once inside the helper. Added a test asserting null titles forward as undefined to fetchMetadata.
- **LOW** — Added `mockReturning.mockReset()` + a default `mockResolvedValue` in `beforeEach` so queued `mockResolvedValueOnce` values can't bleed across tests.
- **MEDIUM** — SSE not re-emitted after enrichment lands: filed as #628 (pre-existing controller behavior, not introduced here).
- **LOW** — Naming inconsistency: deferred.

Local checks: typecheck ✓, lint ✓ (0 errors), format ✓, unit 1379/1379 ✓ (+2 new).

Issue #628, opened 2026-04-28
Re-emit liveFs SSE event after metadata enrichment UPDATE lands


jakebromberg:
## Context

Both flowsheet write paths fire LML metadata enrichment as a floating Promise after the row is inserted/upserted, via the shared `fireAndForgetMetadataForRow` in `apps/backend/services/metadata/enrichment.service.ts`:

- `apps/backend/controllers/flowsheet.controller.ts` (dj-site `addEntry`)
- `apps/backend/routes/internal.route.ts` — tubafrenzy webhook, gated on `xmax=0` so only fresh inserts enrich (#627, merged 2026-04-28)

Each path also synchronously emits a `liveFs` SSE refetch event *before* the enrichment promise resolves. Connected SSE consumers receive the refetch signal before the metadata UPDATE lands, so they pull the row with `artwork_url`, `spotify_url`, `artist_bio`, etc. all still null. ~1–2s later (bounded — see below) the enrichment UPDATE commits, but no second SSE fires, so consumers never see the metadata until a user-driven refetch.

## Latency bound on enrichment

Worst-case wall-clock from row insert → enrichment UPDATE committed:

- LML lookup is hard-bounded by `TIMEOUT_MS = 5000` in `apps/backend/services/lml/lml.client.ts` (AbortController). On timeout, `lmlFetch` throws `LmlClientError(504)`.
- The `UPDATE flowsheet WHERE id=…` is bounded by `DB_STATEMENT_TIMEOUT_MS=5000` and runs in well under 100ms in practice (PK lookup, single row).
- Total p100 ≈ 5.1s. Typical happy path is 200–800ms.

On the LML-timeout path, `metadata.service.ts:fetchMetadata` still returns a partial result (synthesized YouTube/Bandcamp/Soundcloud search URLs) and an UPDATE still happens — but `artwork_url`, `spotify_url`, `apple_music_url`, `discogs_url`, `release_year`, `artist_bio`, `artist_wikipedia_url` all stay null. So a "second emit on UPDATE-resolved" gated only on UPDATE success would fire even when no user-visible field changed, triggering a wasted refetch. **The second emit should be gated on at least one user-visible field becoming non-null.**

## User-visible impact

iOS users see new tracks appear with no album art for the first ~1–2s after they begin playing. The art never appears for that entry until the user pulls to refresh, or until a *new* track gets added and triggers its own refetch. dj-site users on the now-playing surface see the same thing.

## Affected client paths

[...]

PR #643, opened 2026-04-28
Denormalize library.artwork_url onto flowsheet on bin-pick

jakebromberg:
## Summary

- The bin-pick branch of `addEntry` already calls `getAlbumFromDB` to backfill album info, but did not pull `library.artwork_url` even though the column has existed since migration 0045. This patch adds it to the SELECT so the freshly-inserted flowsheet row carries artwork at INSERT time.
- Eliminates the metadata race for bin-picks on every client surface at once: the `addEntry` HTTP response carries the URL to dj-site (which polls every 60s and won't otherwise see it for up to that long), and the playlist-proxy's synchronous `SELECT artwork_url WHERE …` on the tubafrenzy `created` event reads a non-null value.
- Free-form entries (no `album_id`) still rely on the asynchronous LML enrichment UPDATE — the second-emit + in-process proxy re-enrich tracked in the rest of #628 covers that path.

Part of #628 (does not close it; the free-form-entry second-emit fix is the remainder).

## Test plan

- [x] `tests/unit/controllers/flowsheet.controller.test.ts`: new tests for the populated and null `library.artwork_url` cases on the bin-pick path. All 36 tests in that file pass.
- [x] `prettier --check` on the touched files.
- [ ] After merge, run `SELECT count(*) FILTER (WHERE artwork_url IS NOT NULL), count(*) FROM library` against prod to confirm population is high enough that this patch does meaningful work in practice. (If population is low, this still doesn't regress — null artwork_url passes through and the LML enrichment path takes over as before.)

### DIFF
--- a/apps/backend/services/flowsheet.service.ts
+++ b/apps/backend/services/flowsheet.service.ts
@@ -606,6 +606,11 @@ export const getAlbumFromDB = async (album_id: number) => {
       album_title: library.album_title,
       record_label: library.label,
       label_id: library.label_id,
+      // Denormalized onto the new flowsheet row at INSERT time so dj-site
+      // and the iOS playlist-proxy see artwork immediately, without waiting
+      // for the asynchronous LML enrichment UPDATE to land. Free-form
+      // entries (no album_id) still rely on enrichment. See #628.
+      artwork_url: library.artwork_url,
     })
     .from(library)
     .innerJoin(artists, eq(artists.id, library.artist_id))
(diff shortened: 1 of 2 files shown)


jakebromberg:
Holding this in draft pending the library-side artwork backfill.

Per the precondition comment on #628 ([issuecomment-4338014422](https://github.com/WXYC/Backend-Service/issues/628#issuecomment-4338014422)), the production query reported there shows that `library.artwork_url` is populated for only **155/64163 rows (0.2%)** today.

The denormalization in this PR is correct and harmless, but until the library-side backfill ships, it denormalizes ≈zero rows in practice — the user-visible win is near zero. Filing the backfill as its own follow-up; will rebase + un-draft this once that has run and the population number is materially higher.

In the meantime, surface 2 in #628 (second-emit + in-process proxy re-enrich) is the load-bearing next step and will be picked up in a separate session.

Issue #934, opened 2026-05-16
Suggest endpoints regress to 5s timeouts after mojibake migration #863 (missing ANALYZE on touched tables)


jakebromberg:
## Problem

`GET /flowsheet/suggest/tracks` and `GET /flowsheet/suggest/artists` started hitting Express's 5-second server timeout at 2026-05-15 21:32 UTC and have continued intermittently. These endpoints power inline ghost-text autocomplete in the dj-site flowsheet entry form (introduced in #357), so DJs typing in the entry form experience visible 5s pauses and broken/empty suggestions.

Sentry traces (last 6h):

- https://wxyc.sentry.io/explore/traces/trace/38d1f5d82c0211e1eb0cf5ff93fe0ca9 — `/suggest/tracks?q=man&artist=coupe+cloue&limit=1`, 5052ms; single DB SELECT span 5026ms.
- https://wxyc.sentry.io/explore/traces/trace/71918cd6e51425c0f3f810ce2aa3bc78 — `/suggest/artists?q=The+Ramone&limit=1`, 7973ms; single DB SELECT span 7764ms.

12 internal_error spans on `transaction:"GET /flowsheet/suggest/tracks"` in last 24h vs **0** in the prior 7 days. Brand-new regression.

## Likely cause

Commit 5c126899ccd33080826e65cc5f44a7d2cafc86ab (`Mojibake recovery migration for #863: 61 UPDATEs, 4 dropped`), deployed 2026-05-15, UPDATEd ~61 rows across `flowsheet`, `library`, and `rotation` columns that are covered by GIN trigram indexes (`flowsheet_artist_name_trgm_idx`, `flowsheet_track_title_trgm_idx`, `artist_name_trgm_idx`). The migration did not run `ANALYZE` on the touched tables after the bulk UPDATE.

The Bulk UPDATE playbook (agent memory, not yet promoted to repo docs) notes that per-row cost on `flowsheet` includes the CDC trigger, `search_doc` tsvector regen, and 6 indexes; bulk UPDATEs invalidate query planner stats and can leave GIN indexes bloated. The `suggestArtists` and `suggestTracks` queries rely on these trigram indexes via `ILIKE 'prefix%'` predicates, and the stale planner now appears to be picking a non-trigram path (the all-DB-time 5s+ ceiling — single SELECT span equals total handler time — matches a sequential or bitmap-heap scan that gets cut off by the Express 5s timeout, not by a `statement_timeout`).

## Desired end state

[...]

Issue #892, opened 2026-05-15
[C2] Build CDC-driven enrichment consumer (separate worker)


jakebromberg:
## Problem

Enrichment is wired inline at every insert site (`flowsheet.controller.ts:addEntry`, `internal.route.ts:flowsheet-webhook`). It runs in the BS Express process. The Promise is lost on restart. There's no single component responsible for "this row needs enrichment."

Per `docs/cdc.md`, BS already broadcasts every INSERT/UPDATE/DELETE on `flowsheet` via the CDC WebSocket. So the infrastructure for "subscribe to row inserts" is already running — it's just being consumed only by the reconciliation monitor.

## Consumer cardinality

**Decision: N×N with idempotent claim.** Multiple BS instances each run the consumer; the same CDC event delivered to N instances is safe because the first instance to win the atomic claim does the LML work, and the others see 0 rows updated and move on.

The claim is a single statement:

```sql
UPDATE wxyc_schema.flowsheet
SET metadata_status = 'enriching', enriching_since = now()
WHERE id = $1 AND metadata_status = 'pending'
RETURNING id;
```

If `RETURNING` is empty, a sibling consumer won — log and skip. If `RETURNING` returns the row, this consumer owns the enrichment and proceeds to call LML, then UPDATE to the final state (`'enriched_match'` / `'enriched_no_match'` / `'failed_no_retry'`).

Process death between claim and final UPDATE strands the row at `'enriching'`. The Epic C6 cron sweep handles recovery (see C1 for the index that makes the recovery query cheap and the contract that gives stale claims a 60s grace).

**Why not N×1 with leader election?** Single point of failure that needs its own liveness story. The cron-style pattern is simpler operationally but introduces a routing decision we don't need — the claim is already idempotent.

## CDC fan-out preconditions

The N×N cardinality decision above depends on every BS instance receiving every relevant CDC event independently. Today the only CDC consumer is the reconciliation monitor (`scripts/sync/reconcile.ts`) — single-process. The pattern hasn't been exercised with N simultaneous listeners.

[...]

Issue #873, opened 2026-05-15
Live LML metadata enrichment silently fails on most cold-cache lookups (5 s timeout, no catch-arm fallback)


jakebromberg:
## Symptom

In the wxyc-ios-64 v2 playlist right now, most recently-added playcuts have no inline metadata (no `artwork_url`, no `discogs_url`, no `release_year`, no streaming URLs — not even the synthesized YouTube/Bandcamp/SoundCloud fallback search URLs that don't require any API call). Snapshot from `GET https://api.wxyc.org/flowsheet?limit=20` on 2026-05-15 ~01:00 UTC, single show:

| id | artist / track | inline metadata |
|---|---|---|
| 5210341 | Felt / Evergreen Dazed | full (artwork+discogs+spotify+yt/bc/sc) |
| 5210340 | 22 Beaches / Breathing | all-null |
| 5210337 | El Keamo / Esperando | all-null |
| 5210336 | Rita Villa / Czardas | all-null |
| 5210328 | Able Noise / Inertia | all-null |
| 5210327 | Caifanes / La Negra Tomasa | all-null |
| 5210319 | Sonido Dueñez / La Piragua | all-null |
| 5210317 | The Fly Girlz / Born 2 Be Fly | search-URLs only (yt/bc/sc) |
| 5210316 | Amy Gadiaga / Imma Pick You Up | all-null |
| 5210314 | Jeff Parker / Executive Life | all-null |
| 5210294 | Asha Puthli / Space Talk | full |

iOS renders the empty artwork as a placeholder; ~70% of the show's rows look broken until the 06:00 UTC `flowsheet-metadata-backfill` cron catches up the next morning.

## Root cause: LML lookup latency now consistently exceeds BS's 5s timeout

Direct prod measurement (auth'd `POST /api/v1/lookup` against `library-metadata-lookup-production.up.railway.app`, cold cache, 2026-05-15 ~02:00 UTC):

| Artist | wall time | `cache_stats` |
|---|---|---|
| 22 Beaches | **16.0 s** | `pg_misses=1 api_calls=5 api_time_ms=22927` |
| El Keamo | **9.5 s** | `pg_misses=1 api_calls=5 api_time_ms=13706` |
| Rita Villa | **14.4 s** | `pg_misses=3 api_calls=7 api_time_ms=18582` |
| Sonido Dueñez | **9.5 s** | `pg_misses=1 api_calls=5 api_time_ms=13478` |
[...]

Issue #270, opened 2026-03-30
LML client constructs double /api/v1 prefix in proxy URLs


jakebromberg:
## Problem

The LML HTTP client in `apps/backend/services/lml/lml.client.ts` constructs request URLs by concatenating `LIBRARY_METADATA_URL` with paths that already include `/api/v1`:

```typescript
const url = `${getBaseUrl()}${path}`;
// getBaseUrl() returns: "http://host:8001/api/v1"
// path is: "/api/v1/discogs/search"
// result: "http://host:8001/api/v1/api/v1/discogs/search"
```

When `LIBRARY_METADATA_URL` includes the `/api/v1` prefix (as documented in `.env.example`), every LML request gets a doubled path prefix. FastAPI returns 404 for these URLs, causing all Discogs metadata to be silently skipped via `Promise.allSettled`.

## Impact

- The `/proxy/metadata/album` endpoint returns no Discogs data (release year, genres, styles, label, artwork URL, artist ID)
- Since `discogsArtistId` is never populated, the iOS app skips the `/proxy/metadata/artist` call entirely
- Streaming links (Spotify, Apple Music, etc.) still work since they don't go through LML

## Fix

- Make `getBaseUrl()` strip a trailing `/api/v1` from the env var so the client is tolerant of either format
- Correct `.env.example` to not include the prefix
- Add a regression test that catches the double prefix

PR #1004, opened 2026-05-22
feat(flowsheet): explicit metadata_status enum lifecycle (BS#891)

jakebromberg:
## Summary

Replaces the implicit two-column state machine (`metadata_attempt_at` + populated-column inspection) with an explicit `metadata_status` enum on `flowsheet`. Five states encode the full enrichment lifecycle:

- **`pending`** — never tried OR transient failure (retry-eligible). Default on every new row.
- **`enriching`** — a consumer instance has claimed this row and is mid-LML-call. Set when Epic C C2 (#892) flips a row from `pending` to `enriching` via the idempotent claim. `enriching_since` carries the claim timestamp.
- **`enriched_match`** — LML returned full Discogs metadata; populated columns are authoritative.
- **`enriched_no_match`** — LML succeeded with no Discogs match; only the synthesized YouTube/Bandcamp/SoundCloud search URLs are populated (post-#873 fallback path).
- **`failed_no_retry`** — exceeded the retry budget; terminal.

Companion `enriching_since timestamptz` carries the race-claim timestamp; the C6 (#896) recovery sweep uses a 60s TTL against it to revert stuck rows.

## Migration shape

DDL-only. The `CASE`-derived backfill from existing state is documented but not inlined — a 2.6M-row UPDATE inside the migration transaction would block writes for the duration. Backfill recipe at [`docs/flowsheet-metadata-status-backfill.md`](https://github.com/WXYC/Backend-Service/blob/feat/891-metadata-status-column/docs/flowsheet-metadata-status-backfill.md).

- `CREATE TYPE wxyc_schema.metadata_status_enum AS ENUM(...)`
- `ALTER TABLE flowsheet ADD COLUMN metadata_status ... NOT NULL DEFAULT 'pending'` — constant-default ADD COLUMN is metadata-only on PG11+; no row rewrite, no AccessExclusiveLock beyond the catalog update.
- `ALTER TABLE flowsheet ADD COLUMN enriching_since timestamptz` — nullable, no default.
- Two partial indexes with `IF NOT EXISTS` (same pattern as 0070 / 0074):
  - `flowsheet_metadata_status_pending_idx` on `(id)` WHERE `entry_type='track' AND artist_name IS NOT NULL AND metadata_status='pending'` — covers the C6 sweep.
  - `flowsheet_metadata_status_enriching_stale_idx` on `(enriching_since)` WHERE `metadata_status='enriching'` — covers the stale-claim recovery sweep.

**Ops note:** Build both indexes CONCURRENTLY on prod before merging. The migration's `IF NOT EXISTS` clauses make the apply a no-op when the prebuilt indexes are present. Statements for the CONCURRENTLY build are in the migration's comment block.

## V2 wire format

`transformToV2` now projects `metadata_status` onto track rows. iOS branches on it to decide whether to render inline metadata or fall back to the proxy-fetch path (WXYC/wxyc-ios-64#287 already shipped the decoder; #270 will ship the consumer logic).

`metadata_attempt_at` stays as a historical marker — once Epic C C6 (#896) flips the cron predicate to `metadata_status = 'pending'`, it's no longer used for control flow. The `metadata_attempt_at` partials in `schema.ts` and the corresponding cron query are not removed in this PR; they get dropped alongside the cron flip.

## Test plan

- [x] `tests/unit/database/schema.flowsheet-metadata-status.test.ts` — 14 schema-source tests pin migration + `schema.ts` to the documented enum order, the constant-default ADD COLUMN shape, the two partial-index predicates, the no-CONCURRENTLY-in-DDL rule, and the no-inline-backfill rule.
- [x] `tests/unit/services/flowsheet.transformToV2.metadata-status.test.ts` — 6 unit tests covering all 5 enum values projected onto track rows + the non-track exclusion regression guard.
- [x] Full unit suite: 2059/2059 passing.
- [x] `node scripts/validate-migrations.mjs` clean.
- [x] `npm run lint` introduces no new warnings/errors on the touched files.
- [x] `npm run format:check` clean.

## Critical-path next steps (out of scope for this PR)

1. Build the two partial indexes CONCURRENTLY on prod.
2. Merge this PR.
3. Run the backfill per [`docs/flowsheet-metadata-status-backfill.md`](https://github.com/WXYC/Backend-Service/blob/feat/891-metadata-status-column/docs/flowsheet-metadata-status-backfill.md).
4. Ship #892 (consumer claim), #896 (cron predicate flip), and the other Epic C children.
5. Drop the now-unused `metadata_attempt_at_pending_*` partials.

## Related

- Parent epic: #877 (Epic C)
- Unblocks iOS: WXYC/wxyc-ios-64#287 (decoder, draft), WXYC/wxyc-ios-64#270 (consumer logic)
- Project: post-launch service hardening (#32)
- Sibling references: #873 (synthesized-URL fallback, the `enriched_no_match` reachability requirement), #934 (post-bulk-update `ANALYZE` rule)


jakebromberg:
### Code review

Found 2 issues:

1. Wrong issue reference: the PR body, schema.ts comments, docs/migrations.md, and docs/flowsheet-metadata-status-backfill.md all attribute the cron predicate flip to "Epic C C6 (#896)" / "BS#896". But [#896](https://github.com/WXYC/Backend-Service/issues/896) is titled `[C7] Verify ETL-inserted rows reach the enrichment consumer via CDC` — it does not flip the cron predicate. The actual C6 (`Retune flowsheet-metadata-backfill cron as safety-net`, which contains the `metadata_attempt_at IS NULL` → `metadata_status = 'pending'` predicate change) is [#895](https://github.com/WXYC/Backend-Service/issues/895). Since these references are load-bearing context for the staged drop of `metadata_attempt_at` partials and the backfill ordering, they should point to #895.

https://github.com/WXYC/Backend-Service/blob/576c3b5a734d0d3ae2377677f4e96d0860c030c2/shared/database/src/schema.ts#L749-L774

https://github.com/WXYC/Backend-Service/blob/576c3b5a734d0d3ae2377677f4e96d0860c030c2/docs/flowsheet-metadata-status-backfill.md#L4-L14

https://github.com/WXYC/Backend-Service/blob/576c3b5a734d0d3ae2377677f4e96d0860c030c2/docs/migrations.md#L102-L106

2. Stale runbook path in migration SQL comment: line 33 of the migration points to `docs/runbooks/flowsheet-metadata-status-backfill.md`, but there is no `docs/runbooks/` directory in the repo — the actual runbook lives at `docs/flowsheet-metadata-status-backfill.md` (the path the PR body uses). Comment-only fix, but the migration's content hash is frozen via `applied-hashes.json`, so updating the comment requires re-running `node scripts/freeze-migration-hashes.mjs` (or pinning per the existing `HISTORICAL_NO_ANALYZE_NEEDED_TAGS`-style allowlist if you'd rather leave the file untouched).

https://github.com/WXYC/Backend-Service/blob/576c3b5a734d0d3ae2377677f4e96d0860c030c2/shared/database/src/migrations/0078_flowsheet-metadata-status.sql#L27-L34

Issue #899, opened 2026-05-15
[D3] Update enrichment consumer to upsert album_metadata (linked rows)


jakebromberg:
## Problem

After D1+D2 land, the enrichment consumer from Epic C must be updated to write into `album_metadata` instead of `flowsheet`'s 10 columns. New rows that arrive with a `null` album_id still need handling — they should either get linked to an album_id first (via the existing linkage flow), or write transient data to a non-album-keyed surface.

## Proposal

Update `enrichment.service.ts` (post-Epic-C, where the consumer logic lives):

```typescript
async function writeEnrichment(flowsheetId: number, lookup: LookupResponse) {
  const albumId = await resolveAlbumId(flowsheetId);  // may return null
  if (albumId !== null) {
    // The metadata lives at the album level — upsert it there.
    await db.insert(album_metadata).values({
      album_id: albumId,
      artwork_url, discogs_url, ... ,
    }).onConflictDoUpdate({
      target: album_metadata.album_id,
      set: { ...same fields... },
    });
    await db.update(flowsheet).set({
      metadata_status: lookup.results.length > 0 ? 'enriched_match' : 'enriched_no_match',
    }).where(eq(flowsheet.id, flowsheetId));
  } else {
    // No album linked yet — record the enrichment on the flowsheet row's
    // transient columns. The linkage path may resolve album_id later and
    // promote the data to album_metadata.
    await db.update(flowsheet).set({
      artwork_url, discogs_url, ... ,
      metadata_status: 'enriched_match',
    }).where(eq(flowsheet.id, flowsheetId));
  }
}
```

The transient-row write keeps the 10 columns useful for unlinked rows but doesn't violate the "once per album" goal — once linkage succeeds, the next enrichment touching the same album_id writes album_metadata and the per-row columns become stale-but-harmless (the V2 join falls through to album_metadata).

## Acceptance

- [ ] Linked rows: enrichment writes album_metadata, not flowsheet's 10 columns.
- [ ] Unlinked rows: enrichment writes flowsheet's 10 columns as today.
- [ ] After D4 (column drop): unlinked rows can't enrich at all, which is intentional — they need linkage first.
- [ ] Test: a dj-site addEntry for an already-album_metadata-populated album produces zero `UPDATE flowsheet SET artwork_url = ...` statements.

## Related

- Parent: #878 (Epic D)
- Depends on: D1, D2, C2 (the consumer)

Issue #895, opened 2026-05-15
[C6] Retune flowsheet-metadata-backfill cron as safety-net (hourly, 15-min grace)


jakebromberg:
## Problem

The `flowsheet-metadata-backfill` cron is sized as the safety net for runtime fire-and-forget failures: once daily at 06:00 UTC, sweep `WHERE metadata_attempt_at IS NULL`. For a typical 4-hour show, a track that timed out at 21:00 ET (01:00 UTC) waits 5 hours for the next sweep. Listener experience: broken artwork for hours.

After Epic C's consumer is live (C2) and the inline path is dropped (C5), the cron's role changes. Most rows are now enriched within seconds via the consumer. The cron's actual job is "rows the consumer missed due to deploy/restart/event-loss." That's a much smaller workload — minutes of work, not hours — and it should run on a tighter cadence.

## Proposal

Adjust `jobs/flowsheet-metadata-backfill/`:

1. Change WHERE clause from `metadata_attempt_at IS NULL` to `metadata_status = 'pending' AND inserted_at < now() - interval '15 minutes'`. The 15-min grace lets the consumer try first; the cron only handles whatever fell through.

2. Change `package.json:cron-schedule` from `0 6 * * *` (daily) to `0 * * * *` (hourly) initially. Monitor: if the cron consistently finds < 50 rows per run, drop to every 15 min. If it finds thousands, the consumer has a leak — investigate.

3. The orchestrator-pause-during-active-show logic from #735 stays — we still don't want the cron competing with live LML traffic during peak shows.

## Sizing

Today the cron processes ~50k rows/night per the #641 throughput math. Post-Epic-C, it should process ~50 rows/hour (the long tail of consumer-missed events). Much smaller batches, much faster runs, much less impact on LML's Discogs ceiling.

## Acceptance

- [ ] WHERE clause updated.
- [ ] Cron cadence dropped from daily to hourly (or finer once measured).
- [ ] After 7 days post-Epic-C: hourly run finds < 100 rows median (verify in the existing job logs).

## Related

- Parent: #877 (Epic C)
- Depends on: C2 (consumer must be the primary enrichment path)
- BS#641 — the original cron rollout
- BS#735 — orchestrator cooperative pause; this child preserves that interaction

PR #1013, opened 2026-05-23
feat(enrichment-worker): Dockerfile + healthcheck + two-consumer integration test (BS#892 PR-3)

jakebromberg:
PR-3 of the BS#892 split. PR-1 (#1007) shipped the scaffold + claim primitive. PR-2 (#1010) wired the full chain. PR-3 makes the worker prod-deployable and pins the idempotent-claim contract against real PostgreSQL.

This closes BS#892 once deployed and prod-verified.

## What

**Dockerfile.enrichment-worker** — Multi-stage node:24-alpine. Bundles `@wxyc/database` + `@wxyc/lml-client` workspaces in the build stage, `npm install --omit=dev` + copies built dist in the prod stage. Mirrors `Dockerfile.flowsheet-metadata-backfill` (the closest pattern: long-running container that needs LML client). `CMD npm start` runs the Sentry --import preload via the workspace start script.

**healthcheck.ts + worker.ts wiring** — Minimal `node:http` server on container port 8080 (host port 8083 via `publishPort`). The deploy-service action does `--health-cmd="wget -qO- http://localhost:8080/healthcheck"`; without an HTTP server the container would be marked unhealthy on first poll and restart-thrash. Returns 503 with body `cdc not connected` until the CDC listener flips connected, also catches lost-connection-without-reconnect silently. Pure `node:http`, no Express dep — keeps the container small.

**Two-consumer integration test** (`tests/integration/enrichment-worker-claim.spec.js`) — Pure SQL against real PG. Covers:
1. Two concurrent claims on the same pending row → exactly one wins
2. N=10 concurrent claims on the same pending row → exactly one wins
3. Once `enriching`, no further claim can win
4. All 3 terminal states block further claims (parameterized: `enriched_match`, `enriched_no_match`, `failed_no_retry`)
5. C6-style stranded-claim recovery: backdate `enriching_since` past 60s → recovery sweep reverts to `pending` → re-claim succeeds
6. Recovery sweep does NOT touch rows with recent `enriching_since`

The C6 sweep (#895) ships as its own PR, but the SQL it'll run is pinned here — a schema change that breaks the sweep fails CI now, not when C6 lands.

## Why pure-SQL integration test (not import the claim primitive)

The integration runner is babel-jest with no TS support. Adding a ts-jest transform crashes drizzle-orm's `extractTablesRelationalConfig` on `gin_trgm_ops` indexes (see `library-identity-backfill.spec.js` header for the exact failure mode). The division of responsibility:

- **Unit tests** (`tests/unit/apps/enrichment-worker/claim.test.ts`) — source-code shape: typed builders, WHERE narrowing, race-detector returning behavior. Mocked DB.
- **Integration test** (this PR) — SQL contract under real concurrency. Validates PG row-level locking + BS#891's enum + the partial index.

## Deploy shape

apps/enrichment-worker now slots into the auto-discovery matrix in `deploy-base.yml`:
- Turbo detects affected workspace on push (auto deploy) or explicit target (manual deploy)
- `target_type=app` because it's under `apps/`
- `deploy-service` action does `docker run -d -p 8083:8080 --restart unless-stopped --memory=450m --env-file .env --health-cmd="wget...healthcheck"`

First deploy will need:
1. Push to `main` → triggers Auto Build & Deploy → ECR push + EC2 `docker run`
2. Verify `docker ps` shows `enrichment-worker` healthy
3. Verify `wget localhost:8083/healthcheck` returns `ok` on EC2
4. Tail logs: `docker logs -f enrichment-worker` should show `[enrichment-worker] subscribed to CDC; awaiting events`
5. Trigger a flowsheet INSERT and confirm via Sentry trace explorer: `enrichment.consumer.tick` span with `enrichment.outcome=enriched_match|enriched_no_match|claim_lost`

## NOT in this PR

- **C5 (#894)** drop fire-and-forget — blocked on prod verification of this PR
- **C6 (#895)** cron retune — blocked on prod verification of this PR
- **D3 (#899)** album_metadata consumer upserts — depends on full C2 in prod
- **C7 (#896)** verify ETL CDC delivery — separate scope (rotation-etl + flowsheet-etl); integration test that flowsheet-etl INSERTs produce CDC events the consumer picks up

## Related

- WXYC/Backend-Service#892 (closed by this PR once deployed + prod-verified)
- WXYC/Backend-Service#894, WXYC/Backend-Service#895, WXYC/Backend-Service#896, WXYC/Backend-Service#899 (unblocked by this PR's deploy)
- WXYC/Backend-Service#1007 (PR-1, merged)
- WXYC/Backend-Service#1010 (PR-2, merged)
- WXYC/Backend-Service#891 (the 5-state enum the claim contract depends on)

### DIFF
--- a/Dockerfile.enrichment-worker
+++ b/Dockerfile.enrichment-worker
@@ -1,3 +1,11 @@
+# Long-running CDC consumer (BS#892 / Epic C C2). Subscribes to BS's
+# `pg_notify('cdc')` channel and enriches every new flowsheet track row via
+# LML. Inbound concurrency is bounded by the shared `Semaphore(5)` +
+# `TokenBucket(50/min)` chokepoint in `@wxyc/lml-client`; per-row claim and
+# finalize UPDATEs are single-row PK lookups that finish in milliseconds.
+# Runs alongside `backend` + `auth` on the same EC2; independent restart
+# cycle per `--restart unless-stopped` in the deploy-service action.
+
 #Build stage
 FROM node:24-alpine AS builder
 
@@ -27,12 +35,9 @@ COPY --from=builder ./enrichment-worker-builder/apps/enrichment-worker/dist ./ap
 COPY --from=builder ./enrichment-worker-builder/shared/database/dist ./shared/database/dist
 COPY --from=builder ./enrichment-worker-builder/shared/lml-client/dist ./shared/lml-client/dist
 
-# Long-running CDC consumer (BS#892). The handler chain is bounded by the
-# shared LML Semaphore(5) + TokenBucket(50/min) chokepoint in
-# `@wxyc/lml-client`; individual claim/finalize UPDATEs are single-row PK
-# lookups that finish in milliseconds. Inherit the 5s default — anything
-# longer indicates a stuck connection or runaway query, neither of which
-# should wedge the consumer loop.
+# Inherit @wxyc/database's 5s statement timeout default — every query the
+# consumer issues is a single-row PK lookup; anything longer is a stuck
+# connection or runaway query, not normal load.
 ENV DB_APPLICATION_NAME=wxyc-enrichment-worker
 
 CMD ["npm", "start", "--workspace=@wxyc/enrichment-worker"]
--- a/Dockerfile.enrichment-worker
+++ b/Dockerfile.enrichment-worker
@@ -1,0 +1,38 @@
+#Build stage
+FROM node:24-alpine AS builder
+
+WORKDIR /enrichment-worker-builder
+
+COPY ./package.json ./package-lock.json ./
+COPY ./tsconfig.base.json ./
+COPY ./shared/database ./shared/database
+COPY ./shared/lml-client ./shared/lml-client
+COPY ./apps/enrichment-worker ./apps/enrichment-worker
+
+RUN npm ci && npm run build --workspace=@wxyc/database --workspace=@wxyc/lml-client --workspace=@wxyc/enrichment-worker
+
+#Production stage
+FROM node:24-alpine AS prod
+
+WORKDIR /enrichment-worker
+
+COPY ./package* ./
+COPY ./apps/enrichment-worker/package* ./apps/enrichment-worker/
+COPY ./shared/database/package* ./shared/database/
+COPY ./shared/lml-client/package* ./shared/lml-client/
+
+RUN npm install --omit=dev
+
+COPY --from=builder ./enrichment-worker-builder/apps/enrichment-worker/dist ./apps/enrichment-worker/dist
+COPY --from=builder ./enrichment-worker-builder/shared/database/dist ./shared/database/dist
+COPY --from=builder ./enrichment-worker-builder/shared/lml-client/dist ./shared/lml-client/dist
+
+# Long-running CDC consumer (BS#892). The handler chain is bounded by the
+# shared LML Semaphore(5) + TokenBucket(50/min) chokepoint in
+# `@wxyc/lml-client`; individual claim/finalize UPDATEs are single-row PK
+# lookups that finish in milliseconds. Inherit the 5s default — anything
+# longer indicates a stuck connection or runaway query, neither of which
+# should wedge the consumer loop.
+ENV DB_APPLICATION_NAME=wxyc-enrichment-worker
+
+CMD ["npm", "start", "--workspace=@wxyc/enrichment-worker"]
--- a/apps/enrichment-worker/healthcheck.ts
+++ b/apps/enrichment-worker/healthcheck.ts
@@ -1,0 +1,54 @@
+/**
+ * Minimal HTTP healthcheck server for the enrichment worker (BS#892 / PR-3).
+ *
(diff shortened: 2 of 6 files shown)


jakebromberg:
## Review

Overall: ready to merge once one substantive item is addressed. Cleanly scoped — Dockerfile, healthcheck server, and the SQL-contract integration test are all the right shape for what BS#892 PR-3 promises. CI green; integration test exercises the right invariants (row-level locking, terminal-state contract, C6 sweep pin).

### Substantive

**`healthcheck.ts` over-promises what the code delivers.** The file header states the endpoint "catches lost-connection-without-reconnect silently." But `cdcConnected` only flips back to `false` in `shutdown(signal)` — there's no signal path for a mid-run LISTEN drop. If the underlying postgres connection silently dies (PG restart, network blip without RST, idle-transport death of the kind we just hit in [`project_sentry_idle_transport_death.md`](https://github.com/WXYC/Backend-Service/blob/main/CLAUDE.md)), the process stays alive, `cdcConnected` stays `true`, the probe stays 200, and no events arrive. The deploy framework's `--restart unless-stopped` won't help because the container never goes unhealthy.

Two ways to resolve:

1. **Tighten the comment to what the code actually guarantees.** Drop the "lost-connection-without-reconnect" claim; describe it as a startup/shutdown probe only. File a follow-up for real liveness detection.
2. **Wire a disconnect signal.** If `@wxyc/database` exposes a hook for LISTEN-connection lifecycle (e.g., an `onCdcDisconnect` callback), call it to set `healthState.cdcConnected = false`. If no such hook exists, this PR isn't the place to add it — option 1 is the right move with a follow-up filed.

This is load-bearing because the entire reason for the healthcheck server (per the file header) is to give the deploy framework a way to detect a wedged worker. The current implementation only detects two states where a wedge would be self-evident anyway (during `startCdcListener()` and during `shutdown()`).

### Minor / nice-to-have

- **`tests/integration/enrichment-worker-claim.spec.js:39` SQL drift risk.** The integration test hand-copies `claimRowForEnrichment`'s SQL with a comment that says "when that file is hand-edited the SQL here must follow." Acceptable given the babel-jest + drizzle-orm constraint, but a follow-up grep-based drift check (compare normalized SQL between `claim.ts` and this spec at lint time) would make the contract enforceable. Not blocking.

- **Dockerfile comment placement.** The `# Long-running CDC consumer (BS#892). The handler chain is bounded by ...` block sits immediately above `ENV DB_APPLICATION_NAME=…` and `CMD npm start`. The content (handler timeout reasoning) doesn't describe either of those lines; it documents runtime behavior that lives in worker code. Either move it next to where the timeout is configured, or split into a top-of-file `# Long-running CDC consumer` header plus shorter inline comments.

- **`tests/integration/enrichment-worker-claim.spec.js:122` "N=10 concurrent" framing.** With `pool.max=5`, only 5 of the 10 actually race in parallel; the other 5 execute as the first 5 release connections. The contract still holds (exactly one wins) and the test passes for the right reason, but the framing implies more parallelism than the runtime delivers. Either rename to `N=10 concurrent attempts (5 in parallel + 5 serialized via pool)` or bump pool.max temporarily for this spec. Cosmetic.

- **`apps/enrichment-worker/healthcheck.ts:33` URL match.** `req.url === '/healthcheck'` won't match `/healthcheck?foo=bar`. wget doesn't append query strings so this is fine for the actual probe — but if a canary or operator ever curls with a cache-buster, they'll see a 404. Two-line fix with `URL` parsing; not load-bearing.

- **No unit test for `healthcheck.ts`.** 54 lines, four behaviors (200 connected, 503 disconnected, 404 unknown path, listen on port). Worth a small Jest spec for the 503/200/404 paths against a `HealthState` stub. Not blocking — but the connection-loss gap above is exactly the kind of thing a unit test would have made visible during implementation.

- **Dockerfile runs as root.** No `USER node` (or equivalent) directive. Likely consistent with sibling Dockerfiles in the repo, but if this PR is the first new one in a while, it's a chance to drop privileges. Out-of-scope for this PR; just flagging.

- **`publishPort: "8083"` is a string, not a number** (`apps/enrichment-worker/package.json:7`). Worth confirming the deploy-service action accepts both, or matching whatever the sibling apps use. Cosmetic.

### Tests

- Integration coverage is strong: two concurrent, N=10, already-enriching, all three terminal states (parameterized), C6 sweep pinned with both positive (stranded → reverted) and negative (recent → untouched) cases. The C6 SQL pinning is the right kind of defense — catches breakage at PR review for #895 instead of at deploy time.

- Hand-copied schema in `insertPendingRow` doesn't pass `show_id`. If the schema later gains a `NOT NULL` constraint there, this spec breaks. Minor — known WXYC pattern that show_id is nullable for non-show rows, but worth a one-line comment locking in the assumption.

### Verdict

Address the healthcheck connection-loss gap (either tighten the comment + file follow-up, or wire a hook if `@wxyc/database` has one). The rest is cosmetic and can ride along or be deferred. Once that one item is resolved, this is good to merge — and once deployed + Sentry-verified, BS#892 closes.

Issue #15, opened 2024-05-20
Test & Deploy Current Version


dvdokkum:
Hey @AyBruno! 

So just want to confirm here that what's currently here in `main` is not deployed to prod, and that the blocker for deploying is just testing all of the new changes that you pushed 2 months ago... correct or no?

AyBruno:
@dvdokkum  That is correct. What's currently on main is not deployed. I do think it would be a good idea to test locally before deploying, but I would also feel comfortable deploying and testing on prod. Deploying itself is quite the blocker. It takes a while and AWS permissions are a pain. If you have any tips on how I could automate deployment of containers to the ec2 instance that would be huge.

dvdokkum:
So just to circle back to this... there's now a deploy action in Github. By default it will dry-run and do the following:
- build the docker image
- compare that image to what is currently latest in ECR and indicate whether it is new
- show currently running images on the prod server

If you set the dry-run prop to `false` it will do the deploy as outline in the Readme:
- build and push to ECR
- pull latest on the prod server and restart

I haven't been able to test the actual deploy yet, so let me know when would be a good time (when we're ready to push main to prod) and I'll make sure it is working for real for real.

AyBruno:
@dvdokkum The program at this point should build and compile successfully so we should be good to go with testing the deploy action out.

dvdokkum:
great, i'll give it a shot now!

dvdokkum:
ok that's done. what's on `main` has been deployed

PR #1154, opened 2026-05-26
feat: Library endpoints and search filters for MD catalog editing

JacksonMeade:
## Summary

Backend support for Music Director catalog editing and richer catalog search on the Card Catalog.

- Adds `PATCH /library/:id` album update with validation (artist in genre, label resolution, etc.).
- Adds artist discovery helpers: in-genre artist search, peek artist number, and related service methods used by the add/edit forms.
- Extends library query/search with multi-value genre and format filters, rotation bin tags, and a missing-album filter; aligns cascade and SQL paths with the new params.
- Wires routes under existing `catalog:read` / `catalog:write` permission checks; integration tests cover update, artist search, and query filter behavior.

Pairs with the dj-site `feat/md-interface` PR for end-to-end MD catalog maintenance.

## Test plan

- [ ] `PATCH /library/:id` updates album metadata for authorized callers; rejects invalid artist/genre pairings
- [ ] Artist search and peek-code endpoints return expected shapes for form autocomplete
- [ ] `/library/query` honors `genres`, `formats`, `rotation_bins`, and `missing` filters
- [ ] Run integration suite (`library.spec.js`, `library-query.spec.js`)


jakebromberg (dismissed):
Found 15 issues, ranked most-severe first. The new `PATCH /library/:id` is the riskiest surface — it uses PATCH-route semantics but the handler unconditionally rewrites every column from the body, so several optional fields can be silently wiped on a partial edit.

### Blocking

**1. `genres filter ORs multiple genre names` test will fail in CI** — `tests/integration/library-query.spec.js:95`
CI loads only `dev_env/seed_db.sql` (`LOAD_CLONE_FIXTURE=false`); the seed has 11 library rows across genre_ids 11/6/15 only — no Jazz. `/library/query?genre=Jazz` returns total=0, so `expect(jazzOnly.body.results.length).toBeGreaterThan(0)` fails before the combined-filter assertions run. Either seed a Jazz row or pick two genres that the seed already covers (e.g. Rock + Hiphop).

**2. `parseEnumQueryList` crashes with TypeError on repeated query keys** — `apps/backend/services/library-search.service.ts:448`
Express 5's default `'simple'` parser (Node `querystring.parse`) yields a `string[]` for repeated keys. `?genres=Rock&genres=Jazz` gives `req.query.genres = ['Rock','Jazz']`; inside `parseEnumQueryList` the array passes the truthy `if (!value)` guard, then `value.split(',')` throws `value.split is not a function` → 500. Same path for `?genre=`, `?format=`, `?formats=`, `?rotation_bins=`. Normalize each rest-arg via `Array.isArray(x) ? x : [x]` or coerce upstream.

**3. `searchArtistsInGenre` `q.trim()` crashes on repeated `q`** — `apps/backend/controllers/library.controller.ts:284`
`/library/artists/search?genre_id=11&q=Bu&q=lt` → `req.query.q` is `['Bu','lt']` → `(req.query.q ?? '').trim()` throws TypeError → 500. The integer-checked params coerce arrays to NaN and 400 cleanly; `q` is the only explosive parameter.

**4. Cascade in-memory filter omits the `missing` constraint** — `apps/backend/services/library-search.service.ts:178`
SQL primary path applies `missing` via EXISTS. When primary returns 0 rows and `passesCascadeGate(q)` is true, `runCascade` filters on on_streaming/genres/formats/rotation_bins but not `missing`. `TaggedLibraryViewEntry` has no `date_lost`/`date_found` to inspect, so cascade results from CTA/LML leak through to the librarian-facing 'missing' view as if they were missing. The new `rotation_bins AND missing both apply` test only asserts `rotation_bin`, masking this. Easiest fix: skip the cascade when `params.missing === true`.

### High — silent data loss / corruption on PATCH

**5. `updateAlbumInDB` SETs every column unconditionally** — `apps/backend/services/library.service.ts:1272`
The handler defaults omitted optional fields (`disc_quantity` → 1, `alternate_artist_name` → null) and `updateAlbumInDB` writes them unconditionally. A partial PATCH that fixes only a typo in `album_title` will overwrite a 2-disc album to disc_quantity=1 and wipe a curator-set `alternate_artist_name` (a load-bearing V/A filing-display surface — see `project_librarian_va_invariant`). Either only `SET` fields the body explicitly provided, or use PUT semantics and require every optional field.

**6. Empty `label: ''` wipes existing `label_id`** — `apps/backend/controllers/library.controller.ts:527`
Missing-params guard only rejects `=== undefined`, so `label: ''` slides past. `if (label_id === undefined && body.label)` is false (empty string is falsy), so `createLabel` is skipped, `label_id` stays `undefined`, and `updateAlbumInDB` writes `label_id: undefined` which drizzle maps to NULL. The FK to a long-stable label is destroyed without any signal.

**7. `updateAlbum` doesn't regenerate `code_number` when `artist_id` changes** — `apps/backend/services/library.service.ts:1272`
No `UNIQUE(artist_id, code_number)` exists. `addAlbum` calls `generateAlbumCodeNumber(artist_id)` to assign the per-artist sequence; `updateAlbum` carries the old number over. Re-attributing an album to an artist that already has that code_number produces two rows with the same (code_letters, code_artist_number, code_number) — librarian V/A invariant violation, silently filed.

**8. Empty `album_title: ''` persists silently** — `apps/backend/controllers/library.controller.ts:498`
Same `=== undefined`-only guard. `''.trim() === ''` and the NOT NULL is satisfied by empty string. addAlbum has the same hole; updateAlbum exposes it on long-stable rows. Either reject empty trimmed strings at the controller, or `CHECK (album_title <> '')` at the schema.

### Medium

**9. `missing=false` is a silent no-op** — `apps/backend/controllers/library.controller.ts:645` vs services L342
Controller parses to `missing: false` and forwards; `buildFilterClause` only branches on `missing === true`. A UI toggle that round-trips both states will silently disable filtering when set to `false`. Either reject `missing=false` as 400 or implement the inverse NOT-EXISTS arm.

**10. `updateAlbum` creates an orphan label row on the 404 path** — `apps/backend/controllers/library.controller.ts:524`
`labelsService.createLabel(body.label)` runs before `updateAlbumInDB`. If the album doesn't exist (or another row is concurrently deleted), the label persists. Enumerating album IDs against this endpoint bulk-pollutes the `labels` table. Move the album existence/lookup before the label upsert.

**11. `createLabel` un-trimmed vs `library.label` trimmed — labels duplication** — `apps/backend/controllers/library.controller.ts:527`
`createLabel('  Drag City  ')` inserts a padded `labels.label_name`; `library.label` then writes `'Drag City'`. A later submission with the trimmed form misses the ON CONFLICT and creates a second labels row. `addAlbum` is internally consistent (both writes un-trimmed); only `updateAlbum` desyncs. Trim before `createLabel`.

**12. `updateAlbum` doesn't refresh streaming / artwork / `canonical_entity`** — `apps/backend/controllers/library.controller.ts:493`
`addAlbum` (L93–135) fires `checkStreamingAvailability`, `lookupMetadata`, and the canonical-entity resolver. `updateAlbum` doesn't. When a librarian fixes an album's artist or title, `library.on_streaming` / `library.artwork_url` / `library_identity` / `canonical_entity_id` stay bound to the OLD identity. `metadata_attempt_at`-driven drain jobs don't repair this because the row's attempt-at is already set. Either invalidate the enrichment columns on metadata-affecting fields or fire the same allSettled enrichment.

**13. `disc_quantity` lacks type/range validation** — `apps/backend/controllers/library.controller.ts:534`
`Math.max(1, 'abc') = NaN`, `Math.max(1, 1.5) = 1.5`, `Math.max(1, 40000) = 40000` — all reach PG smallint and fail with `invalid input syntax` / `value out of range` → 500. `Number.isInteger` + range check at the controller, mirroring `genre_id`.

**14. `searchArtistsInGenre` ILIKE prefix doesn't escape `%` / `_`** — `apps/backend/services/library.service.ts:1149`
`q='%a'` (2 chars passes the length check) builds pattern `'%a%'` and returns any artist whose name contains 'a'. `q='__'` → `'__%'` matches any 2+ char artist. The autocomplete shorts its prefix-search contract; future planner shortcuts on trigram will be bypassed. Same shape exists in `searchLabels` (pre-existing). Escape `%`, `_`, `\` in the bound prefix.

**15. `rotation_bins` filter inherits row duplication from `library_artist_view`** — `apps/backend/services/library-search.service.ts:341`
The rotation table comment (`shared/database/src/schema.ts:534`) explicitly permits "multiple active rows per (album_id, rotation_bin)" (re-bins, re-promotes). The view LEFT JOINs rotation without distinctOn, so an album with two unkilled rotation rows appears twice. Pre-existing in the view, newly surfaced by the `inArray(rotation_bin)` filter and the `COUNT(*)` total. `distinctOn(library_artist_view.id)` on the projection or a subquery that picks one rotation row per album.

### Also worth a quick look

- Client-supplied `body.label_id` isn't validated against the labels table — a stale/guessed integer bypasses `createLabel` and surfaces as PG 23503 → 500.
- `body.label_id: null` (vs `undefined`) plus a non-empty `body.label` skips `createLabel` AND writes label_id=NULL with non-empty `library.label` text — desync.
- `searchArtistsInGenre` returns `{ artists: [] }` for any positive `genre_id` regardless of whether the genre actually exists — silent empty results for stale dropdown IDs.
- `recordActivity`'s new SELECT-then-INSERT adds a DB round-trip to every authenticated request (including all `proxy_route.use(trackActivity)` endpoints). The previous FK violation also surfaced anonymous/orphaned user IDs to Sentry; the silent no-op masks that signal. A catch-on-23503 would preserve both the latency floor and the visibility.
- No integration tests cover `PATCH /library/:id` at all — every error branch and the response shape ship unverified.
- `api.yaml` doesn't declare `genres`/`formats`/`missing`/`rotation_bins`, `PATCH /library/{id}`, or `GET /library/artists/search` — codegen drift across dj-site / iOS / Android.

apps/backend/services/library.service.ts
```diff
@@ -1399,18 +1443,129 @@ export const generateArtistNumber = async (code_letters: string, genre_id: numbe
...
+    'code_number',
+  ] as const) {
+    if (updates[key] !== undefined) set[key] = updates[key];
+  }
+  // Identity-affecting edits invalidate LML-derived columns: the old
+  // on_streaming / artwork / canonical-entity linkage belongs to the previous
+  // (artist, title) identity. The controller re-fires enrichment after the
+  // write; nulling first means a failed lookup leaves the row visibly
+  // unenriched instead of silently bound to the wrong identity.
+  if (opts?.resetEnrichment) {
```
jakebromberg, line 1681: **[#1 — severity 1]** PATCH re-identification nulls `library.canonical_entity_*` but leaves the stale `library_identity` row, `rotation.discogs_release_id` (tier-1 of `resolveRotationPickerSource`), and `compilation_track_artist` rows intact — silent wrong-tracklist data corruption persists for days.

Librarian PATCHes album X with `artist_id=B` (re-attribution). `updateAlbumInDB(resetEnrichment:true)` nulls `library.canonical_entity_*`. `library_identity` (referenced by the rotation-tracks picker's tier-2 fallback at `library.service.ts:551`) still holds the OLD entity's `discogs_release_id`. `library-identity-consumer`'s SELECT only re-processes rows where `canonical_entity_id IS NOT NULL OR last_verified_at > 7d old`, so neither arm fires post-PATCH — the stale identity persists up to a week. `rotation` rows for X keep the OLD denormalized `artist_name`/`album_title`; `compilation_track_artist` rows still claim track-level credits for the OLD album. The dj-site rotation tracks picker renders the OLD album's tracklist for the renamed album.

Wants an `invalidateAlbumIdentity(albumId)` helper that nulls the library row's enrichment **and** marks `library_identity` stale (delete the row, or backdate `last_verified_at`), called from PATCH.

apps/backend/controllers/library.controller.ts
```diff
@@ -470,6 +513,236 @@ const parseAlbumId = (rawId: string): number => {
...
+        throw new WxycError('label_id must be a positive integer or null', 400);
+      }
+      // Validate against the labels table so a stale/guessed id surfaces as
+      // 400 instead of a PG 23503 → 500.
+      const labelRow = await labelsService.getLabelById(body.label_id as number);
+      if (!labelRow) {
+        throw new WxycError('label_id does not reference an existing label', 400);
+      }
+      updates.label_id = labelRow.id;
+      updates.label = trimmedLabel ?? labelRow.label_name;
```
jakebromberg, line 688: **[#2 — severity 2]** `{label_id: N}` alone silently overwrites a custom `library.label` string with `labels[N].label_name`, and `{label, label_id}` together persists the user-supplied label string even when it disagrees with `labels[N].label_name` — the partial-update contract the PR explicitly claims (for issues 5–8/10–13) is broken on both label paths.

Example 1: album has `library.label='Drag City (US)'` and `label_id=5` with `labels[5].label_name='Drag City'`. Client PATCHes `{label_id: 5}` (re-affirming the linkage). Line 654: `updates.label = trimmedLabel ?? labelRow.label_name` → 'Drag City'. Custom textual label silently lost.

Example 2 (inverse): PATCH `{label_id: 5, label: 'Custom Display'}` persists `library.label='Custom Display'` AND `label_id=5` (which resolves to 'Drag City'). `getAlbumFromDB` returns 'Custom Display'; a JOIN through `labels.label_id` returns 'Drag City'. Two views of the same album disagree based on column choice, with no validator forcing them to agree.

apps/backend/controllers/library.controller.ts
```diff
@@ -470,6 +513,236 @@ const parseAlbumId = (rawId: string): number => {
...
+ */
+async function enrichAlbumAfterIdentityChange(
+  albumId: number,
+  displayArtistName: string,
+  albumTitle: string,
+  canonicalArtistName: string | null
+): Promise<void> {
+  if (!displayArtistName) return;
+
+  const [streamingResult, artworkResult] = await Promise.allSettled([
```
jakebromberg, line 744: **[#3 — severity 3]** PATCH `/library/:id` awaits `enrichAlbumAfterIdentityChange` on every identity-changing edit; `checkStreamingAvailability` inside that helper passes no `timeoutMs` so it inherits the LML client's 30 s default (`shared/lml-client/src/index.ts:66` — `TIMEOUT_MS = 30000`) — librarian PATCH UI can hang up to 30 s.

Librarian PATCHes `{album_title: 'Corrected'}` — `identityChanged===true`. Controller awaits `enrichAlbumAfterIdentityChange` → `Promise.allSettled` with `checkStreamingAvailability(artist, title)`. `lmlFetch`'s `effectiveTimeoutMs` is `timeoutMs ?? TIMEOUT_MS = 30000`. If iTunes/Spotify probe stalls, the PATCH response is held the full 30 s — brushing the 35 s app server timeout. `fireAndForgetCanonicalEntity` at line 743 is correctly non-awaited; the streaming+artwork branches should follow the same fire-and-forget pattern (the row write already committed; both branches are best-effort).

apps/backend/services/library.service.ts
```diff
@@ -1399,18 +1443,129 @@ export const generateArtistNumber = async (code_letters: string, genre_id: numbe
...
+    set.on_streaming = null;
+    set.artwork_url = null;
+    set.canonical_entity_id = null;
+    set.canonical_entity_confidence = null;
+    set.canonical_entity_resolved_at = null;
+  }
+  const result = await db.update(library).set(set).where(eq(library.id, album_id)).returning({ id: library.id });
```
jakebromberg, line 1688: **[#4 — severity 4]** `resetEnrichment:true` unconditionally nulls `on_streaming`/`artwork_url`/`canonical_entity_*` before LML runs; a transient LML failure (documented `on_streaming: null` 'do not persist / retry later' sentinel) leaves the row permanently de-enriched — no job under `/jobs` targets `library.on_streaming IS NULL` for repair.

Album X has `on_streaming=true` (resolved at `addAlbum` time, drives `/library/query?on_streaming=false` 'Browse Exclusive Albums'). Librarian PATCHes `{album_title: 'Corrected'}` — typo fix. `updateAlbumInDB` nulls `on_streaming` + `artwork_url` + `canonical_entity_*`. `enrichAlbumAfterIdentityChange` awaits `checkStreamingAvailability` — LML returns `{ on_streaming: null, errored_sources: ['discogs'] }`. Controller honors the sentinel at line 720 and skips the write. Row stays `on_streaming=null`. `library-etl` repairs only from tubafrenzy on legacy syncs; `flowsheet-metadata-backfill` targets flowsheet rows, not library; grep across `/jobs` finds no `library.on_streaming IS NULL` repair. The typo fix silently destroys a correct enrichment value with no recovery path.

apps/backend/controllers/library.controller.ts
```diff
@@ -470,6 +513,236 @@ const parseAlbumId = (rawId: string): number => {
...
+      budgetMs: LIBRARY_LML_BUDGET_MS,
+      caller: 'library-update-album',
+      warm_cache: true,
+      requireSearchType: 'direct',
+    }),
+  ]);
+
+  if (streamingResult.status === 'fulfilled' && streamingResult.value.on_streaming !== null) {
+    try {
+      await libraryService.updateOnStreaming(albumId, streamingResult.value.on_streaming);
```
jakebromberg, line 756: **[#5 — severity 5]** `updateOnStreaming` (`library.service.ts:804`, outside this diff) writes unconditionally — no `isNull(on_streaming)` predicate — so concurrent identity-changing PATCHes can persist `on_streaming` resolved from a stale identity. `updateArtworkUrl` (`library.service.ts:821`) guards exactly this race with `isNull(artwork_url)`; `updateOnStreaming` does not.

Operator A: PATCH `/library/X { artist_id: A }` — `resetEnrichment` nulls `on_streaming`, then `enrichAlbumAfterIdentityChange` awaits `checkStreamingAvailability(A's name, title)` and writes `true`. Concurrently Operator B: PATCH `/library/X { artist_id: B }` — re-nulls `on_streaming`, awaits B's `checkStreamingAvailability`. If A's slower network response lands AFTER B's write completes, the final row has `artist_id=B` but `on_streaming=true` sourced from A's identity. The same `isNull` guard `updateArtworkUrl` uses belongs on `updateOnStreaming`.

apps/backend/services/library-search.service.ts
```diff
@@ -176,6 +201,11 @@ export async function searchLibrary(
   // LML's `Semaphore(5) + TokenBucket(50/min)` chokepoint; pagination beyond
   // page 0 stays empty so clients don't scroll a bounded fallback list.
   if (params.page !== 0) return { results, total };
+  // Cascade rows (`TaggedLibraryViewEntry`) carry no date_lost/date_found, so
+  // neither `missing=true` nor `missing=false` can be honored in-memory —
+  // skip the cascade entirely rather than leak CTA/LML rows into the
+  // librarian-facing missing view.
+  if (params.missing !== undefined) return { results, total };
```
jakebromberg, line 335: **[#6 — severity 6]** Cascade is skipped whenever `params.missing` is set, including `missing=false` — a dj-site default toggle for the catalog browse view loses all CTA/LML typo-suggestion fallback hits.

Librarian queries `GET /library/query?q=Bioluminescnce&missing=false` (typo for Bioluminescence, with the 'hide missing' default UI toggle on). DB query returns 0. Line 208 short-circuits — `if (params.missing !== undefined) return { results, total };` — cascade never runs. Pre-PR (or with the toggle off) the cascade would return the LML/CTA suggestion.

The PR's comment 'cascade rows carry no `date_lost`/`date_found`' is genuinely a problem for `missing=true` (unknown state could be missing), but for `missing=false` the safe reading is 'cascade rows are by definition not in the missing pile' — the skip should be one-arm, not all-arm.

apps/backend/services/library-search.service.ts
```diff
@@ -202,8 +232,19 @@ async function runCascade(params: LibraryQueryParams, q: string): Promise<AlbumS
...
+    if (params.genres !== undefined && params.genres.length > 0 && !params.genres.includes(row.genre_name)) {
+      return false;
+    }
+    if (params.formats !== undefined && params.formats.length > 0 && !params.formats.includes(row.format_name)) {
+      return false;
+    }
+    if (
+      params.rotation_bins !== undefined &&
+      params.rotation_bins.length > 0 &&
+      (row.rotation_bin === null || !params.rotation_bins.includes(row.rotation_bin as CatalogRotationBin))
```
jakebromberg, line 371: **[#7 — severity 7]** Cascade fallback (`searchLibraryByTrackRaw` and `searchLibraryByCTARaw`) LEFT JOINs rotation without dedup; for an album with multiple active rotation rows, the in-TS Map collapse keeps a non-deterministic surviving `rotation_bin` — combined with the new `rotation_bins` filter, this can false-exclude albums actually IN the requested bin.

Album X is in active H+M rotation and matches the CTA cascade for query 'Foo' (SQL path returns 0 rows). `searchLibraryByCTARaw` at `library.service.ts:1916-2012` LEFT JOINs rotation, emitting two rows for X (`bin=H` and `bin=M`). Outer ORDER BY at `library.service.ts:1980` sorts only by `library_rank, cta_track_title` — rotation_bin order is undefined. TS dedup at `library.service.ts:1989-2002` (`byLibraryId.set` first-wins) keeps a non-deterministic bin per album. Same shape in `searchLibraryByTrackRaw`'s `rowsByLegacyId` map at `library.service.ts:1870` where `Map.set` silently overwrites. User searches `?q=Foo&rotation_bins=H`; if the M row survived, `runCascade`'s filter on this line evaluates `!['H'].includes('M')` → true, X filtered out despite being in heavy rotation.

apps/backend/controllers/library.controller.ts
```diff
@@ -470,6 +513,236 @@ const parseAlbumId = (rawId: string): number => {
...
+      throw new WxycError('disc_quantity must be an integer between 1 and 99', 400);
+    }
+    updates.disc_quantity = body.disc_quantity;
+  }
+
+  if (body.format_id !== undefined) {
+    if (!Number.isInteger(body.format_id) || body.format_id < 1) {
+      throw new WxycError('format_id must be a positive integer', 400);
+    }
+    updates.format_id = body.format_id;
```
jakebromberg, line 620: **[#8 — severity 8]** PATCH validates `format_id` is a positive integer but never checks the row exists in `format` — stale dropdown IDs surface as PG 23503 → 500, inconsistent with `label_id` (validated via `getLabelById` at line 649) and `artist_id` (validated via `getArtistNameById` at line 601).

Librarian's stale dj-site formats dropdown POSTs PATCH `/library/X {format_id: 99999999}`. Passes `Number.isInteger` and `≥1`. `updateAlbumInDB` writes `format_id=99999999` → PG raises 23503 on FK `library_format_id_format_id_fk` → uncaught → 500 with no useful client-actionable message. No `formatExists` helper exists in `library.service.ts`; the PR's review-issue framing addresses label/artist symmetry but leaves format behind.

apps/backend/services/library.service.ts
```diff
@@ -1399,18 +1443,129 @@ export const generateArtistNumber = async (code_letters: string, genre_id: numbe
...
+export const albumCodeNumberTaken = async (
+  artist_id: number,
+  code_number: number,
+  exclude_album_id: number
+): Promise<boolean> => {
+  const rows = await db
+    .select({ id: library.id })
+    .from(library)
+    .where(
+      and(eq(library.artist_id, artist_id), eq(library.code_number, code_number), ne(library.id, exclude_album_id))
```
jakebromberg, line 1724: **[#9 — severity 9]** `albumCodeNumberTaken` + `generateAlbumCodeNumber` + `updateAlbumInDB` is a check-then-write TOCTOU window with no DB-level unique index on `(artist_id, code_number)` (verified at `schema.ts:404-419`); two concurrent re-attributions to the same target artist can both regenerate the same `code_number`.

PATCH-A `/library/X { artist_id: A }` at T0 reads `existing.code_number=5`; `albumCodeNumberTaken(A,5,X)` returns false (max under A is 4). Concurrent POST `/library { artist_id: A, ... }` at T0+10ms runs `generateAlbumCodeNumber(A)` → also returns 5 → INSERT `(artist_id=A, code_number=5)`. PATCH-A's UPDATE at T1 writes `(X.artist_id=A, code_number=5)`. Both albums now share `(artist_id=A, code_number=5)`; no constraint catches it. `addAlbum` has the same race shape — the new PATCH path widens the contention surface for librarian editing.

apps/backend/controllers/library.controller.ts
```diff
@@ -470,6 +513,236 @@ const parseAlbumId = (rawId: string): number => {
...
+      throw new WxycError('album_title must be a non-empty string', 400);
+    }
+    updates.album_title = body.album_title.trim();
+  }
+
+  if ('alternate_artist_name' in body) {
+    if (body.alternate_artist_name !== null && typeof body.alternate_artist_name !== 'string') {
+      throw new WxycError('alternate_artist_name must be a string or null', 400);
+    }
+    updates.alternate_artist_name = body.alternate_artist_name?.trim() || null;
```
jakebromberg, line 606: **[#10 — severity 10]** `alternate_artist_name` silently coerces whitespace-only strings to null via `body.alternate_artist_name?.trim() || null` even though it passes `typeof === 'string'` validation — asymmetric with `album_title` (line 562) which 400s on whitespace-only, and the silent clear wipes long-stable overrides AND triggers spurious enrichment reset.

Existing row has `alternate_artist_name='Stevie Wonder'` (a stable override). Librarian PATCHes `{alternate_artist_name: '   '}` (accidental space-only paste from CMS). `typeof === 'string'` passes. `updates.alternate_artist_name = '   '.trim() || null = null`. Line 671-672: `(null ?? null) !== ('Stevie Wonder' ?? null)` → true → `identityChanged=true` → `resetEnrichment` nulls `on_streaming`/`artwork_url`/`canonical_entity_id`/`canonical_entity_confidence`/`canonical_entity_resolved_at` + re-fires LML against the canonical (now alternate-less) artist name. The librarian wipes both a long-stable override AND its bound enrichment. Either 400 on whitespace-only (mirror `album_title`) or document the silent normalization.

apps/backend/controllers/library.controller.ts
```diff
@@ -51,6 +51,11 @@ export const addAlbum: RequestHandler = async (req: Request<object, object, NewA
   ) {
     throw new WxycError('Missing Parameters: album_title, label, genre_id, format_id, artist_name, or artist_id', 400);
   }
+  // '' satisfies the NOT NULL constraint but is never a valid title — reject
+  // before it lands in the catalog (PR #1154 review issue 8).
+  if (typeof body.album_title !== 'string' || body.album_title.trim() === '') {
```
jakebromberg, line 59: **[#11+#12 — severities 11 & 12]** This added non-empty title check is good, but `addAlbum` and `updateAlbum` still diverge on two related rules — both asymmetries are reachable from the dj-site.

**(#11) Trim asymmetry.** `addAlbum` stores `body.album_title` untrimmed (line 89) and passes `body.label` raw to `createLabel` (line 80); `updateAlbum` trims both. Two consequences:
- Title: legacy row written via `addAlbum` carries `album_title='Confield '` (trailing space). Librarian PATCHes `{album_title: 'Confield'}` (typo fix). `updates.album_title='Confield'`. Line 671: `identityChanged = ('Confield' !== 'Confield ')` = true → `resetEnrichment` nulls all four enrichment cols + re-fires LML for what was semantically a no-op edit.
- Label: POST `/library {label: '  Drag City  ', ...}` → `createLabel('  Drag City  ')` (untrimmed) → labels row inserted with `label_name='  Drag City  '` (the UNIQUE on `label_name` is whitespace-sensitive). Later PATCH `{label: 'Drag City'}` → `createLabel('Drag City')` (trimmed) → `onConflictDoNothing` on `label_name` UNIQUE does NOT match → NEW labels row inserted. Two semantically identical labels split the catalog.

**(#12) disc_quantity validation asymmetry.** `addAlbum` performs no `disc_quantity` range/type validation; `updateAlbum` enforces integer in `[1, 99]`. POST `disc_quantity=0` succeeds (no `addAlbum` check). PATCH `/library/X {disc_quantity: 0, label: 'X Records'}` (a UI that always sends all fields on save) — controller validates `disc_quantity` at line 576 → 400 'disc_quantity must be an integer between 1 and 99'. The librarian's edit to a completely unrelated field is blocked by a validation that was never applied at insert time.

apps/backend/controllers/library.controller.ts
```diff
@@ -470,6 +513,236 @@ const parseAlbumId = (rawId: string): number => {
...
+    }
+    const effectiveArtistId = body.artist_id ?? existing.artist_id;
+    const effectiveGenreId = body.genre_id ?? existing.genre_id;
+
+    const canonical_artist_name = await libraryService.getArtistNameById(effectiveArtistId);
+    if (!canonical_artist_name) {
+      throw new WxycError('Artist not found', 404);
+    }
+
+    const inGenre = await libraryService.artistExistsInGenre(effectiveArtistId, effectiveGenreId);
```
jakebromberg, line 640: **[#13 — severity 13]** PATCH with an unknown `genre_id` returns a misleading 'Artist is not catalogued in the selected genre' (400) because `artistExistsInGenre` returns false for stale genre IDs without distinguishing the absent-genre case; the companion GET `/library/artists/search` at line 312 correctly calls `genreExists` and 404s.

Librarian's stale dj-site dropdown PATCHes `{genre_id: 99999}` (genre deleted server-side, or hand-crafted typo). `Number.isInteger` and `≥1` pass at line 595. `artistExistsInGenre(existing.artist_id, 99999)` runs `WHERE artist_id=X AND genre_id=99999` → 0 rows → controller throws 'Artist is not catalogued in the selected genre' (400). The librarian sees a baffling error — they verified the artist IS in their other catalog work. Add a `genreExists` check (already exported at `library.service.ts:1625`) before `artistExistsInGenre` when `body.genre_id !== undefined`.

apps/backend/controllers/library.controller.ts
```diff
@@ -470,6 +513,236 @@ const parseAlbumId = (rawId: string): number => {
...
+    if (artworkUrl) {
+      try {
+        await libraryService.updateArtworkUrl(albumId, artworkUrl);
+      } catch (e) {
+        console.warn('Failed to persist artwork URL after album update:', (e as Error).message);
+      }
+    }
+  }
+
+  fireAndForgetCanonicalEntity(albumId, canonicalArtistName, albumTitle);
```
jakebromberg, line 777: **[#14 — severity 14]** `enrichAlbumAfterIdentityChange` burns two distinct LML coordinator cache keys per identity-changing PATCH whenever `alternate_artist_name` differs from the canonical `artists.artist_name` — the artwork branch uses `effectiveAlternate || canonicalArtistName`, the canonical-entity branch (this `fireAndForgetCanonicalEntity` call) uses `canonicalArtistName` only, so the canonical-entity call cannot share the artwork call's cached response.

PATCH `{album_title: 'New Title'}` on a row where `alternate_artist_name='Stage Name'` and `artists.artist_name='Real Name'`. Line 711-718: `lookup('Stage Name', 'New Title', ...)` for streaming+artwork (key A). This line: `fireAndForgetCanonicalEntity` → `lookup('Real Name', 'New Title', ...)` for canonical-entity (key B). Two distinct cache keys → two LML wire calls on miss instead of one. Multiplies LML rate-limit consumption (`Semaphore(5)`, `TokenBucket(50/min)`) on every alternate-credited identity edit, when one canonical-form lookup would suffice.

apps/backend/app.yaml
```diff
@@ -856,6 +856,181 @@ paths:
         '200':
           description: Successfully added album
 
+  /library/query:
```
jakebromberg, line 984: **[#15 — severity 15]** Two compounding OpenAPI / wxyc-shared codegen gaps in the new endpoints.

**Missing `security:` on `/library/query`.** Every sibling protected route added in this diff (`PATCH /library/{id}`, `/library/{id}/missing`, `/library/{id}/found`, `GET /library/artists/search`) declares `BearerAuth`, but `/library/query` doesn't. An `openapi-typescript-codegen` consumer emits a typed client where `/library/query` has no Authorization header injection — runtime requests 401 from `requirePermissions({catalog:['read']})` at `library.route.ts:15`. The swagger UI at `/api-docs` also misleads human readers into thinking the endpoint is public.

**PATCH response schema is stale.** The new PATCH `/library/{id}` operation (around line 144 of this diff) refs `#/components/schemas/Album`, but `getAlbumFromDB` (called by PATCH) now projects fields the existing `Album` schema doesn't declare: `disc_quantity`, `alternate_artist_name`, `artist_id`, `genre_id`, `format_id`, `label`, `album_artist`. Generated TypeScript types miss `disc_quantity` and `alternate_artist_name` — the very fields PATCH is designed to round-trip. Both fields are documented in the request body but absent from the response schema. Update `Album` (or add an `AlbumDetail` schema) to match `getAlbumFromDB`'s actual projection.

jakebromberg (approved):

PR #1558, opened 2026-07-08
feat(auth): token-based onboarding completion for new DJs

JacksonMeade:
## Summary
- Add `POST /auth/wxyc/complete-onboarding` to atomically validate an invite setup token (or authenticated incomplete session), set the chosen password, update optional profile fields, and mark `hasCompletedOnboarding`.
- Make `password` optional on admin provision-user; generate an unguessable server-side bootstrap credential when omitted.
- Point provision invite emails at `/onboarding` instead of `/login`.

## Why
The shared `NEXT_PUBLIC_ONBOARDING_TEMP_PASSWORD` conflicted with invite-email password resets, causing better-auth `INVALID_PASSWORD` when onboarding called `changePassword` with the stale temp password.

## Deploy / pairing
Merge and deploy **before** the companion dj-site PR (https://github.com/WXYC/dj-site/pull/827). The new endpoint is backward compatible; old clients can keep using temp password until dj-site ships.

After both are live, remove `NEXT_PUBLIC_ONBOARDING_TEMP_PASSWORD` / `PREVIEW_NEXT_PUBLIC_ONBOARDING_TEMP_PASSWORD` repo variables.

## Test plan
- [ ] Unit tests: `complete-onboarding.test.ts`, updated `provision-user.test.ts`
- [ ] CI auth unit/integration suite
- [ ] Manual: provision user without password → invite link lands on `/onboarding?token=…` → complete onboarding → login with chosen password


jakebromberg (commented):
### 🔴 High — `provision-user` now hard-rejects `password`, causing a provisioning outage during the deploy window

`apps/auth/app.ts:221`

```ts
if ('password' in body) {
  return res
    .status(400)
    .json({ error: 'password must not be supplied; users set their password via the invite onboarding flow' });
}
```

The currently-deployed dj-site `main` sends `password` on **both** admin create paths, and both POST directly to `/auth/admin/provision-user` (there are no `authClient.admin.createUser` calls on `main`):

- `src/components/experiences/modern/admin/roster/RosterTable.tsx:93` — single-account creation
- `src/components/experiences/modern/admin/roster/ImportCSVModal.tsx` — CSV bulk import

Because the contract flipped from *password required* → *password forbidden*, **no deploy order is clean**:

- **Backend first** (as this PR's deploy note prescribes): the live old dj-site keeps sending `password` → `400 "password must not be supplied"`. Admin single-account creation **and** CSV roster import break until the companion dj-site PR ships.
- **dj-site first**: the new dj-site omits `password` → the still-old backend returns `400 "Missing required fields: password"`.

Either way, admin account creation fails for the whole rollout window, so the "backward compatible" claim doesn't hold for the provisioning endpoint.

**Suggested fix:** tolerate-and-ignore a supplied `password` (optionally log a deprecation warning) instead of returning 400. `provisionUser` already generates its own unguessable bootstrap credential and ignores the supplied value, so accepting-and-dropping the field makes the change order-independent and genuinely backward compatible.

apps/auth/app.ts
```diff
@@ -209,21 +247,23 @@ app.post('/auth/admin/provision-user', async (req, res) => {
...
+    const missing = ['email', 'username', 'name', 'organizationSlug', 'role'].filter(
+      (field) => !body[field] || typeof body[field] !== 'string'
     );
     if (missing.length > 0) {
       return res.status(400).json({ error: `Missing required fields: ${missing.join(', ')}` });
     }
+    if ('password' in body) {
+      return res
+        .status(400)
+        .json({ error: 'password must not be supplied; users set their password via the invite onboarding flow' });
```
jakebromberg, line 260: This breaks the stated deploy ordering. The deployed dj-site sends `password` from both provisioning paths (`RosterTable.tsx` `password: newAccount.temporaryPassword` and `ImportCSVModal.tsx` `password: tempPassword`), so from the moment this deploys until dj-site#827 ships, every admin create-account and CSV import 400s here — the PR body's "backward compatible; old clients can keep using temp password" doesn't hold for provisioning. A loud 400 arguably beats silently creating accounts the old temp-password onboarding can't complete (the bootstrap credential is unguessable), but the window needs coordinating: pair-deploy, or accept-and-ignore `password` behind a deprecation log until #827 is live.

apps/auth/complete-onboarding.ts
```diff
@@ -0,0 +1,192 @@
...
+      throw new CompleteOnboardingError(
+        error.statusCode ?? 400,
+        error.body?.message ?? 'Invalid or expired setup token',
+        typeof error.body?.code === 'string' ? error.body.code : 'INVALID_TOKEN'
+      );
+    }
+    throw error;
+  }
+
+  await markOnboardingComplete(userId, { realName, djName, emailVerified: true });
```
jakebromberg, line 171: Not atomic, despite the module doc's "atomically": by this point `resetPassword` has consumed the token (better-auth's `consumeVerificationValue` deletes the row) and set the password. If this `updateUser` throws on a transient DB error, the user gets a 500 with password set, `hasCompletedOnboarding=false`, `emailVerified=false` — and retrying the form fails with INVALID_TOKEN because the token is burned. Recovery is sign-in attempt → blocked by `requireEmailVerification` → verification email → auto-sign-in → session-mode completion, with the originally submitted realName/djName lost. Worth a retry around the updateUser, or at least a distinguishable error steering the user to the sign-in path, plus softening the docblock.

apps/auth/complete-onboarding.ts
```diff
@@ -0,0 +1,192 @@
...
+/**
+ * Session completion for a signed-in incomplete user (the `/login?incomplete=true`
+ * form). They authenticated with a password they already know, so only profile
+ * fields and the completion flag change.
+ */
+export async function completeOnboardingWithSession(
+  headers: Headers,
+  input: { realName?: string; djName?: string }
+): Promise<CompleteOnboardingResult> {
+  const session = await auth.api.getSession({ headers });
```
jakebromberg, line 185: `getSession` here also returns anonymous-plugin sessions (the iOS clients use anonymous auth). An anonymous user row has `hasCompletedOnboarding=false`, so an anonymous session can "complete onboarding" on its own row and write arbitrary realName/djName. Damage is confined to the anonymous row, but it's an unintended public write path and pollutes anything keying off these fields. Consider `if (session.user.isAnonymous) throw new CompleteOnboardingError(403, ...)`.

jakebromberg (commented):
Reviewed the full diff plus the deployed dj-site (`origin/main`) and the installed better-auth 1.6.20 source. 9 inline comments below; two findings don't anchor to changed lines so they live here.

**1. `scripts/resend-setup-emails.ts:61` still sends `redirectTo: ${frontendUrl}/login`.** provisionUser now points invites at `/onboarding`, so the resend script drifts from the new flow. After dj-site#827 ships, a resent invite lands the DJ on the login reset form, which consumes the token via plain `resetPassword`: the password gets set, but `hasCompletedOnboarding` and `emailVerified` are never flipped. With `requireEmailVerification: true` their first sign-in is then blocked pending a second verification-email round trip, and profile fields are never collected. One-line fix to `/onboarding`.

**2. CLAUDE.md Auth Server section is now stale** (~line 90): no bullet for the new public `/auth/wxyc/complete-onboarding`, the documented `POST /auth/admin/provision-user` contract now rejects `password` with no note about the invite-onboarding flow, and the test-only endpoint list omits `/auth/test/reset-incomplete-user` (the in-code log enumeration was updated; the doc list wasn't).

Verified non-findings, for the record: the `reset-password:${token}` → userId resolution, expiry handling, and `APIError.statusCode` mapping are all correct against better-auth 1.6.20 (`consumeVerificationValue` enforces expiry itself; the drizzle schema returns real `Date`s so the `expiresAt < new Date()` comparison is sound); `express.json()` covers the new route; the rate-limiter registration order is correct; `PASSWORD_RESET_REDIRECT_URL`/`buildResetUrl` does not override the new `/onboarding` callback; `cookieCache` is off and there are no `user.update` databaseHooks that the internalAdapter writes would skip; the `test-incomplete` seed row exists for the new test endpoint; CI is green on the branch.

apps/auth/app.ts
```diff
@@ -316,6 +372,7 @@ if (!isTestEnv) {
 }
 
 app.post('/auth/wxyc/lookup-email', lookupEmailHandler);
+app.post('/auth/wxyc/complete-onboarding', completeOnboardingHandler);
```
jakebromberg, line 393: Schema-first rule: this new public endpoint's request/response shapes exist only as private TS types — `wxyc-shared/api.yaml` (the cross-repo SSOT) has no entry, and auth-surface endpoints are demonstrably in scope there (`/auth/device/*` is specced). dj-site#827 is hand-writing the `token/newPassword/realName/djName` request and `{status,userId,email,username}` response on the client, which is exactly the drift the rule exists to prevent. Worth adding to api.yaml alongside this PR.

apps/auth/complete-onboarding.ts
```diff
@@ -0,0 +1,192 @@
...
+  const verification = await context.internalAdapter.findVerificationValue(identifier);
+
+  if (!verification || verification.expiresAt < new Date()) {
+    throw new CompleteOnboardingError(400, 'Invalid or expired setup token', 'INVALID_TOKEN');
+  }
+
+  return verification.value;
+}
+
+type IncompleteUserRecord = {
```
jakebromberg: This hand-rolled type re-declares fields the canonical `User` type already defines (`shared/database/src/schema.ts:75`, `InferSelectModel`), and the `as IncompleteUserRecord` cast below is untethered from the schema: if `hasCompletedOnboarding` is ever renamed or retyped, the cast keeps compiling, the `=== true` guard goes always-false, and this endpoint silently starts consuming completed users' reset tokens — the exact invariant `assertIncompleteUser` protects. `Pick<User, 'id' | 'email' | 'username' | 'hasCompletedOnboarding'>` fails the build instead (`username` is `string | null` there, so the result builders would emit `user.username ?? undefined`).

apps/auth/complete-onboarding.ts
```diff
@@ -0,0 +1,192 @@
...
+
+const trimOptional = (value: unknown): string | undefined => {
+  if (typeof value !== 'string') return undefined;
+  const trimmed = value.trim();
+  return trimmed.length > 0 ? trimmed : undefined;
+};
+
+async function resolveUserIdFromToken(token: string): Promise<string> {
+  const context = await auth.$context;
+  const identifier = `reset-password:${token}`;
```
jakebromberg, line 71: Verified this matches the installed better-auth 1.6.20 exactly (`password.mjs` stores `reset-password:${token}` with `value` = userId) — but the identifier format is an internal, not public API. A minor bump that changes the prefix or storage encoding turns every invite token into a silent INVALID_TOKEN while the mocked-adapter unit tests stay green. Worth a comment pinning the dependency and an integration test that mints a real token via `requestPasswordReset` and resolves it through this path.

apps/auth/app.ts
```diff
@@ -159,9 +160,46 @@ if (process.env.NODE_ENV === 'development' || process.env.NODE_ENV === 'test') {
...
+  // Reset the seeded test_incomplete user for session-onboarding E2E
+  app.post('/auth/test/reset-incomplete-user', async (req, res) => {
+    try {
+      const userId =
+        typeof (req.body as { userId?: string })?.userId === 'string'
+          ? (req.body as { userId: string }).userId
+          : 'test-incomplete-id-0000000000001';
+      const tempPassword = 'temppass123';
+
+      const { auth } = await import('@wxyc/authentication');
```
jakebromberg: This re-imports and shadows the `auth` this file already statically imports at line 9 — it resolves to the same cached module instance. The dynamic imports of `@wxyc/database`/`drizzle-orm` match the neighboring test endpoints' style, but none of them re-import `@wxyc/authentication`; this can use the module-level `auth` directly.

apps/auth/app.ts
```diff
@@ -159,9 +160,46 @@ if (process.env.NODE_ENV === 'development' || process.env.NODE_ENV === 'test') {
...
   });
 
+  // Reset the seeded test_incomplete user for session-onboarding E2E
+  app.post('/auth/test/reset-incomplete-user', async (req, res) => {
+    try {
+      const userId =
+        typeof (req.body as { userId?: string })?.userId === 'string'
+          ? (req.body as { userId: string }).userId
+          : 'test-incomplete-id-0000000000001';
+      const tempPassword = 'temppass123';
```
jakebromberg: Duplicates `TEMP_PASSWORD = 'temppass123'` from `dev_env/setup-e2e-test-users.ts:18`, which seeds this same user. If that constant ever changes, this endpoint resets the user to a password the E2E setup no longer expects — confusing to trace. Worth sharing the constant.

apps/auth/complete-onboarding.ts
```diff
@@ -0,0 +1,192 @@
...
+        typeof error.body?.code === 'string' ? error.body.code : 'INVALID_TOKEN'
+      );
+    }
+    throw error;
+  }
+
+  await markOnboardingComplete(userId, { realName, djName, emailVerified: true });
+
+  return {
+    status: true,
```
jakebromberg: Byte-identical to the session flow's result block at 157–162. Any response-shape change now has to land twice or the two flows return different shapes from the same endpoint — a one-line `toResult(user)` helper collapses it.

apps/auth/app.ts
JacksonMeade: I think we're going to have synchronization issues like this until we monorepo this thing - because of the onboarding flow, no accounts still using the temp password have ever been logged into - they're less secure, but that's okay. We're going to merge these at practically the same moment, so the intermediate window does not need coordination, particularly not in code.

apps/auth/complete-onboarding.ts
JacksonMeade: Added retry (3 attempts) on updateUser. If it still fails after resetPassword consumed the token, returns 503 with code PROFILE_UPDATE_FAILED and a message steering the DJ to sign in at /login?incomplete=true. Softened the module doc to describe the two-phase flow.

apps/auth/complete-onboarding.ts
JacksonMeade: Rejects anonymous sessions with 403 FORBIDDEN in session mode.

apps/auth/complete-onboarding.ts
JacksonMeade: Replaced with Pick<User, ...> from @wxyc/database.

apps/auth/complete-onboarding.ts
JacksonMeade: Added a comment pinning it to better-auth 1.6.20. Added integration test tests/integration/complete-onboarding-token.spec.js that mints a real token via requestPasswordReset and completes onboarding end-to-end.

apps/auth/app.ts
JacksonMeade: Uses the module-level auth import now.

apps/auth/app.ts
JacksonMeade: Shared via apps/auth/e2e-test-constants.ts (used by the test endpoint and setup-e2e-test-users.ts).

apps/auth/complete-onboarding.ts
JacksonMeade: Added toCompleteOnboardingResult(user)
Both completeOnboardingWithToken and completeOnboardingWithSession use it instead of duplicating the { status, userId, email, username } block

apps/auth/app.ts
JacksonMeade: Added CompleteOnboardingRequest, CompleteOnboardingResponse, CompleteOnboardingError schemas
Added POST /auth/wxyc/complete-onboarding path
Added api-spec tests in tests/api-spec.test.ts
Bumped @wxyc/shared to 1.18.1

jakebromberg (approved):

PR #1679, opened 2026-07-16
feat: GET /album-reviews read endpoint with PII-barrier projection

jakebromberg:
Closes #1677 (PR 3 of 3). Stacked on #1678 — retarget to `main` and rebase after #1678 merges. Contract: WXYC/wxyc-shared#229 / #230.

Read surface for the album-review archive, mirroring the concerts stack:

- **Route**: `routes/album-reviews.route.ts`, mounted at `/album-reviews` with `requirePermissions({})` (anon-JWT read posture — the iOS surface convention). Deliberately not `/reviews`, which ADR 0006 reserves for the in-app Review model.
- **Controller**: hand-rolled param validation (`album_id` all-digits int, `artist` non-empty ≤256 chars with repeated-param rejection, `page`/`limit` defaults 1/50 cap 100), `Promise.all([page, count])`, `{ album_reviews, pagination }` per the wxyc-shared `AlbumReviewsResponse` shape.
- **Service**: the explicit `select({...})` projection is the **PII leak barrier** — `reviewer_raw` and `social_consent_raw` are never selected, so reviewer identity cannot reach the wire regardless of the DTO mapper (the form promised names would not be shared). Shared `buildWhere` for page/count parity; `artist` filter matches `norm_artist = normalizeArtistName(param)` on the `…_norm_artist_idx`; order `submitted_at DESC NULLS LAST, id DESC`. Local `AlbumReviewDTO` alias with the concerts-style TODO to switch to `@wxyc/shared/dtos` once the pin carries wxyc-shared#230's types.
- **Docs**: app.yaml Swagger-display mirror + CLAUDE.md route-table row.

## Tests

Unit (controller 400-families/defaults/caps, route auth wiring, service projection-key-set PII assertion + buildWhere parity + compile-time shape pin against api.yaml), `tests/integration/album-reviews.spec.js` (17 cases: 401, wire-shape equality, PII keys absent while DB rows carry them, filters incl. case/leading-The insensitivity, NULLS-LAST ordering, pagination, validation 400s), and `tests/e2e/album-reviews-pipeline.test.ts` (8 cases: real `runEtl` + writer + link with only the sheet fetch stubbed, run twice for idempotency, then a live GET round-trip with wire-level PII exclusion).

The e2e suite is what surfaced the two driver-boundary bugs now fixed in #1678 (raw-Date setWhere binding, `ANY()` array splat) — it is the first test to run the actual TS writer/link path against real PostgreSQL.

Local gate: typecheck, format, 4,593 unit tests, migrations validator; integration spec 17/17 and e2e 8/8 against the Dockerized stack.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -65,6 +65,7 @@ Express 5 application with these route groups:
 | `/config`       | Public app bootstrap configuration                         |
 | `/proxy`        | iOS proxy endpoints (anonymous auth + rate limit)          |
 | `/library`      | Music library catalog                                      |
+| `/album-reviews` | Form-review archive reads (ADR 0011; anon-JWT, PII-excluding projection) |
 | `/flowsheet`    | V1 flowsheet (legacy)                                      |
 | `/v2/flowsheet` | V2 flowsheet (uses `@wxyc/shared` DTOs)                    |
 | `/djs`          | DJ profiles and management                                 |
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -60,20 +60,20 @@ npm workspaces:
 
 Express 5 application with these route groups:
 
-| Route           | Purpose                                                    |
-| --------------- | ---------------------------------------------------------- |
-| `/config`       | Public app bootstrap configuration                         |
-| `/proxy`        | iOS proxy endpoints (anonymous auth + rate limit)          |
-| `/library`      | Music library catalog                                      |
+| Route            | Purpose                                                                  |
+| ---------------- | ------------------------------------------------------------------------ |
+| `/config`        | Public app bootstrap configuration                                       |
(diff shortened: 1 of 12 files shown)


jakebromberg:
Superseded by WXYC/Backend-Service#2290, which revives these files with `git show` rather than merging or cherry-picking this branch — the history here stays readable.

The substantive change is the auth tier: this PR's `requirePermissions({})` admitted anonymous sessions, which would have published non-consented review bodies to the listener app. The replacement gates on a new `reviews: ['read']` permission held by dj and above. Six other changes came along with it — an `isNotNull(review)` body floor, the e2e anonymous assertion inverted from 200 to 403 plus a new dj path, and five prose sites that asserted the anonymous-auth posture. Details in WXYC/Backend-Service#2289.

PR #2025, opened 2026-08-07
Resolve duplicate call-number slots in library ahead of a uniqueness constraint

jakebromberg:
Closes #2024.

Adds `jobs/library-call-number-dedup`, a one-shot that resolves duplicate call-number slots in `library` so a uniqueness constraint can be added to the columns that address the shelf. Dry-run by default; `--execute` to write. Structural donor is `jobs/artist-unicode-dedup` — same `merge.ts` core + thin `job.ts` CLI split, same per-group transaction and repoint-before-delete discipline.

Nothing in the schema enforces that a call number is used once, and `generateAlbumCodeNumber` reads `MAX(code_number)+1` and inserts without a lock, so two adds under one artist at the same moment already collide with no user error involved. Roughly 770 slots hold more than one release.

## The distinction it turns on

Most collisions are one release entered twice under a differently-decorated title, and merge with no disc moving. The rest are genuinely different releases, where one row is renumbered **and the physical disc has to be relabelled** — so a run emits a librarian worklist, and the constraint waits on that shelf work.

Classifying on title equality gets this wrong by better than 2x. `It's a Party 12"` and `It's a Party cd-single` are one record filed twice; so are a dirty LP and its clean-lyric pressing, and `Volume 1` versus `Volume One`. `classify.ts` strips the decoration that names a *pressing* rather than a *work*, and deliberately preserves volume and part numbers — folding `Ethiopiques vol. 21` into `vol. 22` would merge two real releases and delete a row that has plays.

## Three things that are easy to get backwards

**`genre_id` belongs in the slot key.** Code letters are genre-scoped, so `BO 3` (Electronic) and `Bo 3` (Rock) are different shelves. That used to be implicit in there being two `artists` rows; #1897's dedup merges artist rows globally across genres, so genre is now the only thing distinguishing them. A key of `(artist_id, code_number)` would call those a collision and destroy a correct filing.

**The volume letter compares case-insensitively.** `D` and `d` are one slot. The upstream MySQL catalog folds them under its default collation, so duplicates exist that Postgres — comparing case-sensitively — would read as distinct and let straight through.

**The FK table is derived from the live catalog, not `schema.ts`.** The two disagree: `schema.ts` marks `artist_library_crossreference.library_id` as `cascade` while the database enforces `no action` — one of the four drifted constraints in #2015, and a repoint target here. Reading the *creating* migration is not sufficient either, since `album_metadata.album_id` was created `no action`, dropped with its table, then recreated `cascade`. The integration spec pins every action against `information_schema` so the documented table tracks the database.

## Why the ordering is a data-safety property

Five of the 13 sites referencing `library.id` cascade and two null the reference out. Deleting before repointing would silently destroy rotation history, album metadata, and reviews, and silently unlink plays, with no error raised. The four no-action sites are the only ones that fail loudly — which the integration spec demonstrates directly.

## Verification

CI green on `de337e8e`'s predecessor; re-running on this head. Locally: format clean, 0 lint errors, typecheck clean, 6,616 unit tests. The integration tier exercises the real compiled `merge.cjs` against Postgres — the loser's `album_metadata` and compilation tracks survive a merge attached to the survivor, both unique-key collision paths drop rather than violate, a different genre is not a collision, and a case-differing volume letter is.

## Size

1,583 lines, over the 1,000 guideline. It is all new files with no edits to existing code, and 553 lines are tests. Splitting it would mean landing the merge core without the spec that proves its central invariant, which seemed worse than the overage.

## Not in this PR

- The uniqueness constraint migration. It cannot land until the relabel worklist is worked, and its key must be `(artist_id, genre_id, code_number, COALESCE(code_volume_letters, ''))` — the `COALESCE` because prod is PG 14 and `NULLS NOT DISTINCT` is PG 15+.
- The run itself. This must execute **after** the tubafrenzy ETL stops: `jobs/library-etl` carries `code_number` in its refresh set and would overwrite a renumber on its next pass.

### DIFF
--- a/Dockerfile.library-call-number-dedup
+++ b/Dockerfile.library-call-number-dedup
@@ -1,0 +1,39 @@
+#Build stage
+FROM node:24-alpine AS builder
+
+WORKDIR /library-call-number-dedup-builder
+
+COPY ./package.json ./package-lock.json ./
+COPY ./tsconfig.base.json ./
+COPY ./shared/database ./shared/database
+COPY ./jobs/library-call-number-dedup ./jobs/library-call-number-dedup
+
+RUN npm ci && npm run build --workspace=@wxyc/database --workspace=@wxyc/library-call-number-dedup
+
+#Production stage
+FROM node:24-alpine AS prod
+
+WORKDIR /library-call-number-dedup
+
+COPY ./package* ./
+COPY ./jobs/library-call-number-dedup/package* ./jobs/library-call-number-dedup/
+COPY ./shared/database/package* ./shared/database/
+
+RUN npm install --omit=dev
+
+COPY --from=builder ./library-call-number-dedup-builder/jobs/library-call-number-dedup/dist ./jobs/library-call-number-dedup/dist
+COPY --from=builder ./library-call-number-dedup-builder/shared/database/dist ./shared/database/dist
+
+# Bulk repoint workload. Async commit is safe: the job is idempotent (a merged
+# slot drops out of the `HAVING count(*) > 1` set) and each slot's writes are
+# one transaction. High statement timeout so a slot under live load does not
+# abort mid-merge.
+ENV DB_APPLICATION_NAME=wxyc-library-call-number-dedup
+ENV DB_STATEMENT_TIMEOUT_MS=900000
+ENV DB_SYNCHRONOUS_COMMIT=off
+
+# ENTRYPOINT + empty CMD so docker-level args (e.g. `--execute`) pass through to
+# the job rather than replacing the launcher. node resolves `@wxyc/database` via
+# the standard walk up to /library-call-number-dedup/node_modules.
+ENTRYPOINT ["node", "/library-call-number-dedup/jobs/library-call-number-dedup/dist/job.js"]
+CMD []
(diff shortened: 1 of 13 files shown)


jakebromberg:
Converting to draft. Review surfaced four confirmed defects that make this unsafe to merge as-is:

1. `reviews.album_id` carries `reviews_album_id_unique`, declared here as `uniqueKey: null` — a merge where both rows have a review raises a unique violation and aborts the run.
2. `rotation_active_album_bin_uniq (album_id, rotation_bin) WHERE kill_date IS NULL` is a **partial** unique index, which `repointTarget` cannot express. The obvious fix (`['album_id','rotation_bin']`) would delete killed rotation rows and destroy exactly the history this job promises to preserve, so this needs a real design decision rather than a one-line change.
3. The decoration regex strips bare `7` and `12`, so `Ethiopiques vol. 7` and `vol. 12` fold together — silently merging two different records. The unit test used `vol. 21`/`vol. 22`, whose digits fall outside the stripped set.
4. `sameRelease` treats any prefix relationship as a match with no minimum-length or remainder guard, so `Home`/`Homework` and `Songs of Love`/`Songs of Love and Hate` merge.

All four verified directly against the shipped code and the migration chain.

Root cause for 1 and 2 shipping green: `enforced-fk-actions` pins delete actions but never asserts `uniqueKey`, and its anti-join iterates its own EXPECTED map rather than the query result, so it cannot see a constraint the database has and `FK_TARGETS` omits. That test needs to run in the other direction before the collision handling can be trusted.

Issue #1201, opened 2026-05-29
Cron containers: heartbeat counter + CloudWatch alarm on missing heartbeats


jakebromberg:
## Problem

Cron jobs on the prod EC2 host fail silently in multiple ways:
- ECR auth expires → image pull denied → no container → no exit code (BS#1183).
- Docker daemon is wedged → no container.
- Image was GC'd locally and registry briefly unreachable → no container.
- Cron container starts, then panics during init before any structured log is emitted.

In all these cases nothing exists to look at, nothing pages, and the next person to notice is the human checking a downstream artifact (a queue not draining, a metric not advancing). The two-three-day lag between "cron stopped running" and "someone noticed" is the typical pattern.

Log routing (the sibling issue) makes the failure forensically visible. It does not page anyone.

## Proposal

**Positive monitoring**: each cron emits a heartbeat metric on successful completion. A CloudWatch alarm fires when the heartbeat doesn't advance within the cron's schedule + grace window.

```bash
# At the end of the cron wrapper, after a successful docker run exit:
aws cloudwatch put-metric-data \
  --namespace WXYC/Cron \
  --metric-name HeartbeatSuccess \
  --dimensions CronName=flowsheet-metadata-backfill \
  --value 1 \
  --region us-east-1
```

Per-cron CloudWatch alarms:
- `flowsheet-metadata-backfill` (daily 06:00 UTC) — alarm on 0 datapoints in 30 hours (24 h schedule + 6 h grace).
- `artist-identity-etl-cron` (hourly) — alarm on 0 datapoints in 3 hours.
- `rotation-etl-cron` / `library-etl-cron` / `flowsheet-etl-cron` (every 30 min) — alarm on 0 datapoints in 90 minutes.

Per the org's CloudWatch conventions (see WXYC/CLAUDE.md "CloudWatch Metric & Alarm Conventions"), each metric needs a dimensionless companion if any alarm queries the aggregate. Per-cron alarms query the dimensioned series directly, so no companion needed unless we add a "cron failure rate across the host" composite alarm later.

## Acceptance

- [ ] Cron wrapper script emits `WXYC/Cron/HeartbeatSuccess` with `CronName` dimension on successful exit.
- [ ] CloudWatch alarms for each cron with appropriate threshold (schedule + grace).
- [ ] Alarms route to the existing WXYC SNS topic / oncall channel.
[...]

PR #2066, opened 2026-08-09
feat(legacy-linkage-resolve): give the cron a liveness signal (Sentry monitor + heartbeat + drain check)

jakebromberg:
Closes #2064

`jobs/legacy-linkage-resolve/` has been the only thing repairing `flowsheet.album_id` / `rotation.album_id` after write time since Phase 3 of the tubafrenzy decommission ([WXYC/wiki#88](https://github.com/WXYC/wiki/issues/88), landed in #1858) unscheduled the two ETLs whose tail pass `58a9cd76` lifted it out of. Because a healthy idle run writes nothing and logs `candidates: 0`, a cron that had **stopped running** was byte-identical to one that ran and found nothing — and `run()`'s `captureError` only ever covered "the job ran and threw", never the crontab entry being dropped, the image failing to pull, docker wedging, or the host rebooting.

## What landed: (a) + (b) + (c)

The ticket offered a Sentry cron monitor, a `cronjob_runs` heartbeat, and a cohort-age detector, preferring "(a), or (b)+(c)". All three shipped, in a form that makes each catch something the others miss.

| Signal | Detects | Latency | Survives |
| --- | --- | --- | --- |
| **(a)** `Sentry.withMonitor` check-in | The run did not happen at all | ~40 min | An error-quota exhaustion — check-ins are a separate Sentry quota category |
| **(b)** `cronjob_runs` heartbeat + gap read | A gap between successful runs | Only once the job runs again | Sentry being down entirely — the row is in Postgres |
| **(c)** Per-pass drain check | The job ran, checked in green, and did not repair what it found | Same run | — |

**(a)** wraps the real run in `Sentry.withMonitor(JOB_NAME, …, MONITOR_CONFIG)` — `checkinMargin: 10` on the `*/30` cadence surfaces a skipped run ~40 min after its slot; `maxRuntime: 25` sits under the cadence so a wedged run is flagged before the next fires, and above the arithmetic worst case (four statements × the image's 5-minute `DB_STATEMENT_TIMEOUT_MS`). The monitor is upserted from the config on the first check-in, so there is nothing to click in the Sentry UI. `withMonitor` re-throws, so the existing `captureError` path is untouched. `CRON_SCHEDULE` is pinned against `package.json`'s `cron-schedule` by a unit test — tsup doesn't bundle `package.json` into `dist/`, so the value has to be duplicated in code and can drift from what the deploy installs.

**(b)** writes the heartbeat via the fleet's `updateLastRun`, inside the monitored callback so the check-in reports `ok` only when repair *and* heartbeat committed. The next run reads it back to log the elapsed gap and warn past `LINKAGE_RESOLVE_MAX_GAP_HOURS`. The ticket's own note that "(b) alone is insufficient because nothing would read it" is why the read exists at all; making the *next run* the consumer avoids adding a 47th standing cron. The read is **fail-open** — a malformed env var or an unreachable `cronjob_runs` logs, captures, and continues, because observability must not be able to stop the repair or skip the check-in (which would turn an observability fault into a phantom "cron is down" alert).

**(c)** is folded into the job, not a new job: a pass that saw candidates and wrote fewer rows than it saw. `library.legacy_release_id` carries a unique index, so a pass's `COUNT(*)` over the join counts exactly the distinct rows its UPDATE targets — `resolved < candidates` cannot happen benignly. (`resolved > candidates` can and is fine: `library-etl` shares the half-hourly slot, so a matching `library` row can land between the COUNT and the UPDATE.)

`--dry-run` deliberately emits none of the three.

## ⚠️ The heartbeat is not a delta bound

**Both SELECTs remain anti-joined on `album_id IS NULL` with no time bound, and must stay that way.** The `cronjob_runs` row this PR introduces is a **liveness heartbeat** and nothing may ever read it back into a repair predicate. A `last_run` filter would permanently strand every row whose `library` row landed during a window the job missed — the exact bug this job exists to prevent.

The distinction is stated in three places so it survives the next author: a comment on `resolveFlowsheetAlbumIds` (referenced from the rotation pass), the README's rewritten "No watermark" note, and a unit test that asserts neither statement contains `cronjob_runs`, `last_run`, `INTERVAL`, `NOW(`, or an `add_time`/`updated_at` lower bound. **Reviewers: please reject any future change that reads `cronjob_runs` into these SELECTs.**

## Deviation: no cohort-age detector, and why

The ticket describes (c) as "alert when a row has sat at `album_id IS NULL` with a matching `library` row for more than N hours". That shape is not implementable on the current schema, and the reason is worth recording:

**`library.add_date` and `library.last_modified` are legacy-sourced.** `jobs/library-etl/job.ts` writes them from tubafrenzy's `release_time_created` / `release_last_modified` (`LEGACY_SOURCED_LIBRARY_COLUMNS`), not `now()`. No column records when a `library` row became *visible to Backend*, so an age derived from them measures the librarian's filing date, not repair latency — a catalog gap-fill importing a 1998 release today would report a 27-year-old "candidate" on a perfectly healthy run. `rotation.add_date` is additionally a `date`, not a timestamp. Anchoring on the candidate row's own age has the mirror-image problem (it overstates whenever the `library` row landed second, which is the dominant case by construction).

Rather than add a column, (c) is implemented on the one anchor that is provably quiet on a healthy run — the drain invariant above. Adding a Backend-local `library.first_seen_at` would make the age detector viable later; it wasn't worth a migration here.

## Sentry Crons quota and cost

Every paid Sentry plan includes exactly **one** cron monitor; additional monitors are **PAYG-only at $0.78/monitor/month** ([docs.sentry.io/pricing](https://docs.sentry.io/pricing/)). WXYC is on Team ($29/mo, $100 PAYG cap, post-#1291), so **this monitor costs $0** if it's the org's first, and $0.78/mo otherwise — either way inside budget, no plan change, nothing to approve.

Check-ins bill under a **quota category separate from errors**, which is the main argument for (a) over a `captureMessage`-based signal: a missed-check-in alert survives an error-quota exhaustion of the kind that took every WXYC project dark on 2026-06-03 (#1291). That property does **not** extend to (b)'s and (c)'s warnings, which are ordinary events — their Sentry-independent value is that `cronjob_runs.last_run` is a durable, human-queryable row in Postgres.

**Before fanning this across the fleet, price it:** 46 crons would be ~$35/mo of PAYG against the $100 cap. The recipe in `docs/ops-cron-scheduling.md` says to monitor the crons whose silence is expensive rather than all of them.

## Thresholds and where they come from

| Knob | Value | Derivation |
| --- | --- | --- |
| `checkinMargin` | 10 min | Under the 30-min cadence; a skipped run alerts ~40 min after its slot = "one cadence plus margin", while absorbing a deploy that recreates the crontab entry a few minutes late |
| `maxRuntime` | 25 min | Under the cadence so a wedged run is flagged before the next fires; above the worst case of four statements × the image's 5-min `DB_STATEMENT_TIMEOUT_MS` |
| `LINKAGE_RESOLVE_MAX_GAP_HOURS` | 4 h | Eight consecutive missed runs at `*/30` — far past any deploy, reboot, or maintenance window. Deliberately much looser than (a)'s 40 min so the two don't double-report one blip; (b) is the backstop, not a second copy |

Derived from the cadence and the container's own statement timeout rather than from prod data — deriving them from prod would have meant materializing the prod DB password through an SSH tunnel, which isn't warranted for threshold-picking on signals this coarse.

## Fleet-general record (last acceptance criterion)

`docs/ops-cron-scheduling.md` gains a **"Cron liveness (BS#2064)"** section: the three-step recipe, the `CRON_SCHEDULE`-vs-`package.json` drift trap, the "a heartbeat nothing reads is not a signal" rule, the per-monitor cost, and the relationship to **#1201** (the CloudWatch-side version — a heartbeat metric from the cron *wrapper*). The two are complementary and nothing here forecloses #1201; the section repeats the org convention that a plain-form CloudWatch alarm needs a dimensionless companion series or it sits at "no data" forever (the wxyc-canary#13 post-mortem).

## Deploy / ops

- **No schedule or crontab change.** `deploy-base.yml` reads `job-type` from the job's `package.json`; this job has no `job-type` key, which defaults to `cron` → `TARGET_TYPE=job` → the crontab entry is registered and verified as before. `cron-schedule` is unchanged at `*/30 * * * *`. Verified by reading the workflow, not assumed.
- **Land-time ops step — verify `SENTRY_DSN` is in the EC2 `~/.env`.** The container runs with `--env-file .env`; without a DSN the SDK no-ops and signal (a) silently does nothing (as does the pre-existing `captureError`). `docs/env-vars.md:96` still carries an open TODO about provisioning `SENTRY_DSN` for a sibling cron, so this is worth confirming rather than assuming. (b) works regardless — part of why it's here.
- Still DB-only, no `@wxyc/lml-client`, still exempt from the LML cron-spacing policy. The slot table is unchanged.
- `LINKAGE_RESOLVE_MAX_GAP_HOURS` is optional; the in-code default applies when it's absent from `.env`.

## Shared test util

`tests/utils/render-sql.ts` (the BS#2051 canonical renderer that replaced 40+ hand-rolled copies) threw on one chunk shape: an interpolated mock schema table, which this job's SQL uses on every statement and its two existing consumers don't. Rather than make this suite the 41st private renderer, `renderValue` learns that shape — narrowly, via a key-equals-value test that matches how `tests/mocks/database.mock.ts` models tables, so a genuine bound object still throws — with coverage added to `tests/unit/utils/render-sql.test.ts`.

## Checks

`format:check`, `typecheck` (plus `tsc --noEmit` on the job package, which the root script doesn't cover), `lint` (0 errors), `build`, `test:unit` (**421 suites / 6,858 tests green**), and the CI guard scripts: `lint:migrations`, `check-legacy-entry-id-writes`, `check-lml-caller-classification`, `check-bulk-update-analyze --strict`, `check:auth-tables-doc`. No migration in this PR.

## Note for the reviewer

A sibling PR for #2065 (the `shows.end_time` Phase 3 gap) may also touch `docs/ops-cron-scheduling.md` or `CLAUDE.md`. Both edits here are additive and confined to a new section / one table row, so the rebase should be trivial.


jakebromberg (commented):
## Review — approve with nits

Reviewed twice, independently. The mechanism is sound: both SELECTs are genuinely unbounded, the error path is preserved, the dry run is inert, a zero-candidate run is silent on all three signals, the deploy path is unchanged, and the Sentry cost figures check out. Nothing below blocks merge. The first two are worth fixing because they will mislead whoever responds to the first alert; the third is worth fixing because it guards the ticket's hardest constraint.

---

### 1. The drain check's "cannot happen benignly" is false

`jobs/legacy-linkage-resolve/job.ts` — `hasUnresolvedResidue`, and the same claim in `README.md` and the PR body.

The unique index on `library.legacy_release_id` makes the COUNT and the UPDATE agree **only under a single snapshot**. They are two separate `db.execute` calls with no wrapping transaction, so anything that removes a row from the cohort between them yields `resolved < candidates`. The docblock already accepts the symmetric race in the other direction and then denies its mirror image.

Most likely path first:

- **An operator run.** `jobs/broken-fk-recovery/job.ts` issues the byte-identical flowsheet UPDATE. It is `job-type: one-shot` and absent from the crontab, so it never races on a schedule — but it is exactly what someone runs by hand when investigating unlinked rows, and landing it between this job's COUNT and UPDATE produces `candidates = N, resolved = 0`: a warning asserting the job "did not repair what it found," on a run where everything worked. (Cron self-overlap is *not* a path — the deploy installs `docker rm -f <target>-cron` ahead of `docker run`, so a successor kills its predecessor.)
- **A DJ deletes a flowsheet entry**, or an MD edits one and picks an album (`updateEntry` writes `album_id`), or an MD deletes the `library` row a candidate joins to.
- **If uniqueness is ever relaxed** — `broken-fk-recovery` carries a "collision" classification explicitly against that possibility — the COUNT counts join pairs while the UPDATE counts target rows, and the warning fires on every run, permanently, describing a real condition that is not "the job stopped repairing."

**Suggested fix:** measure the invariant after the fact instead of across the race. Re-run the same COUNT after the UPDATE and warn on a non-zero *residual*. A row deleted, edited, or repaired by someone else drops out of the residual too, so every benign path goes quiet while a genuine "the UPDATE isn't draining what the COUNT sees" still fires. Failing that, soften the comment to "expected to be zero but race-prone" and require the condition on two consecutive runs before capturing.

While here: `broken-fk-recovery`'s docblock says it "Mirrors the resolver in jobs/flowsheet-etl/job.ts," which Phase 3 unscheduled. Worth repointing at this job now that it is the canonical one.

---

### 2. Nothing bounds the repeat rate of the new warnings

`jobs/legacy-linkage-resolve/job.ts`, `logger.ts` (`captureWarning`)

Per run the ceiling is three warning events (two drain, one gap) plus one exception from the fail-open gap catch — 144/day at `*/30`. Fingerprinting bounds *issue* creation to three stable issues; it does not bound *events*, and events are what the quota meters. That is roughly 4.3K/month, about 9% of the Team plan's 50K budget, on an org whose #1291 postmortem records going dark org-wide at 5K/month. Any of the persistent triggers in finding 1 turns this on and leaves it on — and because the job keeps working, nothing else prompts an investigation.

For calibration, the sibling `legacy-mirror-reconcile` has produced 13 and 11 events over 19 days at a daily cadence. This job runs 48× more often.

**Suggested fix:** gate `captureWarning` on a change of state — capture when the condition becomes newly true, or at most once per N runs — and keep the unconditional `log('warn', …)` for the per-run record.

---

### 3. The no-time-bound guard is a partial denylist, and no runtime signal backstops it

`tests/unit/jobs/legacy-linkage-resolve/job.test.ts`

The assertion is a regex denylist (`cronjob_runs|last_run|INTERVAL|NOW\s*\(|add_time\s*>|updated_at\s*>`). It would not catch `AND l.last_modified > $1`, `AND f.add_date >= $1`, `CURRENT_TIMESTAMP`, `age(...)`, or a subquery against a different watermark table — all plausible spellings of the same mistake by exactly the author the guard exists for, the one "optimizing the scan."

The sharper problem is that none of the three new runtime signals would notice either. A time bound collapses `candidates` toward zero, and a zero-candidate run is *the healthy shape*: (a) stays green, (b) stays green, and (c) is silent by construction because it requires `candidates > 0`. So the regression class this PR most wants to prevent is caught only by a partial static check, and if it slips through it runs undetected indefinitely — silently stranding exactly the rows the job exists to rescue.

Concretely: someone adds `AND l.last_modified > (SELECT last_run FROM cronjob_runs …)` and the test catches it. Someone adds `AND l.last_modified > $1` with the bound computed in JS, and the rendered SQL carries a numeric literal, no listed token matches, the tests pass, and every row whose library row landed before the window is stranded permanently.

**Suggested fix:** flip to an allowlist — snapshot the rendered WHERE clause and assert equality, so *any* added conjunct fails and forces a deliberate update.

---

### 4. The `maxRuntime` derivation undercounts the statements

`jobs/legacy-linkage-resolve/job.ts`, `README.md`

"Four statements against a pool whose image sets `DB_STATEMENT_TIMEOUT_MS=300000`" — the monitored callback issues up to **seven**: two COUNTs, two UPDATEs, two conditional `ANALYZE`s, and the heartbeat upsert. The arithmetic worst case is 35 minutes, which is *above* the 25-minute ceiling, not below it.

25 is still the right number, for a better reason worth stating instead: the deploy's `docker rm -f <target>-cron` means the next slot SIGKILLs any run that overruns 30 minutes, so no run can legitimately exceed the cadence — and a killed run sends no terminal check-in and writes no heartbeat, which is exactly what (a) and (b) should report.

---

### 5. Corrections to the fleet-general recipe

`docs/ops-cron-scheduling.md`, plus the same figures in `README.md`

This section is going to be copied, so the numbers matter:

- **"~40 min after its slot" is the wrong reference point.** Sentry marks a check-in missed at expected time plus `checkinMargin` — about 10 minutes after the skipped slot, which is 40 minutes after the last *successful* run. Detection is 4× faster than documented; an operator reading this would wait 40 minutes past a slot before trusting the signal.
- **"46 crons would be ~$35/mo of PAYG."** There are 44 job directories, of which 22 resolve to a scheduled cron (the other 22 are `job-type: one-shot`). With one monitor included, fan-out is 21 × $0.78 ≈ **$16/mo**.
- **"`@sentry/node` ≥ 7, already a dependency of every job"** — 31 of the 44 declare it, not all.
- `timezone: 'Etc/UTC'` reads as boilerplate. It is a no-op for an interval expression like `*/30`, but load-bearing for the wall-clock crons this recipe is aimed at (`0 9 * * *`, `07 15 * * *`). Worth saying so, since a copier of this recipe needs to confirm the host's TZ rather than inherit the line.

---

### 6. "(b) survives Sentry being down" is an overclaim, and the reason generalizes

`jobs/legacy-linkage-resolve/README.md`

The `cronjob_runs` row survives Sentry being down; the *signal* does not. The gap's only automatic consumer is `captureWarning`, a Sentry event, and the accompanying `log()` line goes to a container the next slot destroys.

That second half is worth stating explicitly because it is easy to assume otherwise, and it was checked on the host: the crontab installs `docker run` with no `--log-driver awslogs` and no redirect, the docker logging driver is `json-file`, and while the CloudWatch agent *is* running, its config has no `file_path` entries and no container collection — it is the metrics agent only. So container stdout is readable via `docker logs` until the next run's `docker rm -f`, and nowhere else. With Sentry down, (b) is forensic-only: valuable, but it needs a human to run the query.

---

### Verified

- **Both SELECTs stay unbounded, and the heartbeat never enters a predicate.** The cohort is `legacy_*_id IS NOT NULL AND album_id IS NULL` plus the join, nothing else; `getLastRunTimestamp`'s value flows only into a gap computation, a log line, and a warning.
- **`withMonitor` re-throws and the existing `captureError` path is untouched** — confirmed in the SDK source, which finishes an `error` check-in and rethrows. A failure produces both an error check-in and a captured exception, which is intended and not double-billed, since check-ins meter separately.
- **A dry run emits none of the three**, and **a healthy zero-candidate run is silent on all three**.
- **`job-type` resolves to `cron`**, so the crontab entry registers as before and there is no land-time schedule step. The code ↔ `package.json` ↔ crontab loop is closed and pinned by the new test.
- **The Sentry Crons cost figures are right**: one monitor included per paid plan, $0.78/monitor/month PAYG thereafter, and the org currently has zero monitors — so this genuinely is the first, at $0. Check-ins do ride a separate quota bucket from errors.
- **`SENTRY_DSN` is present and non-empty in the host `.env`, and the crontab entry passes `--env-file .env`** — so signal (a) will check in on the first post-deploy run. The open TODO in `docs/env-vars.md` about provisioning it is stale with respect to the host.
- **The "sole repairer" premise is accurate for the scheduled path.** The only other code running this join is the one-shot `broken-fk-recovery` (flowsheet only, never in the crontab) and the two unscheduled ETLs. `setFlowsheetLinkage` writes `flowsheet.album_id` but has no callers. Live `album_id` writers are the two webhooks and `updateEntry`, none of which re-run the legacy join.
- **No JS `Date` reaches a raw `sql` template** — the heartbeat goes through the insert builder, so the #1917/#1918 serializer trap does not apply. No array binding, no CloudWatch metrics.
- **The `NODE_ENV !== 'test'` auto-invoke guard is the established repo pattern**, byte-identical in `legacy-mirror-reconcile` and `catalog-popularity-freetext-resolve`. The one new silent-no-run path it creates is itself caught by (a) as a missed check-in.

### Nits

- `isMockTableShape` returns true vacuously for `{}`, so every empty plain object now renders `''` instead of throwing, as does any plain object whose values all equal their keys. It also sits *after* the `.value` dispatch, so a future mock table with a `value` column would render the literal text `value` into the SQL. No current table in the mocks has a `value`/`raw`/`sql`/`join`/`queryChunks` column, so there is no failing case today. A nominal marker on the mock tables would beat a structural heuristic if the guard is wanted at full strength.
- The heartbeat write is not fail-open, unlike the gap read fifteen lines above. The comment argues that observability must not stop the repair, then places `updateLastRun` — pure telemetry — inside the monitored callback un-caught, where a failure exits 1 and sends an `error` check-in even though the repair already autocommitted. The stated rationale is defensible; the asymmetry just is not acknowledged. No realistic isolated failure could be constructed, so this is a preference.

### Worth confirming after the first check-in lands

Whether a missed-check-in issue actually routes to a human. The project has four enabled issue-alert rules, but a missed check-in creates a cron-category issue that may not match a rule scoped to error-level events. A monitor whose failures notify nobody satisfies the letter of "produces an alert within one cadence plus margin" and not its intent.

PR #2067, opened 2026-08-09
Detect shows left open past a plausible duration after a dropped tubafrenzy show_end

jakebromberg:
Closes #2065.

Since [Phase 3](https://github.com/WXYC/wiki/issues/88) of the tubafrenzy decommission (#1858, landed 2026-08-08) unscheduled `flowsheet-etl`, the `show_end` fast-path on `/internal/flowsheet-webhook` is the only writer that ever closes a webhook-originated show. A delivery lost to a tubafrenzy restart, a 500, or a network blip leaves `shows.end_time` NULL **permanently** instead of for at most one ETL cycle — and `addEntry` (`flowsheet.controller.ts` ~L411) and `leaveShow` (~L695) gate on that column, so the departed DJ's show reads as live until the next show starts.

## Where the detector lives, and why

It rides the existing daily `jobs/legacy-mirror-reconcile` cron (`0 8 * * *` UTC) as a third, read-only arm — no new cron in a `jobs/` directory that already has 46 entries.

Of the recurring jobs, that one is the right host on three counts:

- **Same lifetime as the bug class.** This whole class dies at Phase 6a, when the webhook is retired. `legacy-mirror-reconcile` is retired then too — it exists only to mirror BS → tubafrenzy. Neither leaves standing machinery behind. `metadata-no-match-digest` (the other daily candidate, with SES plumbing already pointed at jake@wxyc.org) *outlives* turndown, so a tubafrenzy-scoped section in it becomes dead weight someone has to remember to remove — and its watermark/"since last run" shape fits an event stream, not a standing-state query.
- **Same shape as an arm it already has.** `runPartialReport` is literally this: read-only detection of legacy `shows` rows in a bad state → structured log + Sentry warning → never auto-fix, route to manual remediation. And it already has the `ReconcilePorts` DI seam, so the new arm is unit-testable with fakes exactly like its siblings.
- **Same window doctrine.** Its two sweeps already scope to `RECONCILE_WINDOW_HOURS` on the stated grounds that "older orphans are the historical-remediation class, deliberately out of scope for this recurring sweep." The detector reuses that knob verbatim for the same reason.

`docs/ops-cron-scheduling.md` needs no edit — `legacy-mirror-reconcile` is already listed under "Excluded / DB-only", and a read-only arm doesn't change its classification. (Noting deliberately: a sibling PR for #2064 may touch that file and the root `CLAUDE.md`; this PR touches neither, so there's nothing to conflict.)

The arm runs **before** the sweeps take their cooperative pause. It writes nothing, so deferring it behind a live DJ would only delay the signal.

## Threshold: 12 hours, derived from prod

Measured read-only against prod `wxyc_schema.shows` on 2026-08-09, over the 3,477 shows completed in the trailing 365 days (`end_time IS NOT NULL AND end_time > start_time`):

| p50 | p95 | p99 | p99.9 | max |
|---|---|---|---|---|
| 2.02h | 3.10h | 5.93h | 14.12h | 19.03h |

`>6h: 33 (0.95%) · >8h: 19 · >10h: 10 · >12h: 5 (0.14%) · >24h: 0`

12h is ~2x p99 and above 99.86% of real completed shows. The full derivation is recorded in the code at `STALE_OPEN_SHOW_HOURS_DEFAULT` (`jobs/legacy-mirror-reconcile/job.ts`), with the env override `STALE_OPEN_SHOW_HOURS`.

Same query established the current open-show population: **2,814** rows with `end_time IS NULL` — exactly 1 under 6h old (the live show) and **2,813** over 30 days old, with nothing in between. So the detector starts from a clean in-window baseline, and the 2,813 are #1543's cohort.

### The genuinely-active show can never trip it

The threshold is not the only guard — three independent bounds, and the acceptance criterion rests on the third, not on the threshold being tuned right:

1. `start_time < now() - 12h` — the plausible-duration threshold.
2. `NOT EXISTS` a flowsheet row newer than the same cutoff — a genuinely-live marathon show is still logging tracks.
3. `id IS DISTINCT FROM (SELECT max(id) FROM shows)` — the row `flowsheet_service.getLatestShow()` returns (`ORDER BY id DESC LIMIT 1`), i.e. exactly what `addEntry` / `leaveShow` / `joinShow` treat as current.

An overnight or special-programming block past 12h is excluded twice over. The integration spec's `current` fixture is deliberately given a stale-looking shape (20h old, no recent entries) so that **only** bound 3 can keep it out — and asserts it does.

Plus an outer bound `start_time > now() - RECONCILE_WINDOW_HOURS`. The in-window band (36h at defaults) is wider than the job's 24h period, so a show that goes stale is always seen by at least one run before it ages out.

## Opportunistic backfill: included

`closeShowFromTerminalShowEndMarker` fires under the BS#1861 option (b) guard in `joinShow`, when that guard has just proven the show is closed while `end_time` still reads NULL. It stayed clean, so it's in:

- Writes the **terminal marker row's own `add_time`**, re-read inside the statement — the same value the fast-path derives from its single clock reading per delivery, so the column stays consistent with the marker.
- Re-checks the marker type **inside** the `UPDATE`, so there's no TOCTOU against the caller's separate `isLatestEntryShowEnd` read.
- Carries the same `WHERE end_time IS NULL` guard as the fast-path, so a later delivery of the same `show_end` — or #1543's authoritative dump pass — still wins.
- **Best-effort by construction.** A DB error is captured to Sentry and swallowed, returning 0. It sits between the (b) guard's read and the start-vs-join decision that guard drives; letting a cosmetic backfill throw would 500 a go-live that (b) was about to route correctly, turning the acute hazard (b) exists to prevent back on. Pinned by a test.

It only fires when someone next goes live, so it is a complement to the detector, never a substitute.

## Repair pass: out of scope

The accumulated residue is repaired from the final tubafrenzy `mysqldump` under **#1543 item 3**, which already does exactly this for `start_time` on the same rows. That item's scope was extended to cover `end_time` alongside `start_time` in [an existing comment on #1543](https://github.com/WXYC/Backend-Service/issues/1543#issuecomment-5232239200) — no sweep is duplicated here. The detector counts the out-of-window cohort (`historical_open_shows`) but never lists it, so #1543's pass has a verification signal and the nightly report isn't buried by rows it will not act on.

## BS#1861 option (b) is intact

Verified before and after. `latestEntryIsShowEnd` and the `if (current_show?.end_time !== null || latestEntryIsShowEnd)` routing are byte-identical; the backfill is inserted between them and is purely additive. Existing (b)/(c) tests unchanged and green, plus three new controller tests pinning that the backfill fires only on the (b) path and never changes the routing.

## No "the ETL will fix it"

Every comment added here names what actually repairs each case (detector / opportunistic backfill / #1543 dump pass) and says explicitly that no scheduled job re-derives the value. The `shows` docblock in `schema.ts` gains a bullet on the post-Phase-3 meaning of a NULL `end_time` — the same stale assumption `58a9cd76` rewrote those two comment blocks for.

## Validation

- **Generated SQL rendered and run read-only against prod.** This caught a real bug pre-merge: drizzle renders `${flowsheet.show_id}` / `${shows.id}` fully-qualified inside `where(...)` but **bare** inside a `select({...})` projection, so the projection subqueries came out as `WHERE "show_id" = "id"` — a silent `flowsheet.show_id = flowsheet.id` self-correlation returning arbitrary rows. Fixed with an explicit `fe` alias and a table-qualified outer reference; the trap is documented at the call site. Post-fix, prod returns 0 rows in the 12h/48h band (matching the measured baseline), correct per-show `last_entry_type` values when the window is widened, and `EXPLAIN ANALYZE` at 7.6 ms / 1,771 buffer hits.
- **Unit**: 6,839 pass. New coverage for the detector arm (report shape, silence at zero, never-repairs, ordering vs. the cooperative pause, historical count-not-list, bounded Sentry sample), the `STALE_OPEN_SHOW_HOURS` option, and the backfill (guard preserved, marker timestamp not `now()`, no-op, failure swallowed).
- **Integration**: new `tests/integration/legacy-mirror-reconcile-stale-open-shows.spec.js`, a hand-written SQL twin in the same contract as the BS#1707 sibling spec. Run locally against a seeded Postgres — 7/7 pass, including the current-show exclusion.
- **Local CI gates green**: `format:check`, `typecheck`, `lint` (0 errors), `test:unit`, `lint:migrations`, `check:auth-tables-doc`, `check-legacy-entry-id-writes`, `check-lml-caller-classification`, `check-bulk-update-analyze --strict`, `build`.

### Sentry-quota note

The detector's per-show detail goes to the **structured log** (CloudWatch), which is the durable sink; Sentry gets one aggregate warning per run carrying the count plus a bounded 10-show sample. This deliberately diverges from `runPartialReport`'s per-show capture: a detector that fires on a backlog would burn error-quota proportional to the backlog, and the WXYC Sentry org has exhausted its org-wide quota before (#1291, 2026-06-03), taking every project's error ingest down for days. A detector must not be the thing that does that — and if Sentry is dark, the log line still lands.

## Related

- #1861 — added the fast-path this is the residue of; its option (b) guard is preserved
- #1543 — owns the final-dump repair half, `end_time` alongside `start_time`
- #2064 — the sibling Phase 3 "nothing repairs this anymore" gap, on `album_id` linkage
- [wiki#88](https://github.com/WXYC/wiki/issues/88) Phase 3, and #1858 which landed it


jakebromberg (commented):
## Review — two changes requested, then this is good

Reviewed twice, independently. The detector, the threshold derivation, and the backfill are carefully built, and the pre-merge SQL-rendering catch was a genuinely good one. Three things below; the first two are worth fixing before merge.

Both reviews independently arrived at finding 1, from different starting points, which is the main reason it is written up first.

---

### 1. The `max(id)` exclusion removes exactly the state the ticket describes as harmful

`jobs/legacy-mirror-reconcile/orchestrate.ts` — `latestShowId`, and the `${shows.id} IS DISTINCT FROM ${latestShowId}` bound.

The bound excludes `max(shows.id)` unconditionally, on the reasoning that "the row the API calls live is never reported." But `max(id)` **is** the state in which a dropped `show_end` does its damage: `addEntry`, `leaveShow`, and `getOnAirDJName` all gate on `getLatestShow()`, i.e. `max(id)`. So the detector can only report a stale show *after* the acute harm has ended — once a newer show row exists — and if no newer row appears while the show is still inside the reporting band, it is never reported at all.

Concrete scenario, quiet week:

1. Legacy DJ signs off on tubafrenzy Saturday 22:00 PT. The `show_end` delivery is lost, so `end_time` stays NULL and the show remains `max(id)`.
2. Sunday's run: age 3h — under threshold, correctly not reported.
3. Monday's run: age 27h — inside the 12–48h band, no recent entries, but **still `max(id)`** → vetoed by this bound.
4. Tuesday 10:00 PT a legacy DJ signs on via the webhook. That path does not call `joinShow`, so the opportunistic backfill never fires. The stale show stops being `max(id)`.
5. Wednesday's run: age 75h — outside `RECONCILE_WINDOW_HOURS` → counted in `historical_open_shows`, never listed.

A dropped delivery that held "who's on air" wrong for 60 hours is never surfaced, and silently joins the historical cohort. The PR body's "the in-window band is wider than the job's 24h period, so a show that goes stale is always seen by at least one run" is true of the window arithmetic but does not survive this bound vetoing the runs that would have seen it.

The acceptance criterion is "the genuinely-active show never trips the detector," and `max(id)` is a stronger exclusion than that requires. A show whose newest flowsheet entry is a `show_end` marker is provably not genuinely active — that is BS#1861 option (b)'s insight, which this PR already relies on in `joinShow`.

**Suggested fix.** Make the bound conditional rather than absolute. Bound 2 (`NOT EXISTS` a flowsheet row newer than the cutoff) already encodes "genuinely live," so `id <> max(id) OR <no activity since the cutoff>` preserves the acceptance criterion while letting the max-id case be reported. Alternatively gate on the already-computed `newestEntryType IS DISTINCT FROM 'show_end'`. At minimum, report the max-id case under a distinct step or counter so it is observable rather than invisible.

Worth a test for the converse of the existing fixture: latest show id, `show_end` as newest entry, `end_time` NULL, past threshold — expect it reported.

---

### 2. A read-only detector failure now aborts the two repair sweeps

`jobs/legacy-mirror-reconcile/orchestrate.ts` — `runStaleOpenShowReport` is the first `await` in `runReconcile` and is unguarded; `job.ts` wraps the whole run in one try/catch that sets a non-zero exit code.

The detector is explicitly "read-only, never repairs," but an exception from `selectStaleOpenShows` or `countHistoricalOpenShows` propagates out of `runReconcile` and prevents `runShowCreateSweep`, `runEntrySweep`, and `runPartialReport` from running at all.

This is the same asymmetry the PR reasons about correctly for the backfill — "letting a failed cosmetic backfill throw would 500 a go-live that (b) was about to route correctly" — and then does not apply here.

Concrete scenario: the `shows` scan plus `flowsheet` anti-join trips the job image's statement timeout during an RDS burst-credit dip, or hits a transient connection reset. The detector rejects, `runReconcile` rejects, the run exits 1. That night's BS#1707 mirror self-heal never runs, so shows that lost their `res.finish` mirror attempt stay orphaned an extra 24 hours — and the advisory lock is released, so nothing retries until the next day. It repeats for as long as the condition persists.

**Suggested fix.** Wrap the detector arm in its own try/catch that logs, captures, and continues — the isolation `closeShowFromTerminalShowEndMarker` already uses.

---

### 3. The "CloudWatch is the durable sink" premise is false — verified on the host

The `STALE_OPEN_SHOW_SENTRY_SAMPLE` docblock routes per-show detail to the structured log only, keeping Sentry to one aggregate plus a 10-row sample, and justifies it with "the structured log (CloudWatch), which is the durable sink."

Keeping Sentry to one event per run is the right call — the reasoning about #1291 is sound. The durability half does not hold:

- The installed crontab line is `docker rm -f <target>-cron ...; docker run --name <target>-cron --env-file .env <image>` — no `--log-driver awslogs`, no redirect to a log file. The host's docker logging driver is `json-file`.
- The CloudWatch agent **is** running on the host, but its config contains no `file_path` entries and no container log collection — it is the metrics agent only, consistent with the IAM doc. Nothing ships container stdout to CloudWatch Logs.
- Each run begins by `docker rm -f`-ing the previous container, which destroys the prior run's json-file logs.

So a run reporting 14 stale shows emits 14 log lines and one Sentry event carrying 10 of them. If nobody reads the log that day, the next run deletes it, and the 4 shows outside the sample are unrecoverable — they have also aged out of the window, so no later run re-lists them.

**Suggested fix.** Put the full list in the single Sentry event rather than a 10-row sample. The quota argument is about *event count*, not payload size: one event with 200 ids costs exactly what one event with 10 costs. That keeps the quota discipline the design wants and gives it the durability it assumed. Alternatively, correct the comment to say the sink is `docker logs` with one-run retention.

No CloudWatch metrics are emitted by this job, so the dimensionless-companion convention does not apply.

---

### On the historical cohort, and what it means for #1543

The PR reads the 2,813 out-of-window open shows as the #1543 repair cohort, and presents `historical_open_shows` as a counter whose shrink after that pass is verifiable. That inference is worth checking before it is relied on, because the code path that produced those rows argues the dump cannot repair them:

- `jobs/flowsheet-etl/job.ts` copied `end_time` straight from tubafrenzy, with the shared epoch helper returning `null` for `0`/`NULL`.
- Its UPSERT set `end_time: excluded.end_time` unconditionally on every tick — so any show whose MySQL sign-off was non-NULL received it, repeatedly, for as long as the ETL ran.

Therefore a BS row that still holds NULL `end_time` alongside a `legacy_show_id` most likely reflects a tubafrenzy row that itself carries no sign-off timestamp — a legacy DJ who never signed off in the classic UI. Twenty-odd years of that produces exactly this shape: a large, uniformly old population, cleanly separated from live traffic. The bimodality is evidence for that reading, not for a dropped-delivery leak, since a leak would smear across the duration range.

If that holds, `historical_open_shows` will not shrink after #1543's pass, because the dump carries the same NULL — and #1543's `end_time` arm may have nothing to write for most of the cohort. Two cheap read-only queries would settle it: the origin split of the open shows (`primary_dj_id IS NULL AND legacy_show_id IS NOT NULL`), and the count of matching tubafrenzy rows whose end is NULL or zero. Worth recording the answer in the `historical_open_shows` docblock either way, so nobody later reads a flat counter as a failed #1543 pass.

This does not block the PR.

---

### Verified

- **BS#1861 option (b) is intact.** `current_show` is read once and never re-read; `latestEntryIsShowEnd` is computed before the insertion; the routing predicate reads only those two, and the diff shows no removed lines in that region. The backfill's return value is discarded and cannot alter the branch taken.
- **The rendering fix is correct, and the trap does not recur.** Rendered both shapes: the pre-fix projection form produces `WHERE "show_id" = "id"`, with both names resolving inside the subquery's own scope — a real self-correlation. The post-fix form is properly aliased and outer-qualified. The `NOT EXISTS` and `IS DISTINCT FROM` bounds sit in `where(...)`, where bare interpolations render fully qualified, so they are correct as written. One forward-looking note: the safe and bare forms now sit a few lines apart, and moving that `NOT EXISTS` into a projection later would silently reintroduce the bug.
- **The backfill's two subqueries cannot disagree** — same predicate, same ordering on a unique PK, one statement, one snapshot, and both correlate to a bound parameter rather than an outer column.
- **The written `end_time` matches what the fast-path would have written** — `markerTimestamp` is computed once and used for both the marker's `add_time` and the column; the backfill re-reads that exact value.
- **No throw escapes the backfill's catch**, and the swallow is the right call given where it sits in `joinShow`.
- **Latency is not added to every go-live** — the backfill fires only when `end_time IS NULL` *and* the newest entry is a `show_end`, i.e. the already-broken case.
- **No JS `Date` reaches a raw `sql` template** — `staleCeiling`/`windowFloor` bind a number into `now() - (interval '1 hour' * $n)`, so the #1917/#1918 serializer trap does not apply.
- **`schema.ts` and `internal.route.ts` changes are comment-only** — no migration implied.
- **Sentry cannot fan out** — one fingerprinted warning per run, fired only when the in-window set is non-empty, so a nightly `historical_open_shows: 2813` emits nothing.
- **Window scoping keeps the historical cohort out of the listing**, so the first run does not emit a wall of findings.
- **Scope holds** — both new queries are selects, the detector calls only log and capture, and the orphan alert total deliberately excludes the stale counts.
- **The threshold is derived, recorded, env-overridable, and validated** against zero and negatives.

### Nits

- `STALE_OPEN_SHOW_HOURS` is missing from `docs/env-vars.md`, where the job's other knobs are documented. The env lint only checks trailing whitespace, so CI will not catch it.
- The per-show log message asserts `likely a dropped tubafrenzy show_end delivery` for every row, including a dj-site show whose DJ never left (`last_entry_type: 'track'`) or a show row with no entries at all. `last_entry_type` is in the payload, so a reader can tell — but consider conditioning the wording on it rather than pre-committing to a cause.
- `countHistoricalOpenShows` has no current-show exclusion, so a genuinely current show older than the window would be counted in the repair cohort. Cosmetic, it is a count.
- The comment "so a later delivery of the same `show_end` still wins" is inverted for the redelivery case — the fast-path is itself `WHERE end_time IS NULL`, so a redelivery matches zero rows once the backfill has written. Moot in practice, since a redelivery carries the same timestamp, and the sentence *is* correct with respect to #1543's `IS DISTINCT FROM` pass. Just worth tightening.
- No test covers a *successful* backfill (return 1) leaving routing unchanged — all three new controller tests run against the default zero-row mock. Low risk, since the return value is discarded, but it is the one assertion that would pin the invariant the PR leads with.
- Out of scope, worth a follow-up ticket: the marker insert and the `end_time` stamp in `internal.route.ts` are two statements in one handler, not a transaction. That gap is the only failure mode the backfill can address at all — if the whole delivery is lost there is no marker, so the (b) guard never fires. Wrapping both writes in a transaction would eliminate the partial-write case at source.

Issue #1011, opened 2026-05-23
Complete historical drain at last_id=14603, then retire the daily flowsheet-metadata-backfill cron


jakebromberg:
## Problem

The `flowsheet-metadata-backfill-cron` is the historical drain introduced in PR #673. It is currently stopped (`Exited (1)` since 2026-05-22 ~16:35 UTC), paused at `last_id=14603` of run_id `aece22ed-b830-4312-bf17-6dcbc4e34a9b`. PR #1001 shipped the static rate-limit + concurrency gate that makes it safe to run under load (token bucket at `BACKFILL_LML_RATE_PER_MIN=20`, semaphore at `BACKFILL_LML_MAX_CONCURRENT=1`); the parent epic, Epic C "event-driven enrichment" (#877), reframes the cron entirely — child #895 will rewrite its WHERE clause from `metadata_attempt_at IS NULL` to `metadata_status = 'pending' AND inserted_at < now() - interval '15 minutes'`, making the cron a recovery sweep for what the CDC consumer (#892) missed, not a routine drain.

The current daily cron sits awkwardly across that transition. It still has historical work to finish (the rows past `last_id=14603` were never enriched), but its daily cadence and its `metadata_attempt_at IS NULL` semantics become obsolete the moment #891 lands.

## Desired end state

1. The historical backlog from `last_id=14603` to end-of-data is fully drained — the one-time job PR #673 was originally built to do.
2. After that drain completes once, the cron's daily cadence is retired (container removed or `restart: no` + not started). Its successor is whatever shape #895 lands: a recovery sweep that runs against `metadata_status = 'pending'` after Epic C ships, not before.

## Approach

1. **Resume the cron once, gated by the static limiter.** Re-deploy backend (so PR #1001's image is on EC2), `docker start flowsheet-metadata-backfill-cron`. Verify real-time LML p95 stays within the +20% envelope BS#995 specified (~9 s → ≤11 s) over the first hour via Sentry trace explorer.
2. **Let the drain complete.** With the static gate, the cron's pace is bounded at 20 LML calls/min; the remaining backlog (~10k–15k rows depending on what `last_id` reaches at restart) drains in ~10–15 h of cumulative runtime. Existing run-id persistence + idempotent claim already survive restarts.
3. **Stop the cron permanently once the drain reports `done`.** `docker stop flowsheet-metadata-backfill-cron`; remove from compose; document the retirement in the EC2 runbook.
[...]

PR #2179, opened 2026-08-17
feat(jobs): recurring cause-agnostic re-enrichment sweep for terminal enriched_no_match rows

### LINKED ISSUE #2176: enriched_no_match is terminal but the condition is not: rows never re-enrich after the upstream resolves
## Problem

`flowsheet.metadata_status = 'enriched_no_match'` is **terminal**, but the condition it records is **not permanent**. A playcut can be a correct no-match at write time and become fully resolvable days later — when discogs-etl's next rebuild caches the release, when an LML matcher fix ships, or when the librarian files the album. Nothing revisits the row, so it stays wrong forever.

Demonstrated on the 2026-08-16 `metadata-no-match-digest`:

- `flowsheet` **#5308981** — `Vladislav Delay — Kohde · Entain · Kepler`. Written `enriched_no_match` on 2026-08-16. Replayed against prod LML on 2026-08-17 (`/api/v1/lookup/bulk?allow_release_resolution_fallback=true`, the worker's exact config): **matches**, Discogs release `28327348`, artwork present. The row is still `enriched_no_match`.

The digest reader has no way to tell these apart from genuine permanent misses, so the calibration rule in the job's README ("new adds self-heal on the next rebuild") is misleading: the **cache** self-heals, the **row** does not.

## Why a recurring mechanism, not a fourth one-shot

This repo has now written **three** one-shot rescue drains for exactly this shape:

| Drain | Cohort | Cause |
|---|---|---|
| `jobs/flowsheet-reenrichment` (WXYC/Backend-Service#1433) | 11,965 unlinked rows | frozen before LML#583 |
| `jobs/flowsheet-linked-reenrichment` (WXYC/Backend-Service#1638) | 22,773 linked rows | frozen before `add_time < 2026-06-16` |
| WXYC/Backend-Service#1979 (open, `status:blocked`) | post-2026-06-16 cohort | class-5 4s budget cutoff |

WXYC/Backend-Service#1998 then found a fourth cohort (26,286 rows from the 2026-08-03/04 breaker incident) that "falls outside #1979's window and has no owner." Each drain is bounded by a frozen predicate tied to one historical cause, so every new cause mints a new orphan cohort and a new ticket. That is the pattern this ticket exists to stop.

**#1979 measured the recovery rate: replaying one day's 41 rotation-linked no-match playcuts recovered 24 (59%) as full Discogs matches.** A recurring sweep would have collected those without anyone noticing the cohort existed.

## Desired end state

A bounded, recurring re-enrichment pass that re-asks LML for `enriched_no_match` rows on a **cause-agnostic** schedule, so no future freeze cause needs its own one-shot drain or its own ticket.

## Where

- [`apps/enrichment-worker/enrich.ts#L642`](https://github.com/WXYC/Backend-Service/blob/ce3c74bc/apps/enrichment-worker/enrich.ts#L642) and [`#L655`](https://github.com/WXYC/Backend-Service/blob/ce3c74bc/apps/enrichment-worker/enrich.ts#L655) — the two terminal `enriched_no_match` write arms (linked / unlinked).
- [`jobs/flowsheet-metadata-backfill/`](https://github.com/WXYC/Backend-Service/blob/ce3c74bc/jobs/flowsheet-metadata-backfill) — **the seed to generalize.** Its epic-#1810 W4 pass already does exactly this, but only for rotation-linked rows whose `rotation.discogs_release_id` transitioned NULL→present, reusing `rotation.discogs_release_id_resolve_attempted_at` vs `flowsheet.metadata_attempt_at` as the state-change gate. The mechanism is right; the trigger is too narrow.
- `shared/database/src/schema.ts` — `flowsheet.metadata_status`, `flowsheet.metadata_attempt_at`. **Note the existing invariant:** the live CDC worker deliberately leaves `metadata_attempt_at` NULL on no-match rows, and WXYC/Backend-Service#1011 / WXYC/Backend-Service#895 depend on that. Both prior drains were careful never to write it. A new marker column is likely needed rather than reusing it.
- [`jobs/metadata-no-match-digest/README.md`](https://github.com/WXYC/Backend-Service/blob/ce3c74bc/jobs/metadata-no-match-digest/README.md) — the reader-calibration rule that needs correcting once this lands.

## Constraints

- **Do not reuse `flowsheet.metadata_attempt_at` as the retry marker.** It carries the WXYC/Backend-Service#1011 / #895 writer-discriminator meaning (worker leaves NULL, backfill stamps it); overloading it breaks the C6 gap-recovery sweep's candidate predicate.
- **LML budget is the binding constraint, not DB cost.** The no-match population is large (WXYC/Backend-Service#1998 alone is 26,286 rows). This must be a bounded drip with a retry TTL, not a full-cohort sweep — see `docs/ops-cron-scheduling.md` and the WXYC/library-metadata-lookup#995 pacing policy. Reuse the `rotation-release-id-backfill` shape: an `attempt_at` marker plus a no-match TTL so definitive misses back off while transient errors stay retryable.
- Cooperative live-DJ pause, per every other recurring job in `jobs/`.
- Must not overwrite a populated field with a null — reuse the fill-null COALESCE UPSERT + `updated_at < NOW()` race guard from `jobs/flowsheet-linked-reenrichment`.
- Never set `pending` (the WXYC/Backend-Service#1443 Option-2 rejection).
- Track-context trust gate (`isTrustedLmlTrackContextMatch`) applies to any new `extractArtwork` call site — see WXYC/Backend-Service#1959, which is already tracking three sibling un-gated paths.

## Acceptance criteria

- [ ] A recurring job (or an extension of `flowsheet-metadata-backfill`'s W4 pass) re-asks LML for `enriched_no_match` rows on a cause-agnostic trigger, bounded per run.
- [ ] A retry TTL prevents a permanently-unresolvable row from being re-asked indefinitely; a row that exhausts its retries is distinguishable from one never retried.
- [ ] Row #5308981 (`Vladislav Delay` / *Entain*) flips to `enriched_match` with Discogs release `28327348`.
- [ ] The new marker does not disturb `metadata_attempt_at`; add a test pinning the C6 sweep's candidate predicate against an exact-match allowlist (the `legacy-linkage-resolve` precedent).
- [ ] Dry-run mode reporting candidate counts and projected LML call volume before any write.
- [ ] `jobs/metadata-no-match-digest/README.md`'s calibration rule updated: distinguish cache self-heal from row self-heal.
- [ ] Decide and record whether this **supersedes** WXYC/Backend-Service#1979 (`status:blocked`) or whether that one-shot still runs first to clear the existing backlog faster than a bounded drip can.

## Notes for implementer

- Reproduce the #5308981 case: `POST /api/v1/lookup` to prod LML with `{"artist": "Vladislav Delay", "album": "Entain", "song": "Kohde"}` — returns 1 result with artwork today.
- The digest job itself is read-only and is **not** the right place for this; it reports, it should not remediate.
- Worth checking whether the trigger can be cheap rather than time-based: an `album_metadata` row gaining a `discogs_url`, or a `library` row appearing for a previously-unlinked `(artist, album)`, are both state changes that could gate a re-ask far more cheaply than re-asking on a timer. The W4 pass already proves the state-change-gated shape works.

## Related

- WXYC/Backend-Service#1979 — the third one-shot drain; this ticket generalizes it (decide supersede-vs-sequence)
- WXYC/Backend-Service#1433, WXYC/Backend-Service#1638 — the first two one-shot drains
- WXYC/Backend-Service#1998 — the orphan cohort that motivated "no owner" framing
- WXYC/Backend-Service#1977 — complementary: a shed/timeout wrongly written as terminal no-match. That ticket fixes *writing* a wrong terminal state; this one fixes *never revisiting* a correct one
- WXYC/Backend-Service#1810 — the epic whose W4 self-heal pass is the seed mechanism
- WXYC/Backend-Service#1959 — trust-gate coverage for sibling `extractArtwork` call sites
- WXYC/Backend-Service#1629 — related classification gap in `metadata_status`

- WXYC/library-metadata-lookup#1206 — a recall bug whose two victim rows (#5308909, #5308853) stay permanently unenriched precisely because of the gap this ticket describes

jakebromberg:
Closes #2176

## Reviewing this PR

`git diff --stat origin/main..HEAD` reads 31 files / 8729 insertions, but **6520 of those insertions are `shared/database/src/migrations/meta/0150_snapshot.json`** — Drizzle's mandatory cumulative schema snapshot. It is tool-generated, never hand-edited, and carries no review value. Excluding it the diff is 30 files / 2209 insertions / 16 deletions.

Where the real diff is:

| Area | Lines | Notes |
|---|---|---|
| `jobs/flowsheet-no-match-recheck/**` | ~1028 | The new job package (of which `README.md` is 140 and the doc comments are a large further share) |
| `tests/unit/jobs/flowsheet-no-match-recheck/**` + the two pinning tests | ~958 | 931 for the new job, 26 for the C6-predicate pin, 1 for the policy pin |
| `Dockerfile.flowsheet-no-match-recheck` | 48 | Copy of the sibling job Dockerfile |
| `shared/database/src/migrations/0150_*.sql` | 58 | Mostly the ops header — 2 lines of actual DDL |
| `shared/database/src/schema.ts` | 44 | New column + partial index declaration |
| Docs, cron table, policy, mock, journal | ~89 | `CLAUDE.md`, `docs/env-vars.md`, `docs/ops-cron-scheduling.md`, `jobs/metadata-no-match-digest/README.md`, `shared/lml-client/src/policy.ts`, `tests/mocks/database.mock.ts` |
| `shared/database/src/migrations/meta/0150_snapshot.json` | 6520 | **Tool-generated. Skip it.** |

This is still above the org's ~1000-line PR target even after excluding the snapshot. The migration and the job are genuinely one unit — the job's candidate query cannot run without the column and its partial index — so splitting them would produce a first PR that adds a column nothing reads and a second that is still ~2000 lines. The compressible part is the tests, and shrinking those is not an improvement.

## ⚠️ Ops prerequisite — read before merging

**The `CREATE INDEX` in migration 0150 must be run out-of-band as `CREATE INDEX CONCURRENTLY` on prod BEFORE the deploy that applies this migration.** Drizzle wraps each migration file in a transaction and `CREATE INDEX CONCURRENTLY` cannot run inside a transaction block (the same constraint that governs 0057, 0068, 0070, 0074, 0078, 0080, 0139, 0144, 0148), so the in-file form is the plain one. Applied as-is against prod it takes an `AccessExclusiveLock` on the live ~2.6M-row / ~1.7 GB `flowsheet` table for the whole index build, pausing every flowsheet write — a wedged flowsheet mid-show.

Run this on prod first:

```sql
CREATE INDEX CONCURRENTLY IF NOT EXISTS "flowsheet_no_match_recheck_idx"
  ON "wxyc_schema"."flowsheet" USING btree ("no_match_recheck_attempted_at")
  WHERE "wxyc_schema"."flowsheet"."metadata_status" = 'enriched_no_match'
    AND "wxyc_schema"."flowsheet"."entry_type" = 'track'
    AND "wxyc_schema"."flowsheet"."artist_name" IS NOT NULL;
```

Then deploy. The migration's `IF NOT EXISTS` makes it a no-op against prod where the index already exists, while fresh dev and CI databases pick it up on first migrate. The migration file carries this same instruction in its comment header; it is repeated here so whoever merges cannot miss it.

## Problem

`flowsheet.metadata_status = 'enriched_no_match'` is terminal, but the condition it records is not permanent. A playcut can be a correct no-match at write time and become fully resolvable days later — when discogs-etl's next rebuild caches the release, when an LML matcher fix ships, or when the librarian files the album. Nothing revisits the row, so it stays wrong forever. The demonstrated case from #2176 is flowsheet row #5308981 (`Vladislav Delay — Kohde` / *Entain*), written `enriched_no_match` on 2026-08-16 and matching cleanly against prod LML the next day (Discogs release `28327348`) while the row itself is still `enriched_no_match`.

## Why a recurring mechanism, not a fourth one-shot

This repo has already written three one-shot rescue drains for exactly this shape — `jobs/flowsheet-reenrichment` (#1433, 11,965 unlinked rows), `jobs/flowsheet-linked-reenrichment` (#1638, 22,773 linked rows), and #1979 (open, `status:blocked`) — and #1998 then found a fourth cohort of 26,286 rows from the 2026-08-03/04 breaker incident that "falls outside #1979's window and has no owner." Every one of those drains is bounded by a frozen predicate tied to one historical cause, so each new cause mints a new orphan cohort and a new ticket. A fifth one-shot would buy one cohort; a standing sweep retires the pattern. #1979's own measurement is the argument for the payoff: replaying one day's 41 rotation-linked no-match playcuts recovered 24 (59%) as full Discogs matches.

## Design decisions

**New sibling job, not an extension of `flowsheet-metadata-backfill`'s W4 pass.** W4's self-heal is state-change-gated on one observable transition (`rotation.discogs_release_id` going NULL→present) inside a job whose worklist, batching, and LML pacing are tuned for a different cohort. Generalizing it in place would mean widening that job's candidate predicate — the same predicate the C6 gap-recovery sweep reads — to cover a cohort an order of magnitude larger, on a different cadence, with a different marker column. A sibling job keeps the two blast radii separate, and `jobs/rotation-release-id-backfill` already supplies the exact structural template (attempt marker + no-match TTL + cooperative live-DJ pause + bounded batch). W4 is left untouched by this PR; it becomes a strict subset of this job's cohort and is expected to be low-value redundancy after a few TTL cycles, at which point removing it is a small follow-up PR.

**Timer, not a state-change trigger.** #2176 asked whether a cheaper state-change gate (an `album_metadata` row gaining a `discogs_url`, a `library` row appearing for a previously-unlinked pair) could replace the timer. It cannot cover the motivating case: #5308981 resolved because LML's own matcher/cache improved between the original write and the replay — no local `library` or `album_metadata` row changed at all, so no gate over Backend-Service's own tables can see it. Only re-asking LML can. Generalizing a state-change gate to catch every upstream cause would mean watching an open-ended set of tables and columns, which is the "one cause, one predicate, one ticket" pattern this job exists to stop. And the binding constraint is LML call budget, not trigger cheapness: a state-change gate mainly saves calls on rows that were never going to resolve anyway, whereas the TTL + batch ceiling bounds cost directly and independently of cohort size.

**A new marker column, not `flowsheet.metadata_attempt_at`.** That column's NULL-vs-stamped split is a load-bearing writer discriminator (#1011 / #895): the live CDC worker deliberately leaves it NULL on no-match rows and only `flowsheet-metadata-backfill` stamps it, and the C6 gap-recovery sweep's candidate predicate reads it. Overloading it here would silently break the discriminator for every no-match row this sweep touches. Migration 0150 adds `flowsheet.no_match_recheck_attempted_at` (nullable timestamptz) instead, plus a partial index scoped to the exact cohort the sweep scans. **Nothing in this diff reads or writes `metadata_attempt_at`**, and `tests/unit/jobs/flowsheet-metadata-backfill/worklist.test.ts` now pins the C6 predicate against an exact-match allowlist (the `legacy-linkage-resolve` precedent) so a future change cannot quietly pull the new column into it.

**Marker stamped only on a definitive response.** A fresh no-match or a trust-gate rejection stamps `no_match_recheck_attempted_at` so the row backs off behind the TTL. A transient LML failure — a throw, a `{timeout:true}` cascade-exhaustion body, or a breaker-open/shed response with no usable answer — leaves the row entirely untouched so it stays immediately retryable. This is the invariant that stops the job from manufacturing a new false no-match while fixing old ones; it is the same split `rotation-release-id-backfill` uses.

**Defaults: TTL 14 days, batch 200, schedule `47 */6 * * *` UTC.** Candidates are ordered `no_match_recheck_attempted_at ASC NULLS FIRST, id ASC`, so never-attempted rows lead and a cohort too large for one pass drains fairly oldest-first rather than re-visiting the same head every run. Worst case per run is 200 LML calls against the shared concurrency-1 / 20-per-minute limiter — under 10 minutes of LML time, predictable regardless of cohort size, and clear of every slot in `docs/ops-cron-scheduling.md` including `rotation-release-id-backfill`'s `:17`. Both values are env-tunable (`FLOWSHEET_NO_MATCH_RECHECK_TTL_DAYS`, `FLOWSHEET_NO_MATCH_RECHECK_BATCH_SIZE`) without a redeploy of anything else.

**Writes are fill-null and race-guarded.** Trusted matches only. Linked rows (`album_id` present) go through the fill-null COALESCE UPSERT into `album_metadata` lifted from `jobs/flowsheet-linked-reenrichment`, with the explicit `updated_at = NOW()` and the `updated_at < NOW()` `setWhere` race guard so a fresher worker or runtime enrichment is never clobbered; unlinked free-form rows get the same COALESCE shape applied to flowsheet's own inline metadata columns. Every flowsheet UPDATE additionally guards on `metadata_status = 'enriched_no_match'`, so a concurrent writer that already moved the row off that status makes the UPDATE affect zero rows, which the orchestrator counts as `raced` rather than treating as success. The 8 extended #1336 columns are deliberately omitted from both arms of the `album_metadata` UPSERT because this job never requests `extended: true` — writing them would clobber a worker-enriched row's extended fields with nulls. `pending` is never set (#1443 Option-2 rejection), and every match passes `isTrustedLmlTrackContextMatch` (#1359) so a same-artist substitution (`fallback` / `alternative` / `song_as_artist`) is never auto-persisted.

## Decision needed from a human: does this supersede #1979?

#2176's last acceptance criterion asks whether this job **supersedes** #1979 (currently `status:blocked`) or whether that one-shot still runs first. **This PR does not decide it, and does not touch #1979.**

The implementer's recommendation is **sequence, not supersede**: run #1979 once when it unblocks, to clear its specific already-measured cohort fast, and then let this job own all ongoing maintenance including #1979's residue — at which point #1979 can close. The reasoning is that this job is cause-agnostic and *would* reach that cohort on its own, but "eventually" is slower than a targeted one-shot with a proven payoff: 26k rows at 200 per 6 hours is roughly 3–4 weeks of steady-state operation, against #1979's measured 59% recovery on its sample. Whether three to four weeks of staleness is acceptable for that specific cohort is a product/urgency call, not a mechanism question. Recorded here and on #1979 for a decision; nothing in this branch depends on the answer.

## Verification status

Passing locally in the worktree:

```
npm run typecheck            -> exit 0
npm run lint                 -> exit 0 (886 pre-existing security/detect-* warnings, all unrelated to this branch)
npm run format:check         -> exit 0
npm run test:unit            -> 445 suites / 7515 tests passed
npm run lint:migrations      -> Migration journal validation passed (148 entries, 0 warnings)
npm run check:lml-callers    -> PASS
npm run check:auth-tables-doc -> PASS
```

**`npm run ci:testmock` was NOT run successfully — the Docker/integration tier is unverified locally.** It failed at `ci-db-init` with `password authentication failed for user "postgres"`, caused by a concurrent session on the same machine holding the shared `dev_env` Compose project namespace with a persisted Postgres volume whose baked password does not match this checkout's `.env`. Tearing that down would have destroyed the other session's environment, so it was left alone. **No reviewer should read the list above as evidence that the Docker CI tier passed** — GitHub Actions is the first place the integration tier actually runs for this branch.

Also unverified: #2176's literal acceptance criterion that **row #5308981 flips to `enriched_match` with release `28327348`** requires the prod DB and prod LML together. That row's exact data is used as a unit fixture, but the flip has not been replayed against production. It should be confirmed after the first prod run.

One flake observed and cleared: `internal.route.test.ts`'s djHandle-truncation case failed once in a full-suite run (401 vs 200) and passed both in isolation and on an immediate full rerun (445/445, 7515/7515). Cross-test-order flakiness in a file this branch does not touch.

## Acceptance criteria from #2176

- [x] Recurring, cause-agnostic, per-run-bounded re-ask of `enriched_no_match` rows
- [x] Retry TTL backs off definitive misses; a stamped row is distinguishable from a never-attempted one (`no_match_recheck_attempted_at` NULL vs set)
- [ ] Row #5308981 flips to `enriched_match` with release `28327348` — **needs prod DB + prod LML; confirm after first prod run**
- [x] Marker does not disturb `metadata_attempt_at`; C6 predicate pinned against an exact-match allowlist
- [x] Dry-run mode reporting candidate count and projected LML call volume before any write
- [x] `jobs/metadata-no-match-digest/README.md` calibration rule corrected to distinguish cache self-heal from row self-heal
- [ ] Supersede-vs-sequence decision on #1979 — **recommendation recorded above and on #1979; awaiting a human decision**


jakebromberg (commented):
## Review

Reviewed at `0371c044` against merge-base `ce3c74bc`. The mechanism is the right call and the structure is faithful to its stated donor — the findings below are about behaviour at the edges, not about the shape. Two of them are load-bearing enough that I would not deploy the job as-is without a decision on them, though neither is a reason to redo the design.

Findings are ordered by severity. The first four were verified against the sources they depend on rather than accepted from the code's own comments.

### HIGH 1 — the unlinked arm's fill-null COALESCE is a guaranteed no-op on all four streaming URLs, and the "matches the donor exactly" claim does not hold

`jobs/flowsheet-no-match-recheck/writer.ts`'s `writeUnlinkedMatch` fills each column with `COALESCE(${flowsheet.col}, <discogs value>)`. The header comment and the PR body both justify this as matching `jobs/flowsheet-linked-reenrichment`. Checked against the donor: **`flowsheet-linked-reenrichment` has no unlinked arm at all.** It is linked-only by construction (its frozen cohort predicate includes `album_id IS NOT NULL`), and its only COALESCE is the `album_metadata` UPSERT — which `upsertLinkedMatch` here does reproduce faithfully, column for column. So the linked arm's claim is accurate; the unlinked arm has no donor and the analogy is doing work it cannot support.

That matters because of what the original no-match write leaves behind. `apps/enrichment-worker/enrich.ts:648-661` — the unlinked terminal `enriched_no_match` arm — writes:

```ts
.set({
  spotify_url: searchUrls.spotify_url,
  youtube_music_url: searchUrls.youtube_music_url,
  bandcamp_url: searchUrls.bandcamp_url,
  soundcloud_url: searchUrls.soundcloud_url,
  metadata_status: 'enriched_no_match',
  ...
})
```

unconditionally, from `synthesizeSearchUrls` (`enrich.ts:477-480`, `https://open.spotify.com/search/…` and friends). Every unlinked candidate this job selects therefore arrives with those four columns **already non-null**, holding a synthesized search URL. `COALESCE(flowsheet.spotify_url, <real Discogs spotify_url>)` keeps the search URL and discards the real one, every time, with no exception. Four of the ten columns in that arm are dead code in production.

The remaining six (`artwork_url`, `discogs_url`, `release_year`, `apple_music_url`, `artist_bio`, `artist_wikipedia_url`) are not written by the no-match arm, so they do fill correctly and the job still delivers its headline value. This is a partial-upgrade bug, not a total one.

Is it right for this job? I don't think so, and the asymmetry with the linked path is the reason. A linked row's search URL is recoverable: `album_metadata` carries `spotify_status` / `bandcamp_status` and the BS#1915 hourly `streaming-reask.ts` sweep re-adjudicates it on a recurring basis. An unlinked row has no status column and no recurring sweep — the only thing that upgrades a search-shaped flowsheet URL is `jobs/streaming-url-upgrade` (BS#1672), and its `package.json` says `"job-type": "one-shot"`. So on the unlinked path a search URL that survives this job survives indefinitely. Given the job's whole premise is upgrading rows that were previously wrong, silently declining to upgrade the one class of value that is *known* to be a placeholder is the wrong default.

Suggested fix, and it should be narrow: on the unlinked arm only, prefer the incoming value over a stored value that is search-shaped. `@wxyc/lml-client` already exports the host predicates (`isSpotifyUrl`, `isAppleMusicUrl`) that `streaming-url-remediation` uses for exactly this arbitration, and `apps/enrichment-worker/precheck.ts:80` already encodes the `LIKE '%bandcamp.com/search%'` shape test. Keep plain fill-null for the other six columns. If instead the intent is to leave this to a future ticket, the code comment and the README should say that the four streaming columns are deliberately not upgraded on the unlinked path and why — right now both assert the opposite.

### HIGH 2 — class-5 registration re-asks LML under the same ~3.8 s budget that produced these false no-matches

`shared/lml-client/src/policy.ts:280` maps `'flowsheet-no-match-recheck'` to traffic class 5. The docstring on that class (lines 61-76) is unusually direct about what that means: the 4 s cutoff "is WRONG for this caller and is actively producing user-visible metadata gaps", LML returns "either an empty `timeout: true` body or `degraded: true, degraded_reason: 'deadline_exceeded'` with `artwork: null`", and the 2026-08-04 replay recovered 59% of the cohort **headerless**. The `budgetMs: null` suppression lever exists for precisely this caller shape and is not used here; flipping it for the worker is BS#1978.

The 35 s `FLOWSHEET_NO_MATCH_RECHECK_LML_PER_CALL_TIMEOUT_MS` does not help — it is a socket abort, not an LML search budget, and per the same docstring the header can only tighten LML's `min(header − 200, LML_SEARCH_BUDGET_MS)`, never extend it. So `lml-fetch.ts:56`'s comment ("so a cold, hard-to-resolve release still gets LML's full cascade on retry") and the identical claim in the README are not true as written.

The practical consequence is that the rows most likely to be false no-matches — cold, hard-to-resolve releases — are re-asked under the budget that made them no-matches, come back empty, and (see HIGH 3) get stamped for another 14 days. Worth resolving before first prod run, even if the resolution is just "accept it, revisit after BS#1978 ships" recorded explicitly.

### HIGH 3 — `deadline_exceeded` is classified as a definitive no-match and stamps the marker

This is the transient-vs-definitive split, and it has a gap. `lml-fetch.ts`'s `isUnansweredDegraded` matches only `degraded_reason === 'upstream_unavailable'` and `shedReasonOf(...)`. Per policy.ts:67, the budget cutoff can also return `degraded: true, degraded_reason: 'deadline_exceeded'` with `artwork: null`. That shape is not a timeout body, is not `upstream_unavailable`, and is not a shed — so it falls through to the `search_type` arms, returns `{ kind: 'no_match' }`, and the orchestrator stamps `no_match_recheck_attempted_at` and backs the row off for the TTL.

That is the job manufacturing a fresh false no-match from an unanswered call — the exact self-defeating outcome `lml-fetch.ts`'s own module docstring says it is avoiding, and the thing BS#1977 was opened to stop. It also compounds HIGH 2: under the class-5 budget this is not an edge case, it is a common response for the cohort.

The breaker/shed/throw arms themselves are correct — a throw, a `timeout: true` body, and a shed with no usable answer all leave the marker untouched, and the orchestrator's `lml_error` path is genuinely non-stamping. The fix is one more `degraded_reason` in `isUnansweredDegraded`, plus a unit case pinning it.

### MEDIUM 4 — the partial index serves the filter but cannot serve the ORDER BY

Checked the predicate/index agreement directly, and the good news first: migration 0150's index predicate is character-identical to the query's three static clauses (`metadata_status = 'enriched_no_match' AND entry_type = 'track' AND artist_name IS NOT NULL`), so **the index is usable and the job cannot silently seq-scan `flowsheet`.** That part of the design is sound and the claim in the migration header is right.

The ordering claim is not. `flowsheet_no_match_recheck_idx` is a plain `btree("no_match_recheck_attempted_at")`, which is ASC NULLS **LAST**. The query orders `ASC NULLS FIRST, f."id" ASC`. Postgres cannot satisfy a NULLS-FIRST ordering from a NULLS-LAST index in either scan direction, and `id` is not in the index, so the plan reads every matching row and sorts before applying `LIMIT 200` — not the no-sort path migration 0150 line 27 and `schema.ts` both assert. On a cohort in the hundreds of thousands that is a real per-run cost.

Fix is cheap but has an ops consequence: build the index as `(no_match_recheck_attempted_at NULLS FIRST, id)`. **If that change is made, the out-of-band `CREATE INDEX CONCURRENTLY` recipe in the migration header and in the PR body must change with it**, since it is meant to be run against prod before the deploy and would otherwise create the wrong index under the right name, leaving `IF NOT EXISTS` to silently accept it.

### MEDIUM 5 — never-stamped failure rows sit at the head of the queue forever

A row whose lookup throws (`lml_error`) or whose write throws (`db_error`) deliberately keeps `no_match_recheck_attempted_at = NULL`. That is the correct retryability guarantee, but combined with `ASC NULLS FIRST` it means such a row sorts ahead of every stamped row and ahead of every higher-id never-attempted row, so it is re-selected at the head of every subsequent run. A row that fails deterministically consumes one of the 200 slots permanently; enough of them and the drip stalls entirely. `rotation-release-id-backfill` has the same shape over a much smaller curated cohort, which is why it has never bitten. A failure counter, or stamping after N consecutive transient failures, would bound it.

### MEDIUM 6 — throughput may never reach the cohort, making the TTL vestigial

200 rows × 4 runs/day = 800/day. The README's "26k-row cohort clears in roughly 3-4 weeks" holds only if BS#1979's cohort were the entire candidate set; it isn't — those rows are interleaved by `id` among every other `enriched_no_match` row, and the org's enrichment no-match floor over a ~2.6M-row `flowsheet` puts the real cohort plausibly in the hundreds of thousands. At that size a single first pass takes one to two years and `FLOWSHEET_NO_MATCH_RECHECK_TTL_DAYS=14` never binds on anything. This does not break the job, but it does undercut the supersede-vs-sequence recommendation in the PR body: if the true cohort is that large, the case for running #1979 first is much stronger than "3-4 weeks vs. immediate". **Worth measuring the actual cohort count before the first prod run** and sizing batch/cadence against it.

### LOW 7 — flowsheet text is not HTML-entity-decoded before the LML hop

`jobs/rotation-release-id-backfill/lml-fetch.ts` — this job's own structural donor — carries `decodeHtmlEntities` because tubafrenzy paste data lands entity-encoded (`Rome&#769;o Poirier`) and LML's normalization never sees the combining mark, so the row stays a no-match. `flowsheet` receives the same tubafrenzy-sourced text, and this job passes `artist_name` / `album_title` / `track_title` through raw. Those rows will miss again and be stamped, for a cause the donor already solved one directory over.

### LOW 8 — `docs/migrations.md`'s attempt-at marker registry not updated

That section opens "Seven `timestamp with time zone` markers…" and enumerates seven bullets. This PR adds an eighth, and both migration 0150 and `schema.ts` point readers at that section as the authority for the marker's shape. Registry now under-reports.

### LOW 9 — `requireLmlConfigured` checks only `LIBRARY_METADATA_URL`

`LML_API_KEY` is documented as required in both the job docstring and the README env table but is not preflighted. With the URL set and the key missing, the run proceeds and burns ~10 minutes of rate-limited calls failing every row as `lml_error`, exiting 0 with `resolved: 0` — indistinguishable from a healthy quiet run.

### LOW 10 — no cron-liveness signal

`docs/ops-cron-scheduling.md`'s BS#2064 recipe targets exactly this shape: a recurring cron whose healthy idle run is a silent no-op, byte-identical to the crontab entry having been dropped by a deploy. Neither `Sentry.withMonitor` nor a `cronjob_runs` heartbeat is wired. The recipe is advisory, but this job's silence is the whole product.

### LOW 11 — cron margin below the documented recommendation

`47 */6 * * *` puts a heavy LML drain 30 minutes after `rotation-release-id-backfill`'s `:17`, on the same four hours. The hard invariant (distinct `HH:MM`) holds and the README's "clear of every slot" is true of that invariant, but `docs/ops-cron-scheduling.md`'s "Adding a new LML-hitting cron" step 2 asks for ≥60 min. Worth either moving it or noting the exception explicitly the way the 04:30/04:45 pair is.

### LOW 12 — marker-only UPDATEs will broadcast on the CDC path

`markRecheckAttempted` issues a real `UPDATE wxyc_schema.flowsheet`, which trips the `cdc_flowsheet` row trigger. `filterMetadataUpdate` accepts any flowsheet UPDATE whose `metadata_status` is in `TERMINAL_STATUSES` (which includes `enriched_no_match`) with no old-vs-new comparison, and the BS#2131 age guard is wired to `liveFs:insert` only. So each stamped row — whose client-visible state did not change — broadcasts to connected clients. Every precedent job that stamps a flowsheet marker is `job-type: one-shot`; this is the first recurring one, so the noise becomes permanent (~200/run × 4 runs/day). Filed low because it is a fan-out cost, not a correctness bug, and the cleaner fix probably belongs in `filterMetadataUpdate` rather than here.

## Verified as sound

These were checked specifically and hold up:

- **Nothing in the diff reads or writes `flowsheet.metadata_attempt_at`.** Confirmed against the full non-snapshot diff; every occurrence is prose, a doc comment, or the negative assertion in the new test. The `metadata-no-match-digest` invariant and the BS#1011/#895 writer discriminator are untouched.
- **The C6 predicate is genuinely unchanged and genuinely pinned.** The only edit to `jobs/flowsheet-metadata-backfill/worklist.ts` is `const pendingPredicate` → `export const pendingPredicate`; the SQL text is byte-identical. The new test asserts the whitespace-normalized predicate `toBe` a fixed string constant — a real exact-match allowlist, not a denylist regex — so any added conjunct fails regardless of spelling. This is the right shape and matches the `legacy-linkage-resolve` precedent it cites.
- **The race guards do prevent clobbering a concurrent writer.** Every flowsheet UPDATE (`markRecheckAttempted`, `flipLinkedStatus`, `writeUnlinkedMatch`) carries `eq(flowsheet.metadata_status, 'enriched_no_match')` in its WHERE, so a row another writer already moved off that status updates zero rows; the orchestrator reads that as `written: false` and counts `raced` rather than treating it as success. The `album_metadata` UPSERT's `setWhere: updated_at < NOW()` with an explicit non-COALESCE'd `updated_at = NOW()` reproduces the donor exactly and correctly prevents clobbering a fresher worker enrichment.
- **The 8 extended BS#1336 columns are correctly omitted** from both `values` and `set` in the linked UPSERT, since this job never requests `extended: true` — including them would null out a worker-enriched row's extended fields.
- **The trust gate is applied from day one** (`isTrustedLmlTrackContextMatch` before any resolve verdict), `pending` is never set, and the `no_match` / `trust_rejected` split is faithful to BS#1516.
- **The counter invariant in the README actually holds** in `runNoMatchRecheck` — every path through the loop increments exactly one bucket.
- **Migration wiring is correct**: journal idx 150, `applied-hashes.json` entry present, schema-qualified partial-index predicate matching the proven 0132/0148 form, `IF NOT EXISTS` plus the out-of-band `CONCURRENTLY` recipe consistent with precedent.

## Bottom line

HIGH 1 and HIGH 3 are the two I would want resolved before this runs against production, and both are small, local changes — a host-shape check on four columns in one arm, and one more `degraded_reason` string in one predicate. HIGH 2 is a genuine tension with the job's premise but is arguably out of scope pending BS#1978, as long as the two comments that currently claim the opposite get corrected. MEDIUM 4 should be fixed before the out-of-band index is created on prod, since fixing it afterwards means dropping and rebuilding a large index. Everything else can follow.

jakebromberg:
CI update, since the body flags the Docker/integration tier as unverified locally: it has now been exercised on GitHub Actions and is green. Run https://github.com/WXYC/Backend-Service/actions/runs/32054679620 at `0371c044` passed all six jobs — detect-changes, auth-tables-doc-drift, lint-and-typecheck, unit-tests, **Integration-Tests**, and **Migration Dry-Run (prod-shaped data)** — as did the companion Schema shape probe self-test. So the tier that could not run locally has run, including the migration against prod-shaped data.

The first run on `88a979c1` failed, and it was a real branch-caused failure rather than infra: `npm ci` aborted with `EUSAGE` / `Missing: @wxyc/flowsheet-no-match-recheck@1.0.0 from lock file`. The new job sits under the `jobs/*` workspace glob, so adding it makes it a workspace, and `package-lock.json` had no entry for it — every job failed at Install Dependencies before any check ran. Fixed in `0371c044` with a lockfile-only regeneration (`npm install --package-lock-only`), which added exactly the workspace entry and its link stub and churned nothing else. Worth noting for anyone adding a job here later: the local `npm run test:unit` / `typecheck` / `lint` set passes against an already-populated `node_modules` and cannot see this, so `npm ci --dry-run` is the cheap local check that would have caught it.

The prod acceptance criterion (row #5308981 flipping to `enriched_match` with release `28327348`) is still unverified and still needs a prod run to confirm.

PR #2189, opened 2026-08-17
fix(auth): take GET /auth/get-session off IP-keyed rate limiting

### LINKED ISSUE #2169: GET /auth/get-session 429s silently log DJs out: better-auth's limiter buckets every user by Cloudflare egress IP
## Problem

`GET /auth/get-session` is rejecting a large share of production traffic with 429, and the user-visible result is not a throttle — **it is a silent logout for DJs on dj-site**.

Verified from `/var/log/nginx/access.log` on the EC2 host, cross-checked line-for-line against Sentry's `http.server` spans (two independent instruments, matching hour by hour):

| Traffic day | 429s | Path |
|---|---:|---|
| Aug 8 | 1,058 | `/auth/get-session` |
| Aug 13 | **2,128** | `/auth/get-session` — 14.1% of that day's 15,126 requests |
| Aug 14 | 602 | `/auth/get-session` |
| Aug 6 | 590 | `/auth/token` — a single IP running `Go-http-client/1.1` in one hour. A bot, correctly throttled. Not this bug. |

Shape is **bursty, not steady**: median day ~70 429s (1.4%), six of 30 days had exactly zero, and Aug 13 alone is 32% of the month. No diurnal pattern (the four largest bursts start 08:00, 20:00, 02:00, 22:00 EDT) and no deploy correlation — Aug 13's burst ran straight through four releases. Latency does not degrade during bursts (200 p95 was 22 ms vs 60 ms baseline), so this is not a capacity problem; the limiter is rejecting cheaply.

These 429s are **invisible to every dashboard**. express-rate-limit responds without calling `next(err)`, so no Sentry issue and no error log is created — only the span carries the status.

### 1. better-auth's limiter is not a fixed window

`/auth/get-session` is **not** in `rateLimitedPaths` ([`apps/auth/app.ts#L345`](https://github.com/WXYC/Backend-Service/blob/716ae10b1440ffe06a524073e17d23b3b07b8e93/apps/auth/app.ts#L345)), so the express limiter never sees it. The only thing covering it is better-auth's internal limiter, running unconditionally on every request (`better-auth/dist/api/index.mjs:163-169`) at the **unmodified defaults** `window: 10, max: 100, storage: "memory"` (`create-context.mjs:169-175`) — `auth.definition.ts` has no `rateLimit` block at all.

`window: 10` does not mean "per 10 seconds". `rate-limiter/index.mjs:37` resets on time since the **last request**, and every *allowed* request bumps `lastRequest` (`:57`):

```js
if (now - data.lastRequest > windowInMs) return { count: 1, lastRequest: now, allowed: true }
```

The real rule is **100 requests per continuously-active stretch, clearable only by 10 full seconds of total silence on that key**. Any caller polling faster than once per 10 s accumulates monotonically to 100 and then 429s, no matter how slow it is.

Production data fits this and refutes the fixed-window reading: both Aug 16 bursts ran at ~4.5 req/s (≈45 per 10 s — a fixed window would never fire) and 429'd after ~100 cumulative requests. Recovery is bounded, because the denied branch returns `next: data` and `consume` only writes on `allowed`, so `lastRequest` freezes and the key clears ~10 s after the last allowed request. That yields a duty cycle: ~22 s of success, ~10 s of 429s, ≈31% rejection at that rate. Observed ≈27%; Sentry's worst hour was 57%.

### 2. The IP in the key has no discriminating power

better-auth keys on `` `${ip}|${path}` `` (`@better-auth/core/dist/utils/ip.mjs:226-228`) and reads only `x-real-ip`.

**100% of `/auth/get-session` traffic arrives from Cloudflare egress IPs** — verified across a full day (614/614) and across Aug 13's 2,128 429s, all in Cloudflare prefixes (104.22, 104.23, 172.64/68/69/70/71, 141.101, 162.158/159, 198.41, 108.162). dj-site is deployed via `@opennextjs/cloudflare`; its SSR arm (`session-cache.ts:35-49` → `serverAuthClient` → `api.wxyc.org`) is ~96% of the traffic and originates as an edge-side fetch carrying only a cookie.

nginx stamps `X-Real-IP $remote_addr` (`nginx -T` line 21, server-level; `location /auth/` at line 61 carries only `gzip off` + `proxy_pass` and inherits it), and there is **no `set_real_ip_from` or `real_ip_header` anywhere in the config** — verified.

Net: every dj-site user is bucketed by *which Cloudflare edge served them*. Roughly ten buckets of 100 for the entire DJ population.

Worth naming the irony: BS#774 moved bucketing from XFF to `x-real-ip` to stop spoofing, and in doing so discarded the only header that still carried a real client IP through a proxy.

## User-visible consequence

Mechanism VERIFIED end to end in dj-site; incidence NOT quantified.

1. `session-cache.ts:44-47` — the fetch `.catch` swallows *any* error and returns `{ data: null }`. A 429 becomes indistinguishable from "no session".
2. `server-utils.ts:12-20` — `getServerSession` sees `!session.data`, returns `null`.
3. `server-utils.ts:43-46` — `requireAuth` sees `!session` → `redirect("/login?bounced=no-session")`.

**A rate-limited DJ is silently signed out and bounced to the login page.** Not an error, not a retry.

Incidence could not be measured from this vantage point: the redirect is issued at Cloudflare so api.wxyc.org never sees it, and `bounced=no-session` appears in nginx only as a referer on later calls (5–7/day) — and `middleware.ts:11-13` emits the same URL for plain cookie-absence, so that counter conflates two causes. Aug 13 is the day it most plausibly bit real people: 429s ran 06:00–11:00 EDT, through the browser-traffic peak.

**Listeners are unaffected.** iOS calls `/auth/token`, not `/get-session`, from real device IPs, and coalesces refreshes through a process-wide actor at ~1 mint per 14 min. Audio and playlist are unauthenticated. Android never hits api.wxyc.org.

## Desired end state

A DJ's session-validity reads are budgeted per DJ, not per Cloudflare edge, and exhausting the budget cannot present as "signed out".

## Constraints

- `/auth/token` must stay IP-keyed. It mints JWTs, its abuse profile is different, and its only observed incident was a bot correctly throttled. **Do not extend this change to it.**
- Storage is a process-local in-memory `Map` (`rate-limiter/index.mjs:234-250`; `secondaryStorage` unset). Today that's one container running one `node dist/app.js` on a single host (`i-0685e373989cd5daa`), so the effective ceiling equals the configured one — but the limit would silently multiply if auth is ever scaled horizontally.
- better-auth's key generator is internal and not configurable, so a replacement has to sit in Express.

## Suggested approach

Take `/auth/get-session` off IP-keying entirely: disable better-auth's rule for that path via `customRules` (it accepts `false` — `rate-limiter/index.mjs:317`; the key is `/get-session`, since `normalizePathname` strips the `/auth` basePath), and mount an express-rate-limit instance in front keyed on the **session cookie**, falling back to IP when absent.

There is a precedent to copy in this repo: [`apps/backend/middleware/rate-limit-key.ts`](https://github.com/WXYC/Backend-Service/blob/716ae10b1440ffe06a524073e17d23b3b07b8e93/apps/backend/middleware/rate-limit-key.ts) namespaces `user:` vs `ip:` keys for exactly this reason (BS#1127 — "so one anonymous client's traffic can no longer exhaust the limit for every other unauthenticated caller"). This is the same defect one layer over.

**This is safer, not laxer:** `get-session` is a read, the cookie is the natural unit of the thing being read, each DJ gets their own budget, and a cookie-less attacker still falls back to IP bucketing.

### Alternatives considered and rejected

- **Recover the client IP via `CF-Connecting-IP` + `set_real_ip_from`.** Fixes only the ~3% proxied-browser arm. The SSR arm is ~96% and originates as a fresh edge-side fetch carrying only a cookie — there is no client IP in it to recover. dj-site would have to thread one through, which reopens the header-trust question BS#774 closed.
- **Raise `max`.** Buys headroom proportional to nothing: the bucket is shared by the whole population, and the trigger is the never-resets-while-busy semantics, not the number 100.
- **Edge-cache `get-session`.** Wrong on a session-validity read.

## Acceptance criteria

- [ ] `/auth/get-session` is no longer bucketed by `x-real-ip` for cookie-bearing requests; two different DJs behind one Cloudflare edge have independent budgets.
- [ ] `/auth/token` and `/auth/sign-in*` bucketing are unchanged — pin with a test.
- [ ] A cookie-less request to `/auth/get-session` still falls back to IP bucketing.
- [ ] `GET /auth` 429 rate over a 7-day window after deploy is under 0.1% (baseline: 4.5% month-average, 14.1% worst day).
- [ ] Whatever limiter is added emits something observable — a log line, metric, or Sentry breadcrumb — so the next occurrence is not invisible.

## Notes for implementer

- Confirm the path attribution yourself before changing anything: `sudo zcat /var/log/nginx/access.log-*.gz | awk '$9==429 {print $7}' | sort | uniq -c | sort -rn`. **logrotate names files by rotation date, so `access.log-20260814.gz` contains Aug 13 traffic** — an off-by-one day here will make the numbers disagree with Sentry.
- `rateLimitKeyFromRequest` returns the constant `'unknown'` when `x-real-ip` is absent and `socket.remoteAddress` is undefined. It cannot fire today because nginx always sets the header, but it would put every affected caller in one bucket if that ever changed — worth a guard while you are here.
- nginx logs the default `combined` format and does not record `CF-Connecting-IP`, so whether that header even reaches the origin on the proxied-browser arm is **unverified**. Adding `$http_cf_connecting_ip` to `log_format` would settle it in one deploy.
- The `/auth/ok` healthcheck self-429 theory was investigated and **ruled out**: that fetch is loopback, never traverses nginx, and the canary's 5-minute cadence exceeds the 10 s idle reset. No 429s appear on any healthcheck path.

## Related

- WXYC/dj-site — companion issue for the `session-cache.ts` error-swallowing that turns this throttle into a logout. Worth fixing independently of this.
- WXYC/wxyc-ios-64#948 — separate iOS auth bug that produces the `/auth/sign-in/anonymous` 429 bursts (29 in 30 days; small, and not this).

jakebromberg:
Closes #2169

## The defect

`GET /auth/get-session` returns 429 in bursts in production — 2,128 rejections on 2026-08-13, or 14.1% of that day's 15,126 requests to the path, against a 4.5% month-average baseline. Two causes compound.

better-auth's internal limiter keys on `${ip}|${path}`, and 100% of `/auth/get-session` traffic reaches us from Cloudflare egress IPs (dj-site's SSR arm is an edge-side fetch carrying only a cookie). So the entire DJ population shares roughly ten buckets of 100 rather than getting a bucket each — the IP in the key has no discriminating power at all.

Worse, that limiter is not a fixed window. It resets on time since the last *allowed* request and leaves the counter untouched on the denied branch, so a continuously-busy key accumulates monotonically to `max` and then 429s regardless of the actual request rate. Recovery requires ~10 s of complete silence on the shared key, which a shared key never gets. A client retrying a 429 at 1–2 s will simply 429 again.

The user-visible result was not a throttle. dj-site collapsed the failed session read into "no session" and redirected the DJ to `/login?bounced=no-session` — a silent logout mid-show.

## What changed

**1. better-auth stops rate-limiting this one path.** `shared/authentication/src/auth.definition.ts` gains a top-level `rateLimit: { customRules: { '/get-session': false } }` block. The rule key is `/get-session`, not `/auth/get-session` — better-auth's `normalizePathname` strips the basePath before matching. A `false` rule short-circuits `resolveRateLimitConfig` and returns before storage is touched. This block does **not** enable rate limiting or change `window`/`max`: those keep their `create-context` defaults (`enabled: ?? isProduction`, 10 s / 100), so nothing else in the auth service changes behavior.

**2. Two Express limiters replace it,** mounted in order on `/auth/get-session` inside `apps/auth/app.ts`'s existing `if (!isTestEnv)` block:

| | Key | Budget | `standardHeaders` | Purpose |
|---|---|---|---|---|
| `getSessionIpRateLimit` | `rateLimitKeyFromRequest` (bare `x-real-ip`) | 600 / min | off | Abuse ceiling |
| `getSessionIdentityRateLimit` | `sessionRateLimitKeyFromRequest` | 120 / min | `draft-7` | Fairness |

`sessionRateLimitKeyFromRequest` (new, in `apps/auth/rate-limit-key.ts`) resolves in precedence order: `Authorization: Bearer <token>` → `` bearer:${sha256(token).slice(0,16)} ``; session cookie → `` session:${sha256(value).slice(0,16)} ``; neither → `` ip:${clientIp} ``. It hashes rather than stores the credential because both are live and express-rate-limit keys surface in logs and error paths. It reads the `Cookie` header directly with a minimal scan — `apps/auth` mounts no `cookie-parser`, so `req.cookies` is `undefined`, and adding one just to serve one route's key function would parse cookies on every request to the service. Both cookie names are checked (`__Secure-better-auth.session_token` first, then the unprefixed form) because `advanced.defaultCookieAttributes.secure` is `NODE_ENV === 'production'` and better-auth only applies the `__Secure-` prefix when `secure` is set. The bearer arm is not optional: `bearer()` is a registered plugin and `GET /get-session` is exercised with a bearer token by the device-authorization integration spec, so a cookie-only key would leave the whole device/iOS arm sharing the ~10 Cloudflare buckets this change exists to eliminate.

**3. A generic buffered CloudWatch emitter,** extracted to a new `@wxyc/observability/metrics` subpath (`shared/observability/src/metrics.ts`) from the buffer/coalesce/flush machinery duplicated between `apps/backend/middleware/responseMetrics.ts` (BS#845) and `apps/backend/services/sse/sse-metrics.ts`. `apps/auth/auth-rate-limit-metrics.ts` consumes it to publish `WXYC/AuthService` `RateLimited` from both limiters' rejection handlers, alongside a Sentry breadcrumb (a breadcrumb, not `captureMessage`/`captureException` — a per-request Sentry event on a rate-limit path is exactly the flood the org's Sentry-quota history argues against). `AUTH_RATE_LIMIT_METRICS_DISABLED=true` opts out; `PutMetricData` rejections are logged and swallowed, never blocking a response.

**4. A stale comment corrected.** The `/healthcheck` loopback fetch's `X-Real-IP: 127.0.0.1` header carried a rationale from #765 claiming an unresolvable client IP "silently disables its internal rate limiter for the rest of the process lifetime." That is no longer true at the installed `better-auth@1.6.26` — the string `Rate limiting skipped` does not exist anywhere in the package. Current behavior buckets every IP-less caller into a single shared `no-trusted-ip|<path>` bucket; only the warning *log* latches to one-shot. That is strictly worse than "disabled." The header stays and is still required; the stated reason is now correct.

## Why there are two limiters, not one

This is the most likely thing to read as over-engineering. It isn't, and dropping the IP-keyed ceiling would be a net security regression.

**Neither the session cookie nor the bearer token is verified at `keyGenerator` time.** Express runs the key function long before better-auth validates anything. So a caller sending a fresh random `Authorization: Bearer <nonce>` on every request mints a brand-new identity bucket every request and is *never* throttled by the identity limiter — while each request still costs the auth service a DB session lookup. Identity keying alone would therefore remove the only DoS bound on a DB-touching endpoint and replace it with one an attacker sidesteps in a single line of curl.

The `apps/backend/middleware/rate-limit-key.ts` precedent (BS#1127) keys on `req.auth.id`, a *verified* identity resolved by upstream middleware. There is no equivalent here. Identity keying buys fairness; it cannot carry the abuse bound. Two concerns, two limiters:

- **Identity-keyed, 120/min — fairness.** Each DJ gets their own budget; one DJ cannot exhaust another's. This is the fix for the reported bug. 120/min is far above any human navigation rate (dj-site's SSR reads dedup through React `cache()` to roughly one `getSession` per navigation) while still bounding a stolen-credential replay.
- **IP-keyed, 600/min — abuse ceiling.** Bounds fabricated-credential floods and preserves a DoS bound on the DB lookup.

600/min per Cloudflare egress IP does not reintroduce the original defect. The defect was never the number 100 — it was better-auth's never-resets-while-busy semantics, where a continuously-active key accumulates monotonically and can never clear. express-rate-limit is a true fixed window that resets unconditionally every 60 s. Against observed traffic (low thousands/day org-wide, single-digit requests/minute across ~10 edges) 600/min/edge is roughly two orders of magnitude of headroom.

**One ceiling worth stating rather than discovering:** when a request carries neither cookie nor bearer, the identity key falls back to the Cloudflare edge IP and the ceiling keys on the same value, so both limiters bucket that traffic identically and the tighter one wins. Anonymous `GET /auth/get-session` (signed-out SSR, pre-cookie cold loads) therefore shares **120/min/edge**, not 600. That is structurally the same shared bucket this fix eliminates, just with a larger number and ~100× headroom against observed volume. AC3 requires the IP fallback, so skipping the identity limiter on credential-less requests is not an option; if this arm ever gets tight, raise its limit rather than removing the fallback.

`standardHeaders` deliberately differs between the two. express-rate-limit writes the draft-7 `RateLimit`/`RateLimit-Policy` headers unconditionally per limiter and both default `requestPropertyName` to `'rateLimit'`, so if both advertised draft-7 the identity limiter's 120 would overwrite the ceiling's 600 and `req.rateLimit` would reflect only whichever ran last. Turning headers off on the invisible backstop resolves both. Each limiter also supplies a custom `handler`, which replaces express-rate-limit's default entirely — so the handler sends `options.message` itself, preserving the one 429 body shape the other two auth limiters get from the default path.

## For reviewers

**`/auth/token` and `/auth/sign-in*` are deliberately untouched and stay IP-keyed.** They are brute-force surfaces where per-IP bucketing is the point, and their `keyGenerator` (`rateLimitKeyFromRequest`) is unchanged. `tests/unit/authentication/rate-limit-config.test.ts` pins this over source text: it asserts `customRules` contains `'/get-session': false` and that the `customRules` object literal contains no key matching `/token` or `/sign-in`. The negative assertion is scoped to the captured object literal rather than the whole file — `/auth/token`, `sign-in`, and `'/device/token'` all appear elsewhere in `auth.definition.ts` and would false-fail a whole-file regex on day one.

**The metric is dimensioned-only. There is no dimensionless companion, on purpose.** Per `WXYC/CLAUDE.md`'s CloudWatch Metric & Alarm Conventions, a companion datum exists solely so a plain-form `PutMetricAlarm` has a series to query; it doubles cost, pollutes the namespace, and invites a future misconfigured alarm otherwise. There is no alarm on `WXYC/AuthService RateLimited` yet — AC4 is verified by an nginx `zcat` grep, not an alarm — so the companion would have no consumer today. A `wxyc-canary` alarm is a filed follow-up, and the companion ships when the alarm does, which is the order `MutationClientError` and wxyc-canary#17 followed. The in-repo precedent agrees: `sse-metrics.ts` ships `EventsBroadcast` dimensioned-only and adds companions only for the two metrics that have alarms. `tests/unit/observability/metrics.test.ts` asserts the *absence* of the empty-`Dimensions` datum, so the follow-up alarm PR has to consciously flip it.

**Why the emitter is a subpath and never a barrel re-export.** `shared/observability/src/index.ts` is loaded eagerly at Sentry preload in both containers via `node --import ./dist/instrument.js`. Re-exporting an AWS-SDK-backed emitter from that barrel would pull `@aws-sdk/client-cloudwatch` into process preload in both images. Adding the subpath took four coordinated edits, not just the `exports` map: a tsup `entry` addition, a tsup `external` addition, a `jest.unit.config.ts` `moduleNameMapper` entry (the anchored bare-specifier mapping does not also match a subpath), and a `tests/tsconfig.json` `paths` entry (ts-jest compiles with `moduleResolution: "Node"`, which ignores `exports` maps entirely). All four are documented in the `@wxyc/observability` row of CLAUDE.md so the next subpath doesn't rediscover them.

**Why all the tests are unit tests.** This bug is structurally invisible outside production. `NODE_ENV=test` sets `isTestEnv` in `app.ts` (so the Express limiters aren't mounted) and better-auth's `enabled ?? isProduction` resolves false, and `getIp` returns `LOCALHOST_IP` in test/dev regardless. No integration test can reproduce a 429 here. The tests target the pure key function, the config object, and the emitter; the two source-text assertions were each observed failing against pre-change source before being committed.

## Acceptance criteria

| Criterion | How it is met |
|---|---|
| `/auth/get-session` no longer bucketed by `x-real-ip` for cookie-bearing requests | `customRules: { '/get-session': false }` removes better-auth's IP bucket; `sessionRateLimitKeyFromRequest` keys cookie-bearing requests on `session:<hash>` and bearer-bearing ones on `bearer:<hash>`. Covered by `tests/unit/auth/session-rate-limit-key.test.ts` (same cookie → identical key, different cookies → different keys, both cookie-name shapes, bearer beats cookie). |
| `/auth/token` and `/auth/sign-in*` unchanged, pinned by a test | `rateLimitKeyFromRequest` and both existing limiter mounts are untouched. `tests/unit/authentication/rate-limit-config.test.ts` asserts the `customRules` literal contains no `/token` or `/sign-in` key. |
| Cookie-less request still falls back to IP bucketing | `sessionRateLimitKeyFromRequest` returns `ip:<x-real-ip>`, then `ip:<socket.remoteAddress>`, then `ip:unknown`. Three tests cover the three rungs. |
| `GET /auth` 429 rate under 0.1% over 7 days post-deploy | **Measured, not implemented** — see Post-deploy verification below. |
| The limiter emits something observable | `WXYC/AuthService` `RateLimited` (dimensions `Limiter` ∈ `ip`\|`identity`, `KeyKind` ∈ `session`\|`bearer`\|`ip`, `Route`) plus a Sentry breadcrumb, from both limiters' handlers. `tests/unit/auth/auth-rate-limit-metrics.test.ts` covers namespace, metric name, dimension set, and the `AUTH_RATE_LIMIT_METRICS_DISABLED` short-circuit; `tests/unit/observability/metrics.test.ts` covers buffering, flush-on-interval vs flush-on-size, coalescing, and swallowed `PutMetricData` rejections. |

Also worth noting for the reviewer's mental model: `KeyKind` is classified by reading the prefix `sessionRateLimitKeyFromRequest` already computed, not re-derived, so it cannot drift out of sync with that function's precedence. It is applied to *both* limiters' rejections — a `KeyKind: 'bearer'` datum on an `ip`-limiter rejection is the fabricated-credential-flood signal, not a mislabel.

## Post-deploy verification

The 429-rate criterion is measured after deploy, not asserted in CI. On the EC2 host, re-run the issue's own command and compare against the 4.5% month-average / 14.1% worst-day baseline:

```
sudo zcat /var/log/nginx/access.log-*.gz | awk '$9==429 {print $7}' | sort | uniq -c | sort -rn
```

logrotate names files by rotation date — `access.log-20260814.gz` holds Aug 13 traffic. An off-by-one there will make the numbers disagree with Sentry.

Cross-check against the new `WXYC/AuthService` `RateLimited` metric once it has a full window of data; the two should agree. One pre-flight before trusting the metric: the emitter follows `responseMetrics.ts` in constructing `CloudWatchClient` without explicit credentials, and the AWS default provider chain resolves environment-variable credentials *before* the EC2 instance role. Both containers start from the same `--env-file .env`, which carries the SES keys, so the emitter authenticates as the **SES IAM user**, not the instance role. Confirm that principal has `cloudwatch:PutMetricData` — the cheapest check is that `apps/backend`'s `MutationClientError` (identical credential-free pattern, identical env file) is landing in prod today. If it isn't, `AUTH_RATE_LIMIT_METRICS_DISABLED=true` is the safe landing state and the metric ships in a follow-up.

Two standing caveats, documented rather than solved: both limiters use process-local in-memory stores with no `secondaryStorage`, so if `apps/auth` is ever scaled horizontally both ceilings silently multiply by N (there is no Redis in this stack); and the `customRules === false` short-circuit plus the `/get-session` path key after basePath stripping are behaviors of a minified upstream build, which a `better-auth` bump could reshuffle — the source-text test is the tripwire.

## Related

- **Client-side half:** WXYC/dj-site#1237 (issue WXYC/dj-site#1225) makes dj-site classify a failed session read instead of collapsing it into "no session" — the half that turns a server 429 into a visible throttle rather than a silent logout. The two are independent; neither blocks the other.
- The plan this implementation followed is committed at `plans/bs2169-get-session-limiter-key.md`.

### Follow-ups deliberately not absorbed here

Filed separately rather than widened into this PR:

- **#2190** — the `'unknown'` fallback in `apps/auth/rate-limit-key.ts`. Same class of shared-bucket defect, much lower volume, but a different blast radius: that function is the `keyGenerator` for *both* existing brute-force limiters, so rekeying it conflicts with AC2 here.
- **WXYC/wxyc-canary#97** — an alarm on `WXYC/AuthService RateLimited`. Filed in `wxyc-canary` because that repo's `template.yaml` owns the org's cross-namespace alarm definitions (it already alarms on `WXYC/BackendService MutationClientError`). The issue records the load-bearing prerequisite: the alarm needs the dimensionless companion datum flipped on in this repo *first*, or it sits at `INSUFFICIENT_DATA` forever.
- **#2191** — migrating `responseMetrics.ts` and `sse-metrics.ts` onto the extracted `@wxyc/observability/metrics` emitter. The emitter was extracted from their duplicated machinery but neither was migrated here, so the repo is transiently at three copies instead of two; keeping this PR near the 1000-line guideline was the trade.
- **#2193** — the abuse ceiling keys on the full `X-Real-IP` rather than the /56 prefix `express-rate-limit`'s own `ipKeyGenerator` uses, so an IPv6 client could mint unlimited ceiling buckets. Raised in review here. Pre-existing on that generator but newly load-bearing, since this ceiling is now the only IP-based bound on the path. Not fixed inline for two reasons: whether nginx can hand us an IPv6 `X-Real-IP` depends on host config that is not in this repo, and rekeying the generator also rekeys the two older brute-force limiters that AC2 requires be left untouched. The open assumption is recorded in the comment above the ceiling.


jakebromberg:
Code review pass done. No HIGH findings; the two-limiter design and the dimensioned-only metric decision both held. Six findings fixed in 8fccc8da, one deferred with a filed issue, two closed as no-action. CI green on the follow-up commit.

**Fixed — `@wxyc/observability/metrics`**

1. **`flush()` didn't await an in-flight size-triggered send.** `record()` drains the buffer synchronously before awaiting `PutMetricData`, so a later `flush()` saw length 0 and resolved while the send was pending — contradicting the interface doc and, concretely, letting a shutdown hook exit before the batch landed. Overlapping flushes now chain through an `inFlight` handle, so the returned promise covers both the pending send and anything still buffered. The test that worked around this ("yield once for the synchronous flush kicked off inside `record()`") is gone, replaced by one that asserts the contract with a deliberately-blocked send.
2. **`coalesceKey()` omitted `unit`, and the companion flag was first-wins.** A `Milliseconds` datum and a `Count` datum sharing a name and dimension set summed into one value stamped with whichever arrived first; and an opted-out record landing first suppressed the dimensionless companion an alarm would query. `unit` is now part of the key and `emitDimensionlessCompanion` is OR-ed across the group. Inert at today's single call site, but this is a shared package whose own docs say each call site owns its units. Both cases now have tests.

**Fixed — `apps/auth`**

3. **No flush on shutdown.** SIGTERM/SIGINT closed the DB but never flushed the emitter, and the flush timer is `unref`'d on a 30 s interval — so every deploy dropped up to a full window of `RateLimited` points, precisely the window worth having during a rollout that trips the limiters. `flushRateLimitMetrics()` now runs before `closeDatabaseConnection()`, internally bounded at 2 s and never rejecting, so it cannot delay or wedge shutdown.
4. **The abuse ceiling emitted no `Retry-After`.** `setRetryAfterHeader` is gated on `legacyHeaders || standardHeaders`, and the ceiling sets both false. That flag was justified purely on the draft-7 header collision, and the reviewer is right that the rationale doesn't extend to `Retry-After` — a client doing `Retry-After` backoff would back off correctly on a fairness rejection and hot-loop against the ceiling. The custom handler now sets it explicitly from `windowMs`, so both limiters return the same 429 contract in headers as well as body, while the ceiling still advertises no `RateLimit` budget of its own.
5. **Missing the `writableEnded` guard** that supplying a custom handler had replaced. Latent, but its absence turns "something already responded" from a silent no-op into `ERR_HTTP_HEADERS_SENT`. Added, with a test.
6. **`extractBearerToken` didn't trim its greedy capture**, so `Bearer abc ` and `Bearer abc` keyed to different buckets for the same token. Trimmed; an all-whitespace remainder now falls through to the IP arm.

**Deferred — #2193**

The ceiling keys on the full `X-Real-IP` rather than the /56 prefix `express-rate-limit`'s `ipKeyGenerator` would use, so an IPv6 client holding a /64 could mint effectively unlimited ceiling buckets. (Its own `keyGeneratorIpFallback` warning stays silent only because our generator's source text doesn't mention `req.ip` — we have the property the warning exists to catch.)

Not fixed here, deliberately. Whether an IPv6 `X-Real-IP` is reachable in this deployment depends on the host's nginx `listen`/`real_ip` configuration, which lives at `/etc/nginx/nginx.conf` on the EC2 box and is not in this repo — so it can't be settled from source, and narrowing a security-relevant key space on an inference is the wrong move. Independently, this generator is also the `keyGenerator` for the sign-in/sign-up/device and `check-request-ban` limiters, so rekeying it silently rekeys both, which directly conflicts with this PR's AC2. #2193 carries the investigation steps and both outcomes; the open assumption is recorded in the comment above the ceiling so the next reader doesn't mistake it for settled.

**No action**

- Cookie/bearer parsing edges are INFO. The reviewer confirmed no cross-user amplification — a fabricated bearer buys the attacker their own bucket, never the victim's. Duplicate same-name cookies resolving to first-in-header is inherent to the header format.
- `@aws-sdk/client-cloudwatch` bundling into `dist/app.js` matches `apps/backend`'s existing pattern; prod Docker stage and lockfile verified fine, and the `instrument.ts` isolation argument for the subpath export holds (`splitting: false`, barrel doesn't re-export metrics).

Also declined: function-valued `message` support in the handler. `message` is a static object literal defined adjacent to both limiters, so the branch would be unreachable code.

Full local gate re-run before pushing — typecheck clean, lint 0 errors, `format:check` clean, 445 suites / 7534 tests passing.

PR #2192, opened 2026-08-17
docs(migrations): record that migrate() applies all pending migrations in one transaction

### LINKED ISSUE #2175: docs/migrations.md: record that migrate() runs all pending migrations in ONE transaction
`docs/migrations.md` is the canonical home for migration rules, and it is missing one that just cost a CI cycle on #2174.

## The rule

`dev_env/init-db.mjs` calls drizzle-orm's programmatic `migrate()`, which applies **all pending migrations in a single transaction**. A migration is therefore not isolated from its predecessors on a fresh database, and two consequences follow:

1. **An enum value added by `ALTER TYPE ... ADD VALUE` cannot be *used* by a later migration** on a fresh database, because both land in the same transaction. Postgres rejects it with `unsafe use of new value "X" of enum type Y`. Compare via `column::text = 'X'` instead, which never constructs the value.
2. More generally, any "object created by an earlier migration" assumption that Postgres gates on transaction boundaries is unsafe on a fresh database, and safe against production.

## Why it needs writing down

**The failure is invisible to the path that looks highest-stakes.** Against production — and against the `Migration Dry-Run (prod-shaped data)` job's RDS snapshot — the earlier migration committed long ago, so only the new file is pending and the bare form works. It breaks only on a *fresh* database: CI's `Integration-Tests`, a new clone, a local `db:start`.

That inverts the usual intuition, where the dry-run is the strictest gate. For this class of defect it is structurally blind, and `Integration-Tests` is the only job that catches it.

## Also worth a line: how to verify locally

Applying migration files one at a time through `psql` (the obvious approach) **autocommits between files** and therefore cannot reproduce this. A faithful local check concatenates every journalled migration into one explicit `BEGIN`/`COMMIT` and applies that to a fresh database:

```bash
{ echo 'BEGIN;'
  for tag in $(jq -r '.entries[].tag' shared/database/src/migrations/meta/_journal.json); do
    cat "shared/database/src/migrations/$tag.sql"; echo ';'
  done
  echo 'COMMIT;'; } | psql -v ON_ERROR_STOP=1 -d <fresh-db>
```

Pair it with the existing PG14-vs-PG18 guidance already in the doc (BS#1424) — prod RDS is 14.22, so the fresh-database check should run against `postgres:14`, not the PG18 dev container.

## Acceptance criteria

- [ ] A `@rule`-annotated section in `docs/migrations.md` covering the single-transaction semantics and the enum-value corollary.
- [ ] The local verification recipe above, or a script under `scripts/`.
- [ ] Cross-reference from the enum guidance if any exists.

Worked example to link: [#2174](https://github.com/WXYC/Backend-Service/pull/2174) / migration `0150`, where the bare comparison passed the dry-run and failed Integration-Tests.

jakebromberg:
## The problem

`dev_env/init-db.mjs` (`runMigrations`) and `scripts/dryrun-migrate.mjs` both call drizzle-orm's programmatic `migrate()` from `drizzle-orm/postgres-js/migrator`, not the `drizzle-kit migrate` CLI. That function wraps **every pending migration in one transaction**, not one transaction per file. A migration is therefore not isolated from its predecessors on a fresh database, and Postgres restrictions that key on transaction boundaries — most sharply, `ALTER TYPE ... ADD VALUE` followed by a bare comparison against the new value — fire on a fresh database while passing everywhere the predecessor already committed.

The consequence worth writing down is not the enum rule itself; several migration headers already record that. It is **where the defect is caught**. Against production, and against the `Migration Dry-Run (prod-shaped data)` job's restored RDS snapshot, a migration committed in an earlier deploy is not pending, so the bare form works and the dry-run passes. The break happens only on a genuinely fresh database: CI's `Integration-Tests`, a new clone, a local `npm run db:start`. That inverts the usual reading of the job list, where the prod-shaped dry-run looks like the strictest gate — for this class of defect it is structurally blind, and `Integration-Tests` is the only job that sees it.

## What the rule says

One new `@rule id=single-transaction-migrate` section in `docs/migrations.md`, docs-only, no other file touched. It covers: the single-transaction semantics and which two callers exhibit them; the enum-value corollary and its fix (`column::text = 'X'` rather than `column = 'X'::enum_type`, since the text cast never constructs the enum value); the general "an earlier migration already created this" hazard beyond enums; the precise shape of the dry-run's blind spot; and a local verification recipe that concatenates every journalled migration into one explicit `BEGIN`/`COMMIT` against a disposable `postgres:14` container, since applying files one at a time through `psql` autocommits between them and cannot reproduce the failure at all.

The blind spot is stated narrowly on purpose. It is **not** "the dry-run can't see enum misuse" — the dry-run applies exactly what is still pending against a prod snapshot, so an `ADD VALUE` and a consumer of that value introduced in the same PR are both pending there, inside one transaction, and are caught. What it cannot see is the **cross-deploy** case: the predecessor committed in an earlier deploy and is not pending against prod at all, while a fresh database still has it pending alongside the new file.

## Measured, not inferred

Three boundary conditions in the rule were measured against a disposable `postgres:14.24` container (prod RDS is 14.22) rather than reasoned about, because the real transaction boundary is narrower than the intuitive reading, and the rule text was corrected to match the measurements:

- A value added by `ALTER TYPE ... ADD VALUE` is rejected in the same transaction **even when the enum type itself was created earlier in that same transaction**. Only a value present in the original `CREATE TYPE` is usable there.
- Therefore the dry-run's blind spot is the cross-deploy case specifically, not enum misuse in general — the same-PR case is caught.
- Migration `0112`'s header comment asserts that fresh CI databases never reproduce the error because `0091` creates the type in the same batch. The first measurement says that reason does not hold; a fresh database fails there too. The rule flags this explicitly rather than quietly contradicting it.

`0112.sql` itself was **not** modified — correcting a migration file's header comment is out of scope for a docs-only change, and a migration already applied in prod is not a file to touch casually. The rule tells a reader to read 0112's version with care and says which half of it is wrong.

## Worked example

PR #2174 / migration `0150_reclassify-and-drop-fictional-n-bin.sql`: a bare `rotation_bin = 'N'` comparison passed the dry-run and failed `Integration-Tests`. `freq_enum` is created in `0000` and gains `'N'` in `0041` — both long committed in prod, both still pending on a fresh database. That is exactly the cross-deploy shape the rule describes, and it cost a CI cycle to find.

## Checks

`npm run format:check` and `npm run lint:migrations` pass. `check:doc-rules` and `check:doc-budget` (both warn-only) report no new findings attributable to this rule.

Closes #2175


jakebromberg:
### Code review

Found 1 issue:

1. The rule's citation list vouches for two migrations whose header comments state the exact transaction model the rule exists to correct, while singling out only `0112` for correction. The new text says "see 0086, 0094, 0109, 0112, and 0117 for prior write-ups of the same restriction", then corrects `0112` alone. But `0117` asserts "Drizzle wraps each migration in a transaction. Isolating the ADD VALUE guarantees it is committed before any later migration or app code references it" — both halves are what this rule refutes, and the second is precisely the mechanism that broke 0150. `0094` states the same premise ("Drizzle's transaction-per-migration shape"), though its conclusion happens to be safe because its consumer writes at runtime rather than migration time. `0109` and `0086` scope their claims to within their own file and are fine, so the implicit endorsement is accurate only for those two. A future author sent to `0117` as trustworthy prior art learns the belief that caused #2174. The fix is one clause in the new prose extending the "read with care" caveat to `0117` (and ideally `0094`) — in scope for a docs-only change, unlike editing the `.sql` files.

https://github.com/WXYC/Backend-Service/blob/d8fd07b58c68e772fbfec1678319554f129c252e/docs/migrations.md#L119-L121

https://github.com/WXYC/Backend-Service/blob/d8fd07b58c68e772fbfec1678319554f129c252e/shared/database/src/migrations/0117_recheck-after-unavailable-enum.sql#L11-L14

https://github.com/WXYC/Backend-Service/blob/d8fd07b58c68e772fbfec1678319554f129c252e/shared/database/src/migrations/0094_rotation-lml-identity-id.sql#L69-L75

Also checked and cleared: every cited path, script, and migration number resolves (`dev_env/init-db.mjs`, `scripts/dryrun-migrate.mjs`, and 0000/0041/0086/0091/0094/0109/0112/0117/0150); the `jq` filter in the verification recipe returns all 148 journal tags and each resolves to a `.sql` file; no migration contains an executable `CREATE INDEX CONCURRENTLY` (all 32 occurrences are inside comments per the out-of-band pattern), so concatenating the set into one `BEGIN`/`COMMIT` is actually viable; and the rule's characterization of `0112` — correct on transaction scope, wrong on fresh CI databases — matches that file verbatim.

One sub-threshold note, not blocking: `incidents=#2174` cites a pull request where `docs/migrations.md`'s convention line says "incident issues (`#NNN`)". `docs/deploy.md` already uses `#run-<actions-run-id>` values, nothing parses the field (`scripts/check-doc-rules.mjs` never reads it), and the prose says "PR #2174" explicitly, so this is a convention nit rather than a defect.

jakebromberg:
### Code review (addendum)

A later pass surfaced 2 further issues, both above the confidence bar:

2. The production-safety dichotomy is overstated — the repo's own deploy doc records an incident class where **production** runs several pending migrations in one transaction. The rule frames this as fresh-database-only ("while being safe against production", "It breaks only on a genuinely _fresh_ database"), but `deploy-cadence-24h` in `docs/deploy.md` says a silent or failed auto-deploy means "subsequent migration PRs land on top of the unapplied chain", and cites the 2026-05-04 wedge where "4 days of accumulated migrations (0071, 0072, 0073) compounded". A stacked deploy is exactly the multi-pending-migration condition this rule describes, so the cross-deploy form is reproducible in prod, not merely on a fresh DB. Sharpening this matters because the framing drives risk assessment: an author shipping an `ADD VALUE` consumer during a wedge would read the rule as saying prod is fine.

Reinforcing that: `dev_env/init-db.mjs` is not a dev-only tool. It is the production migration runner — `Dockerfile.migrate` copies it and its `CMD` is `node init-db.mjs`, and `deploy-base.yml`'s `migrate` job builds that image and `docker run`s it with `SKIP_SEED=true`. The rule lists it alongside `scripts/dryrun-migrate.mjs` and then contrasts both against "production", which invites reading it as dev/CI-only when it is the same file every deploy executes.

https://github.com/WXYC/Backend-Service/blob/d8fd07b58c68e772fbfec1678319554f129c252e/docs/deploy.md#L20-L22

https://github.com/WXYC/Backend-Service/blob/d8fd07b58c68e772fbfec1678319554f129c252e/Dockerfile.migrate#L28-L30

3. The same document, four rules above, states the claim this rule refutes — and it is left uncorrected. `if-not-exists-index` says the in-migration form is not CONCURRENTLY because "Drizzle wraps each migration in a transaction", which is precisely the "not one transaction per file" model the new rule overturns. That rule's practical conclusion still holds (a shared transaction is still a transaction, so CONCURRENTLY remains illegal), so nothing is functionally broken — but for a documentation change whose entire payload is correcting this mental model, leaving the contradiction inside the same file undercuts it. The same phrasing recurs in `0064`, `0078`, `0083`, `0098`, and `0117`; a parenthetical in the new rule noting that the older "each migration" phrasing is imprecise but harmless for the CONCURRENTLY case would reconcile them without touching any `.sql`.

https://github.com/WXYC/Backend-Service/blob/d8fd07b58c68e772fbfec1678319554f129c252e/docs/migrations.md#L59-L61

jakebromberg:
Addressed in fa0a1707. All three findings verified independently against source before fixing, not taken on report.

**Finding 1 (the rule vouching for text that teaches the refuted model) — confirmed and fixed.** `0117` does state "Drizzle wraps each migration in a transaction" and that isolating the `ADD VALUE` "guarantees it is committed before any later migration ... references it"; that second clause is exactly the 0150 failure. `0094:71` invokes the same "transaction-per-migration shape". The cross-reference now says which of the five to read against this rule and which (0086, 0109) scope their claims correctly, and it keeps the practice of splitting `ADD VALUE` into its own file while dropping the guarantee that practice was thought to buy — it holds for every *later* deploy, not for the one that introduces it.

**Finding 2 (the prod dichotomy) — confirmed and fixed, and it is the substantive one.** `Dockerfile.migrate:29` is `CMD ["node", "init-db.mjs"]` and `deploy-base.yml:564` builds that image, so `init-db.mjs` is the production runner despite living under `dev_env/`. And `docs/deploy.md:21-23` documents the 2026-05-04 wedge with 0071/0072/0073 stacked unapplied. So "safe against production" was wrong: the boundary is a property of what is pending, not of the environment. Reworded to say so, and to say the migration should not depend on the boundary at all.

**Finding 3 (self-contradiction at `docs/migrations.md:60`) — confirmed and fixed.** Reworded to the shared transaction, with a note that the `CONCURRENTLY` conclusion is unaffected either way.

`format:check`, `lint:migrations`, and `check:doc-rules` all clean; the new rule raises no doc-rules finding.

One thing left deliberately out of scope: `0117`, `0094` and `0112` still carry the incorrect comments in their own files. Correcting migration SQL is not something to fold into a docs PR, so the rule now tells the reader how to read them instead. Worth a follow-up ticket.

Issue #2217, opened 2026-08-18
Track-context trust gate rejects LML row-less resolutions of the requested album (typed `alternative`, indistinguishable from a substitution)


jakebromberg:
## Problem

`isTrustedLmlTrackContextMatch` rejects LML's **row-less non-library resolutions** — results that carry the *exactly requested* album with real Discogs artwork — because they arrive typed `search_type: "alternative"`, the same type LML uses for a genuine same-artist substitution.

The gate ([`shared/lml-client/src/trust.ts#L98-L100`](https://github.com/WXYC/Backend-Service/blob/6f9efa821acf2b9232bc87fff590dc661f8905e5/shared/lml-client/src/trust.ts#L98-L100)) accepts only `direct` and `compilation`. `extractArtwork` therefore returns `null` ([`apps/enrichment-worker/enrich.ts#L484-L491`](https://github.com/WXYC/Backend-Service/blob/6f9efa821acf2b9232bc87fff590dc661f8905e5/apps/enrichment-worker/enrich.ts#L484-L491)) and the playcut is frozen at the terminal `metadata_status = 'enriched_no_match'`.

The gate's own doc-table justifies rejecting `alternative` as *"closest-match fallback, no track confirmation"*. That description is accurate for a substitution. It is **not** accurate for a row-less resolution, where LML resolved and validated the requested release off Discogs and returned it with `library_item.id = 0` (the [`_make_rowless_item`](https://github.com/WXYC/library-metadata-lookup/blob/bd3ea973c0458b349cecec0d12abf06203c562e0/lookup/rowless.py#L170) "no catalog row" marker). Both shapes are indistinguishable to the current gate, so the correct one is discarded along with the wrong one.

## Evidence (prod, 2026-08-18)

Replayed the 14 Section A rows of the `metadata-no-match-digest` for the 08-17 → 08-18 window against prod LML, in the worker's exact request shape (`POST /api/v1/lookup/bulk?allow_release_resolution_fallback=true`, `extended: true` per item — i.e. with BS#1815's opt-in, which the worker sets at `apps/enrichment-worker/lookup-batcher.ts`). No row was `timeout` or `degraded`; latencies 0.8–3.5s.

| Playcut | Asked album | `search_type` | `library_item.id` | Returned title | Artwork | Trust gate | Correct? |
|---|---|---|---|---|---|---|---|
[...]

PR #2219, opened 2026-08-19
fix(flowsheet-no-match-recheck): unlock the self-stalled queue via headerless lookups, an OFFSET cursor, and a newest-first tiebreak

### LINKED ISSUE #2218: flowsheet-no-match-recheck is self-locked on prod: 64 rows drained from a 137,340-row cohort in 5 runs, head not advancing past id 154
## Problem

`flowsheet-no-match-recheck` can make **zero forward progress indefinitely**. Four design choices compose into a self-lock, and the job's own purpose is what triggers it.

1. [`query.ts#L63`](https://github.com/WXYC/Backend-Service/blob/6f9efa821acf2b9232bc87fff590dc661f8905e5/jobs/flowsheet-no-match-recheck/query.ts#L63) — `ORDER BY f."no_match_recheck_attempted_at" ASC NULLS FIRST, f."id" ASC`
2. [`orchestrate.ts#L211`](https://github.com/WXYC/Backend-Service/blob/6f9efa821acf2b9232bc87fff590dc661f8905e5/jobs/flowsheet-no-match-recheck/orchestrate.ts#L211) — on a transient response the marker is left untouched, "so it is retried on the next pass"
3. [`lml-fetch.ts#L28`](https://github.com/WXYC/Backend-Service/blob/6f9efa821acf2b9232bc87fff590dc661f8905e5/jobs/flowsheet-no-match-recheck/lml-fetch.ts#L28) — `deadline_exceeded` is classified transient (BS#2179 review HIGH 2 / BS#1977)
4. [`lml-fetch.ts#L101`](https://github.com/WXYC/Backend-Service/blob/6f9efa821acf2b9232bc87fff590dc661f8905e5/jobs/flowsheet-no-match-recheck/lml-fetch.ts#L101) — the job "stays registered at LML class 5, which sends `X-Caller-Budget-Ms` unconditionally", arming LML's ~4s empty-state cutoff

A row that deadline-exceeds keeps `no_match_recheck_attempted_at = NULL`, so it stays in the `NULLS FIRST` group with an unchanged `id` and is re-selected on the very next run. If ≥ `BATCH_SIZE` (default 200) of the lowest-id never-attempted rows persistently deadline-exceed, the job re-reads the identical 200 rows every 6 hours forever and never reaches the rest of the cohort.

**The trigger is the job's own reason for existing.** Because it sends the budget header, `deadline_exceeded` is the *expected* response for exactly the population it targets: cold non-library release resolution, measured at 4–20s on 2026-08-04 against a ~4s cutoff. The rows most in need of a recheck are the rows most likely to jam the queue.

Individually all four choices are defensible — (2) and (3) are the correct fix for BS#1977's false-freeze, and (1) is fair queueing. The interaction is the defect.

## CONFIRMED on prod, 2026-08-18 ~18:55 PT

Measured — see [the measurement comment](https://github.com/WXYC/Backend-Service/issues/2218#issuecomment-5336587265) for full data.

- **64 stamped against a 137,340-row cohort** after five runs (~1,000 candidate slots) — a ~94% transient rate, rising to ~97% in the last two runs, with the stamp rate decaying 18 → 20 → 15 → 6 → 6.
- **Direct proof the head is re-ground:** next candidate set spans `id 154..2790`; the 65 stamped ids span `id 168..2741` — interleaved in the same band. After five runs the head has not moved off `id 154`. This evidence holds regardless of the prod `BATCH_SIZE`.
- **Mechanism confirmed from the data side:** the stuck head is entirely `add_time = 2004-11-04` rows (including a literal `ping test` row), and 5 of the first 12 have `album_title IS NULL`, forcing LML's track-only live-resolution path — the 4–20s class, against a ~4s cutoff.

## Second defect found by the same measurement: `ORDER BY id ASC` is the wrong priority

2026 playcuts are **4,880 of 137,340 (3.6%) and last** in `id ASC` order; 132,460 older rows sit ahead of them.

- At the observed marginal rate (24/day): **~15.1 years** to reach 2026.
- At the theoretical maximum if every candidate resolved (800/day): **~5.5 months**.

Even a perfectly functioning job spends five and a half months before touching a row anyone can see. The head is simultaneously the least valuable material and the most expensive (30% NULL-album in 2004, vs 2–5% in recent years). Fair queueing across the TTL rotation is right; the **never-attempted tiebreak** is what is wrong.

## Impact

The frozen `enriched_no_match` cohort is only ever re-asked by this job or by a one-shot drain. **If this job cannot advance, every answer-quality fix delivers nothing to the backlog** — BS#2217, WXYC/library-metadata-lookup#1228 and WXYC/library-metadata-lookup#1229 could all land and the backlog would not move. Only new plays (served by the live enrichment worker, which already has the budget header suppressed via BS#1978) would improve.

BS#1998's 26,286-row orphan cohort sits at the head of the ordering by `id ASC`. The 2026-08-18 digest rows are ids ~5,309,xxx, i.e. at the very back — so if the head starves they are never reached at all.

## Desired end state

The job provably advances through the whole cohort, and can actually resolve the cold-release class it was built for.

## Suggested approach — DECIDED 2026-08-18

Measure first, then fix the cause and add insurance:

1. **Measure** with the query above; record the result in this issue.
2. **Suppress the budget header** using BS#1914's `budgetMs: null` lever — the same mechanism BS#1978 applied to the live enrichment worker, which is what made cold non-library resolution work there. This removes the cause rather than masking it: `deadline_exceeded` stops being the expected answer, so rows resolve or earn a definitive no-match and the marker advances naturally.
3. **Add a cursor / offset** so a chronically-transient head can never starve the tail, independent of (2).
4. **Change the never-attempted tiebreak from `id ASC` to newest-first** (`add_time DESC`, or `id DESC`). Without this, (2) and (3) buy a job that works correctly and still needs ~5.5 months to reach 2026 playcuts.

All of (2), (3) and (4), not one. (2) alone leaves the queue vulnerable to any future chronic-transient condition; (3) alone guarantees progress but leaves the job unable to resolve the class it exists to serve — it would march through failing everything once, then back off behind the TTL.

**Side benefit of (2):** a budget-cut ask leaves a short-TTL known-miss pin in LML ([`lookup/rowless.py#L293`](https://github.com/WXYC/library-metadata-lookup/blob/bd3ea973c0458b349cecec0d12abf06203c562e0/lookup/rowless.py#L293), the LML#824 `crowd_out` variant, self-healing within the hour). Running 4×/day with the header, this job can pin misses that a *live* play then hits inside that window — the recovery job poisoning the live path. Suppressing the header removes that too.

### Cost to weigh

Suppressing the header hands an offline drain LML's full cascade (up to the 25s hard cap) for up to 200 rows/run × 4 runs/day, against the shared Discogs rate bucket (see the `discogs_rate_gate_fail_open` counter, WXYC/library-metadata-lookup#879). BS#1914's original premise was that fast-degrade *is* right for offline drains. That premise holds for drains whose rows are mostly cache-resolvable; it does not hold here, where the cohort is defined by having failed once already. If the load proves too high, the lever to reach for is batch size or run frequency, not the header.

Note BS#2179's review HIGH 3 ("withdrawn") concerned the per-call **timeout constant**, not whether the header should be suppressed. Different question; it does not appear to have been asked.

## Constraints

- Do **not** stamp the marker on transient responses to force progress — that re-introduces the false-freeze BS#1977 and BS#2179's review HIGH 2 just fixed, one TTL removed.
- `cache_only` must stay a definitive answer, not a transient (per `lml-fetch.ts`'s existing classification).
- Any cursor must not skip rows whose TTL has since expired — it is a starvation guard, not a replacement for the TTL rotation.

## Acceptance criteria

- [x] Measurement recorded in this issue — **done 2026-08-18, confirmed locked.**
- [ ] Budget-header suppression applied via the BS#1914 lever, with a test pinning that this caller sends no `X-Caller-Budget-Ms`.
- [ ] Cursor/offset added, with a test proving a batch that returns all-transient does not re-select the identical candidate set on the next run.
- [ ] Transient rows still leave the marker untouched (BS#1977 behavior preserved) — existing tests unchanged.
- [ ] After one full day of runs, `stamped` has increased by roughly `BATCH_SIZE × runs/day`.
- [ ] Never-attempted tiebreak changed to newest-first, with a test pinning that a fresh 2026 playcut is selected ahead of a 2004 one.
- [ ] `docs/ops-cron-scheduling.md` updated if the job's LML-load classification changes.

## Context

Found on 2026-08-18 while investigating the daily `metadata-no-match-digest`. The prior root cause of that digest — the class-5 budget header clamping LML to ~4s — was fixed for the **live worker** by BS#1978 (flipped on prod 2026-08-04 14:33 PT, `set-ec2-env-var` run 30952791886) and verified: digest volume fell 75 → 26, and a 14-row replay returned `timeout: false` / `degraded: false` at 0.8–3.5s. This job did not inherit that fix.

Agreed landing order: BS#2217 first and alone (fixes new plays immediately, and both the worker and this job read the same predicate), then this ticket, then BS#1979's one-shot drain, then the two LML tickets.

## Related

- WXYC/Backend-Service#2176 — created this job
- WXYC/Backend-Service#2179 — its review; HIGH 2 made `deadline_exceeded` transient, HIGH 3 (withdrawn) covered the timeout constant
- WXYC/Backend-Service#1977 — shed/timeout written as terminal; the defect (2)+(3) exist to prevent
- WXYC/Backend-Service#1978 / #1914 — the budget-suppression lever this ticket reuses
- WXYC/Backend-Service#1979 — one-shot drain; immune to this starvation, so it survives, but must land *after* BS#2217
- WXYC/Backend-Service#1998 — the 26,286-row orphan cohort sitting at the head of the ordering
- WXYC/Backend-Service#2217 — trust-predicate fix this job also inherits
- Follow-up worth filing once unblocked: 2004–2005 NULL-album rows may be permanently unresolvable and will consume budget under any ordering — consider an escalating TTL or an exclusion after k definitive failures.

- WXYC/library-metadata-lookup#824 / #1112 — the known-miss pin and the discarded-work shed behind the poisoning side effect

jakebromberg:
Closes #2218

`flowsheet-no-match-recheck` was measured on prod on 2026-08-18 making effectively zero forward progress: **64 rows stamped out of a 137,340-row cohort across five runs** (~1,000 candidate slots), with the stamp rate decaying 18 → 20 → 15 → 6 → 6, and the queue head never moving off `id 154` — a literal `ping test` row from 2004-11-04. The next candidate set spanned `id 154..2790` while the 65 rows stamped so far spanned `id 168..2741`: interleaved inside the same band, i.e. the job was re-grinding substantially the same batch every six hours.

Four individually-defensible choices composed into the self-lock. The retry marker is deliberately left untouched on a transient response (BS#1977 / BS#2179 review HIGH 2), `deadline_exceeded` is classified transient (same review), the ordering is `no_match_recheck_attempted_at ASC NULLS FIRST, id ASC`, and this caller sent `X-Caller-Budget-Ms` unconditionally as an LML class-5 registrant. Because the header's mere presence arms LML's ~4s empty-state cutoff, `deadline_exceeded` was the *expected* answer for precisely this cohort — rows already defined by having failed a live lookup once, disproportionately cold non-library releases that need LML's full 4–20s cascade. A row that transients keeps `attempted_at = NULL` and an unchanged `id`, so it holds its exact position in the ordering and is re-selected next run, forever.

The measurement also surfaced a second, independent defect: 2026 playcuts are 4,880 of 137,340 rows (3.6%) and sort **last** under `id ASC`, behind 132,460 older rows. Even a perfectly functioning job needed ~5.5 months to reach a row a listener can currently see, and the head it was grinding instead was both the least valuable material and the most expensive (30% NULL-album in 2004 vs 2–5% in recent years; NULL album forces LML's track-only live-resolution path).

## The three fixes

**1. Budget-header suppression** (`lml-fetch.ts`) — `budgetMs: null` unconditionally, BS#1914's lever, the same one BS#1978 applied to the live enrichment worker's CDC lane. No feature flag: unlike `enrichment-worker`, this job has no adjacent `'sweep'` lane to scope suppression away from — every candidate it looks up is exactly the population that needs the full cascade. This supersedes BS#2179 review HIGH 3, which concluded the job should keep the class-5 header on the reasoning that the "4s cutoff is wrong" finding applied only to the live CDC lane; that reasoning does not survive the measurement above. A cold release now runs LML's headerless cascade bounded by LML's own `LML_SEARCH_HARD_TIMEOUT_MS` (25s) instead of fast-degrading at ~4s.

**2. Newest-first tiebreak** (`query.ts`) — `ORDER BY no_match_recheck_attempted_at ASC NULLS FIRST, id DESC`. Never-attempted rows still lead and TTL-expired rows still sort oldest-attempted-first; only the `id` tiebreak flips. It stays a plain two-key `ORDER BY` rather than a per-tier `CASE` pair: the previously-attempted tier's `id` tiebreak only arbitrates rows sharing an identical `no_match_recheck_attempted_at`, which cannot happen when each stamp is its own single-row `UPDATE` evaluating `now()` in its own transaction — and a `CASE` ordering key can never be served by a B-tree, which would foreclose the index remedy the module docstring holds in reserve.

**3. OFFSET cursor** (`watermark.ts`, `cronjob_runs.cursor_position`, migration 0152) — a starvation guard independent of (1) and (2). Each run resolves its OFFSET from the stored cursor clamped into a fresh `countCandidates()` range, and advances it past however many of the run's candidates are **still candidates** afterwards. Advancing by `scanned` instead would double-count: a row that got a definitive answer has left the set, so stepping past it too skips rows the job has never read — a whole batch per run at a healthy resolution rate. In the all-transient regime the guard exists for, nothing departs and the advance is the full batch, so the starvation guarantee is identical to an outcome-independent rule. Fail-fast on a cursor/count read error, and skipped entirely under `DRY_RUN`.

## What is deliberately unchanged

- Transient responses still leave `no_match_recheck_attempted_at` untouched. Stamping on transient to force progress was considered and rejected — it reintroduces the BS#1977 / BS#2179 review HIGH 2 false-freeze, one TTL rotation removed. `orchestrate.ts`'s loop and contract are untouched.
- `cache_only` remains a definitive answer, not a transient (`UNANSWERED_DEGRADED_REASON.cache_only = false`).
- The trust-gate predicate `isTrustedLmlTrackContextMatch` is untouched; #2217 owns that and is not bundled here.
- The starvation guard stays job-local rather than being hoisted into shared infrastructure. Of the four siblings with the same attempt-marker + TTL shape, `rotation-release-id-backfill` and `concerts-artist-lml-resolver` load their whole candidate set with no `LIMIT` (no exposure at all), `flowsheet-metadata-backfill`'s W4 self-heal pass has the shape but a cohort its own docstring sizes below 100 rows against a 2,000 limit, and only `library-discogs-unavailable-recheck` is structurally identical — one genuine sibling is not a pattern.

## Known costs, recorded rather than elided

- **The cursor defers the head.** With newest-first ordering, a no-match row written today sorts to position 0, and once the cursor has moved off 0 it does not return until it wraps — a fixed `total / BATCH_SIZE` runs, ~5.7 months at current numbers. Accepted because the *first* pass starts at 0 and walks the cohort newest-first, so the backlog this ticket is about is rescued in the opening days; and because since BS#1978 the live worker already asks headerlessly, so a freshly-written no-match means a full-cascade lookup already missed — a much weaker recheck candidate than a backlog row that only ever failed under the ~4s clamp. `watermark.ts` documents the shape to reach for (reserve a slice of each batch for offset 0) if that trade stops holding.
- **Run duration** — see below.
- **Run duration widens** from ~10 min to ~83 min at LML's 25s hard cap, ~117 min if every call rides the 35s client-side abort, plus up to `LIVE_ACTIVITY_MAX_PAUSE_MS`. Still inside the 6h cadence and no new overlap with `rotation-release-id-backfill`, but `docs/ops-cron-scheduling.md` now says to budget ~2.5h when slotting the next cron.

## The index (migration 0153)

Flipping the tiebreak to `id DESC` made migration 0151's `(attempted_at NULLS FIRST, id ASC)` unable to serve the sort — a B-tree only supplies a mixed-direction order if it was built with those directions — so the planner read the whole cohort and top-N heapsorted it. Measured on prod, same predicate, `LIMIT 200 OFFSET 0`, A/B on the tiebreak alone:

| Ordering | Rows scanned | Execution time |
|---|---|---|
| `id ASC` | 200 (short-circuits at `LIMIT`) | 76 ms |
| `id DESC`, no new index | 137,278 (full scan + heapsort) | 41,326 ms |

~542x. `Buffers: shared hit=27246 read=95110`, `I/O Timings: read=39982 ms` — 97% heap I/O, ~743 MB per run — and unchanged warm (41,172 ms cold → 41,326 ms warm), so steady state rather than a cold cache. Against this job container's `DB_STATEMENT_TIMEOUT_MS=60000` that is ~19 s of headroom on a growing cohort; crossing 60 s would make the job do nothing at all, strictly worse than the stall this ticket fixes. Expected plan after 0153: back to a no-sort index scan short-circuiting at `LIMIT` (a no-sort scan, not index-only — the heap is still visited for the four projected columns the index doesn't cover).

**Replace, not accompany.** `flowsheet_no_match_recheck_idx` has exactly one consumer (`query.ts`, verified by grep — the only other references are its own migration, the schema declaration, and docstrings), and this PR removed the last query wanting its `id ASC` order. The partial predicate is byte-identical between old and new, so any predicate-only planner use is served just as well. Two near-identical partial B-trees on a live-written ~2.6M-row table is pure write amplification.

**`id DESC NULLS FIRST`, not `NULLS LAST`.** `drizzle-kit` emits `DESC NULLS LAST` by default, but `ORDER BY id DESC` means `DESC NULLS FIRST` in Postgres, and an index disagreeing on null ordering does not match the query's pathkeys — so the generated default would have produced an index that could not serve the query it exists for. `id` is the never-null primary key, so this is invisible to results; it decides only whether the index is usable at all.

**Why this is possible at all:** the `/simplify` pass that collapsed the `CASE`-per-tier tiebreak into a plain sort key. No index can serve an `ORDER BY <CASE expression>`; under the previous form there was no index remedy.

**`CONCURRENTLY`: the runner cannot support it, confirmed from source.** `dev_env/init-db.mjs` calls drizzle-orm's programmatic `migrate()`, and `drizzle-orm/pg-core/dialect`'s `migrate` runs every statement inside `session.transaction(async (tx) => …)`. Both `CREATE INDEX CONCURRENTLY` and `DROP INDEX CONCURRENTLY` are rejected inside a transaction block, and there is no escape hatch. Consistent with that, no migration in this repo executes `CONCURRENTLY` DDL — 20 files mention it, all in comments. So 0153 follows the established out-of-band pattern (0144, 0148, 0151, and 17 others): `IF NOT EXISTS` / `IF EXISTS` in-migration, and a runbook to run on prod **before** the deploy.

The runbook is create → verify → drop, in that order, so the cohort is never without a usable index: build `flowsheet_no_match_recheck_id_desc_idx` `CONCURRENTLY` while the old index still serves reads; confirm `indisvalid AND indisready` (a failed `CONCURRENTLY` build leaves an INVALID index behind, and dropping the old one on top of that would leave nothing usable); then `DROP INDEX CONCURRENTLY` the old one. Neither step takes an `AccessExclusiveLock`, so no INSERT pause and no wedged flowsheet mid-show. After that the migration is a no-op on prod, while fresh dev/CI databases — where `flowsheet` is empty — pick it up on first migrate. The in-file statement order is inverted from `drizzle-kit`'s DROP-then-CREATE emission for the same never-without-an-index reason; the end state is identical, so the snapshot is unaffected.

The create-then-revert validator warning (0151 creates, 0153 drops) is suppressed in 0153 with `-- @intentional-create-revert:` — the pair cannot be folded, since 0151 is deployed and hash-frozen, and 0151 carries no precondition guard, so the wedge that check protects against cannot occur. `validate-migrations.mjs` passes with 0 warnings.

## Migration 0152

0152 adds one nullable `integer` to the tiny `cronjob_runs` control table: DDL-only, additive, no rewrite, no backfill, no `CONCURRENTLY` dance. Backward-compatible with deployed code — every existing reader projects explicit columns and every writer omits the new one, so a NULL is legal — and `deploy-base.yml`'s `deploy` job `needs: migrate`, so the column exists before any code that reads it ships. 0151 is the latest on `main`; 0152 is unclaimed. The column is documented as opaque and job-private: a second job opting in may store entirely different semantics under its own `job_name` row. `last_run` stays the shared fleet-wide liveness field, and this job now stamps it on every completed run rather than letting `defaultNow()` freeze it at whichever run created the row.

## Verification

CI green on all 7 checks, including **Migration Dry-Run (prod-shaped data)**, which applies 0153. `npx jest --config jest.unit.config.ts` — 456 suites / 7934 tests pass, including new coverage for `countCandidates`, the `cursorOffset` parameter and its default, the `id DESC` tiebreak and the absence of `CASE` in the `ORDER BY`, `stillCandidates`' per-bucket classification, `nextCursorPosition`'s advance-and-non-advance cases, `wrapCursor`, the `last_run` heartbeat, and a pin that this caller sends `budgetMs: null`. `npm run typecheck`, `npm run lint` (0 errors), `npx prettier --check .`, and `node scripts/validate-migrations.mjs` all clean.


jakebromberg:
Filed three follow-ups from this PR's review, all deliberately out of scope here:

- WXYC/Backend-Service#2220 — `flowsheet-metadata-backfill`'s W4 self-heal pass re-asks the identical `enriched_no_match` cohort and still sends the class-5 budget header, so it inherits the exact ~4s clamp this PR removes. Needs a per-lane override, not a per-job one: the main drain's fast-degrade posture is still correct.
- WXYC/Backend-Service#2221 — `library-discogs-unavailable-recheck` has the same `NULLS FIRST` + `LIMIT` + unstamped-on-transient shape. Latent and unmeasured; filed on structural grounds with the settling query.
- WXYC/Backend-Service#2222 — the cursor's head-deferral (a fresh no-match waits up to a full wrap, ~5.7 months, for its first recheck). Only actionable once this PR merges.

Two notes for the reviewer of this PR:

1. **Migration 0153 needs a manual prod step before deploy.** The concurrent index build cannot run through drizzle's migration runner (it wraps every statement in a transaction), so the create → verify `indisvalid AND indisready` → drop runbook in the migration file must be run on prod first. After that the migration is a no-op on prod and only fresh dev/CI databases execute it.
2. **Merge order:** WXYC/Backend-Service#2217 should land before this. Both the live worker and this job read the same trust predicate, and if the drain is unblocked first it will re-reject exactly the rows #2217 makes recoverable — stamping them behind the 14-day TTL on the way past.

PR #2226, opened 2026-08-19
fix(lml-client): gate the track-context trust predicate on request<->result correspondence instead of search_type (BS#2217)

### LINKED ISSUE #2217: Track-context trust gate rejects LML row-less resolutions of the requested album (typed `alternative`, indistinguishable from a substitution)

jakebromberg:
Closes #2217

## What changed

`isTrustedLmlTrackContextMatch` (`shared/lml-client/src/trust.ts`) used to accept exactly two `search_type` values — `direct` and `compilation` — and reject everything else.

But `search_type` is a **telemetry label**. LML derives it purely from which search strategy ran last (`core/search.py`'s `_search_type_string`, whose own docstring reads *"Derive the search type string for telemetry from state"*). It makes no claim about whether the returned release corresponds to what was asked for. Reading it as a confidence signal conflates two structurally different results that both come back typed `alternative`:

- a **row-less non-library resolution** — `library_item.id === 0`, minted only by lanes that resolve from the request's *own* artist and song — returning the exactly-requested album with real Discogs artwork; and
- a genuine **same-artist substitution** — a different shelf album, carrying a real `library_item.id`.

The old gate discarded both. Since `enriched_no_match` is terminal, the discarded correct ones were frozen permanently.

This PR gates on request↔result **correspondence** instead, which is the question the caller actually wants answered:

- `direct` / `compilation` — trusted outright, unchanged.
- `song_as_artist` — rejected outright, unchanged. That lane matched the played *title* as an artist name, so any title correspondence is coincidence, not evidence.
- everything else (`alternative`, `fallback`, `none`, absent) — trusted only when `results[0].library_item.id === 0` **and** the returned title matches the requested album after casefolding, diacritic folding, and punctuation stripping.

`library_item.id === 0` is the load-bearing test, and it is a *structural* guarantee rather than a heuristic: LML has two producers of row-less items and neither reads the WXYC shelf, so a row-less item can never be a shelf album substituted in. The title check is the secondary guard. (Review corrected an earlier claim here that there was only one producer — see below.)

The requested album is threaded through every caller that has one: `apps/enrichment-worker/enrich.ts` (`extractArtwork`, via `finalizeRow`), `apps/enrichment-worker/streaming-reask.ts` (`reaskOne`), `jobs/flowsheet-no-match-recheck/lml-fetch.ts` (`extractTrustedArtwork`), and — added during review — the two Sentry classifiers in `empty-outcome.ts` via `handler.ts`. Omitting it remains valid and leaves the carve-out inert, so a caller with genuinely no request context still fails closed.

`isTrustedLmlAlbumMatch` is untouched; its function body is byte-identical.

## Live validation against production LML (2026-08-19)

Probed prod LML with the exact `(artist, album, track)` shape the recheck job sends:

| Row | `search_type` | Old gate | New gate | Why |
|---|---|---|---|---|
| Olivier Messiaen / `Diptyque Les Corps Glorieux` | `alternative` | REJECT | **ACCEPT** | returned `Diptyque, Les Corps Glorieux` — differs by one comma |
| Gescom / `MiniDisc` (×2 rows) | `alternative` | REJECT | **ACCEPT** | returned `Minidisc` — differs by casing |
| Vladislav Delay / `Entain` | `alternative` | REJECT | **REJECT** | `library_item.id = 64288`, a real library row (the `Vantaa` substitution) |
| Mei Semones / `Kurage` | `alternative` | REJECT | **REJECT** | `library_item.id = 72180` (the `Animaru / Kabutomushi` substitution) |

Both accepts differ from the request by only punctuation or casing. That is direct evidence for the deliberately-forgiving normalization the ticket decided on — a strict string comparison would have lost both of them. Both rejects carry a real library id and are rejected by the `id === 0` check alone, before the title is ever compared, which is exactly the property that keeps the general `alternative` rejection intact.

## Why this is urgent

A 7,010-release seed landed in the production Discogs cache on 2026-08-19 specifically so the `enriched_no_match` backlog can resolve without Discogs API calls. Five of five probed backlog rows come back `alternative` + `library_item.id === 0` — correct resolutions the old gate throws away. **This predicate is the sole remaining unlock for that seed.**

The seed is also time-boxed: `verify_cache.py --prune` is library-scoped, so the next monthly Discogs cache rebuild prunes the seeded rows. Landing this before that rebuild is what converts the seed into recovered playcuts.

## Safety: the general `alternative` rejection is not weakened

BS#1351's Yenbett→Tzenni defect and BS#1516/#1517's wrong-album rotation pollution are what this gate exists to prevent, and rows 4–5 above confirm that population is live today. Every substitution ever measured carries a real `library_item.id`, so all of them still fail the `id === 0` check. `enriched_no_match` is terminal — a too-permissive gate writes wrong artwork that nothing revisits — so the carve-out is bounded to request-derived, row-less releases and can never resurrect a shelf substitution.

`@wxyc/database`'s canonical `normalizeAlbumTitle` was considered and rejected (per the ticket): `@wxyc/lml-client` depends only on `@sentry/node` and `@wxyc/shared`, and that normalizer carries a dedup-key stability contract tied to a live cron. The local `looseTitleKey` here is a throwaway comparison key with no stability contract of its own.

`extractArtwork` is duplicated across four other jobs; only the enrichment-worker copy is in scope. The rest are BS#1959.

## Tests

New unit coverage in `tests/unit/shared/lml-client/trust.test.ts`, `tests/unit/apps/enrichment-worker/enrich.test.ts`, `tests/unit/apps/enrichment-worker/streaming-reask.test.ts`, and `tests/unit/jobs/flowsheet-no-match-recheck/lml-fetch.test.ts` covers every acceptance criterion on the ticket: row-less + matching title accepted, real-id substitution rejected, row-less + non-matching title rejected, `song_as_artist` + row-less + matching title rejected, and absent `requestedAlbum` behaving identically to pre-change. The existing BS#1359 gate tests pass unchanged.

Local gate, all green:

- `npm run test:unit` — 456 suites / 7980 tests passing
- `npm run typecheck` — exit 0
- `npm run lint` — 0 errors (pre-existing warnings only, none in the touched files)

## Verification left for a human

The ticket's last acceptance criterion needs prod credentials and is not agent-runnable: replay playcuts #5309468 (Agriculture / `The Spiritual Sound`) and #5309308 (Bim Sherman / `Ghetto Dub`) post-merge and confirm they now persist artwork instead of freezing at `enriched_no_match`.

---

## Post-review changes (not in the original ticket)

Review found two holes in the first implementation that the ticket did not anticipate. Both change behavior, so calling them out explicitly rather than letting them read as scope creep.

**1. The carve-out vouches for `results[0]`, but the artwork walk was scanning the whole array.** The gate validates `results[0].library_item` — that is the entirety of its evidence. The artwork extractor, however, returns the first result that *has* artwork. So a response whose `results[0]` was the row-less, title-matching item with no artwork and whose `results[1]` was a same-artist substitution carrying artwork would have persisted the *substitution's* cover onto a row accepted purely on `results[0]`'s evidence — the Yenbett/Vantaa class, smuggled in through the index. This was not reachable before this PR, because the old gate keyed on `search_type` and therefore vetted every entry uniformly; the carve-out is what broke that alignment.

Fixed by having the predicate report *why* a response is trusted (`lmlTrackContextTrust` → `'none' | 'search_type' | 'correspondence'`) and scoping the walk accordingly. `direct`/`compilation` still walk the whole array, preserving BS#961's per-entry compilation artwork; the correspondence path stops at index 0. The rule itself lives in `trust.ts` as `lmlTrackContextVouchedResults` so a future caller cannot reintroduce the hole by copying the loop without the guard.

**2. `flowsheet-no-match-recheck` compared an entity-encoded title against a decoded one.** The job sends LML `decodeHtmlEntities(candidate.album_title)` but was handing the gate the raw column. For a row whose `album_title` is `Rock &amp;amp; Roll`, LML is asked for `Rock &amp; Roll`, returns `Rock &amp; Roll` row-less, and the gate compared `rock amp roll` against `rock roll` → reject. Entity-bearing titles are exactly the population `decodeHtmlEntities` exists for, and since every pass stamps `no_match_recheck_attempted_at`, the rejection would have repeated behind the 14-day TTL indefinitely — the recovery loop entrenching the defect it was built to fix. Now decoded once and reused for both the request and the gate.

Three smaller review fixes:

- `isEmptyOutcome`/`classifyEmptyCause` share `extractArtwork` with `finalizeRow` but were called without the album, so once the carve-out could accept a row-less match with a null `artwork_url`, `finalizeRow` wrote `enriched_match` (a blank tile) while the classifier reported `lml_no_match`, which `SUPPRESSED_EMPTY_CAUSES` drops. That row is invisible to the BS#969 alert *and* to `flowsheet-no-match-recheck`, which only re-asks `enriched_no_match`. `handler.ts` now passes the same `candidate.album_title` the write path gated on.
- A response the gate *accepted* that simply carried no artwork was bucketed `trust_rejected` — the counter the team will read to judge whether this change worked. It now buckets on the gate's decision rather than on `search_type`.
- `looseTitleKey` now folds diacritics (NFD + combining-mark strip) before casefolding. The old doc comment claimed to absorb an accent difference and did not. WXYC's catalog is full of accented titles while flowsheet entries are hand-typed, so this is the same forgiving-comparison argument the ticket already made for casing and punctuation, and it stays bounded by `id === 0`.

### Declined

The review also proposed comparing `library_item.artist` against the requested artist. I did not, and the reasoning is recorded in a PR comment: it is the only *narrowing* on the table, it has no measurement behind it, artist strings are the field most likely to diverge harmlessly (Discogs ANV artifacts, classical/compilation credit variations — plausibly including the Messiaen row this PR recovers), and the one lane that motivated the concern already applies an 80/80 floor to the typed artist. Worth adding against an observed failure; not against a hypothetical, when the errors it would cause are terminal.

### Note for the reviewer

`isTrustedLmlTrackContextMatch` now has **no production callers** — both extractors moved to the verdict pair. It is retained (with a warning in its doc comment) because it is the name BS#1356/BS#1359 established across CLAUDE.md, the job READMEs, and several module docs, and removing it would mean a large prose churn in a time-boxed PR. Happy to delete it in a follow-up if you would rather not carry the export.

One follow-up worth filing, surfaced by the altitude review: this predicate's correctness rests on BS-side knowledge of LML *internals* (which functions mint `library_item.id === 0`). An explicit wire-contract field — something like `library_item.origin: 'library' | 'request_derived'` — would let the gate read a documented guarantee instead of a comment about LML function names. Cross-repo, so out of scope here.


jakebromberg:
## Code review round: 5 findings accepted, 1 half-finding declined

Recording the reasoning for the one thing I did **not** change, since it is a deliberate call rather than an oversight.

### Declined: adding an artist comparison to the correspondence check

The review's medium-severity finding on `shared/lml-client/src/trust.ts` came in two halves. I accepted the first and declined the second.

**Accepted half — the producer claim was factually wrong.** The doc comment asserted that `library_item.id === 0` is minted "exclusively from `_make_rowless_item`". LML's own source says otherwise: `lookup/rowless.py` notes that `_library_miss_discogs_search` (LML#583) "also surfaces `LibraryItem(id=0)` but builds it directly", which is precisely why the LML#681 counter is scoped to the flag-gated producers rather than to all `id=0` items. The comment now names both producers, states the property the gate actually relies on (neither reads the shelf, so neither can return a substitution), and records a second caveat the review caught: producer 2 builds its item as `title=best.album or album`, falling back to the caller's own typed album when the Discogs candidate carries no album string, which can make the title comparison tautological on that lane. That is exactly why `id === 0` — not the title — is described as load-bearing.

**Declined half — comparing artist as well as title.** The suggestion was to also compare `library_item.artist` against the requested artist, on the grounds that producer 2's 80/80 floor can score Hank Williams Sr. and Jr. identically. Three reasons not to do it here:

1. **It tightens the gate with no measurement behind it.** #2217 reasons explicitly about the error asymmetry: `enriched_no_match` is terminal, so a too-strict gate silently re-creates the very defect this PR fixes, and nothing revisits the rows it discards. Every widening in this PR is backed by a prod replay. An artist comparison would be the only narrowing, and it has none.

2. **Artist strings are the field most likely to diverge harmlessly.** LML#971/#972 documented that Discogs ANV `*` markers are a search-API display artifact, and credit variations ("Olivier Messiaen" vs "Olivier Messiaen / Jennifer Bate") are routine on exactly the classical and compilation releases in the recovered population. A loose artist equality would plausibly reject the Messiaen row this PR was written to recover — I have not measured that it wouldn't, which is the point.

3. **The lane already anchors on the typed artist.** `_library_miss_discogs_search` applies its 80/80 floor to artist *and* album against the typed values, so its output is "the Discogs release best matching the typed artist+album at ≥80/80" — correspondence, not substitution. BS#638 closed no-change on that floor after concluding the real lever is structural guards rather than threshold tuning, and `id === 0` is precisely such a structural guard.

If a wrong-artist row-less acceptance is ever observed in practice, adding the artist comparison is the obvious next step and it lands in one place. I would rather add it against a measurement than against a hypothetical, given which direction the errors are terminal.

### Accepted and fixed

| Finding | Fix |
|---|---|
| The artwork walk scanned every `results` entry while the gate had vetted only `results[0]` — a substitution at `results[1]` could donate its cover | `lmlTrackContextTrust()` now reports *why* a response is trusted; the walk covers the whole array only for `direct`/`compilation` (preserving BS#961) and stops at index 0 on the correspondence path |
| `flowsheet-no-match-recheck` sent LML the entity-decoded album but gated on the raw one, so entity-bearing titles self-rejected forever behind the TTL | Decode once, reuse for both the request and the gate |
| `isEmptyOutcome`/`classifyEmptyCause` were asked a different question than `finalizeRow`, so a blank-cover row-less match classified as `lml_no_match` (suppressed) while the write path recorded `enriched_match` | Threaded `candidate.album_title` through both from `handler.ts` |
| A response the gate *accepted* that carried no artwork was bucketed `trust_rejected`, inflating the counter used to judge this change | Bucket on the gate's decision rather than on `search_type` |
| `looseTitleKey`'s documented accent example did not actually hold | Fold diacritics via NFD before casefolding, and replace the invented example with the two real prod rows |

One extra cleanup while in there: the recheck job's test double for the trust predicate hand-mirrored the real rule and had to be updated in lockstep with every refinement. It now delegates to the real predicate, so the suite can no longer pass against a rule the job doesn't run.

PR #2270, opened 2026-08-25
One-shot drain for album_metadata rows whose artwork is an artist photo or a label logo (BS#2258)

### LINKED ISSUE #2258: Wrong non-null artwork_url (Discogs artist images / label logos) is uncorrectable: 6,977 label logos in album_metadata, 1,082 artist images in library
## Problem

Two tables carry `artwork_url` values that are **not album covers** — Discogs *artist images* and *label logos*, written by LML's `_resolve_fallback_artwork` cascade when it could not resolve a real cover. No code path in this repo can correct them. They are permanent.

Provenance is recoverable from the stored URL. A modern Discogs image URL carries a base64url blob **split across `/` path segments** and suffixed `.jpeg`; strip everything through the final `/w:NNN/` segment, drop the extension, remove the remaining `/` separators, then base64url-decode. The result is `s3://discogs-database-images/{R,L,A}-<id>-<ts>.jpeg`, where:

- `R-` — release cover (**correct**)
- `A-` — artist image (**wrong**: a photo of the band rendered as album art)
- `L-` — label logo (**wrong**: e.g. the Warp Records logo rendered as album art)

### Scale, measured against prod 2026-08-20

| table | rows with art | `R-` cover | `A-` artist | `L-` label | other |
|---|---|---|---|---|---|
| `wxyc_schema.library` | 48,585 | 46,300 | 1,082 | 629 | 574 (Apple `mzstatic`, legitimate) |
| `wxyc_schema.album_metadata` | 41,509 | 33,369 | 972 | **6,977 (16.8%)** | 191 |

**dj-site's card catalog reads `album_metadata`, the worse-affected table**, and the two tables disagree per-album. Autechre — *Chiastic Slide* is `R-30797` (a correct cover) in `library` but `L-23528` (the Warp logo) in `album_metadata`. Five Autechre rows carry a label logo there. This is user-visible today.

The `A-` cohort is **growing**. Splitting `album_metadata` by `updated_at` against the WXYC/library-metadata-lookup#687 deploy (`b4a0549`, prod 2026-07-29 10:06 PDT), which removed the label rung: `L-` before 6,976 / after **1**; `A-` before 243 / after **729**. #687 moved the failure from the label rung to the artist rung rather than fixing it. The near-zero `L-` count after the cutoff doubles as a negative control proving `updated_at` is a faithful proxy for artwork-write time — so those 729 artist-image writes in three weeks are real.

### `library.artwork_url` — structurally frozen

Exactly two writers exist, and **both hard-guard on `IS NULL`**:

- [`library.service.ts` L1605–L1613](https://github.com/WXYC/Backend-Service/blob/dcb2b7d9304591dcafcb00d08a79bb6856fde261/apps/backend/services/library.service.ts#L1605-L1613) — `updateArtworkUrl`, `.where(and(eq(library.id, id), isNull(library.artwork_url)))`. Added deliberately by #720 as a race-guard against the concurrent #637 backfill; correct for that purpose, but it also forecloses every later correction.
- [`jobs/library-artwork-url-backfill/enrich.ts` L94–L100](https://github.com/WXYC/Backend-Service/blob/dcb2b7d9304591dcafcb00d08a79bb6856fde261/jobs/library-artwork-url-backfill/enrich.ts#L94-L100) — `WHERE "id" = $1 AND "artwork_url" IS NULL`.

A wrong non-null value is invisible to both, forever.

### `album_metadata.artwork_url` — overwritable, but never selected

The enrichment-worker's conflict branch sets `artwork_url` from `...payload` **unconditionally**, guarded only by `setWhere: updated_at < NOW()` ([`enrich.ts` L341–L369](https://github.com/WXYC/Backend-Service/blob/dcb2b7d9304591dcafcb00d08a79bb6856fde261/apps/enrichment-worker/enrich.ts#L341-L369)). A re-enrichment therefore *would* heal the row.

Nothing ever selects these rows for one. Both existing drains narrow on the null predicate:

- #1209 (executed against prod as #1946) — `enriched_match` rows with `artwork_url IS NULL`.
- [`jobs/flowsheet-artwork-repair/repair.ts`](https://github.com/WXYC/Backend-Service/blob/dcb2b7d9304591dcafcb00d08a79bb6856fde261/jobs/flowsheet-artwork-repair/repair.ts#L1-L30) — both arms narrow on `artwork_url IS NULL`.

#1209 healed the cohort the same upstream LML bug left **null**. This ticket is its exact complement: the cohort that bug left **wrong and non-null**, which #1209's predicate could not reach by construction.

The only content filter anywhere in the write path is `filterSpacerGif`, which catches Discogs' 1×1 spacer placeholder and nothing else.

## Desired end state

A row whose `artwork_url` is a Discogs artist image or label logo is detectable and correctable, in both tables.

The two halves need different remedies:

1. **`album_metadata`** — a provenance selector plus a drain that re-enriches the selected albums. The existing write path already permits the overwrite; this is selection work, not writer work. Model it on #1209 / #1946, swapping the `IS NULL` predicate for the content predicate.
2. **`library`** — a deliberate correction path. Both writers refuse by construction, so this needs either a narrow exception to `updateArtworkUrl`'s guard (e.g. permit an overwrite when the *current* value decodes to `A-`/`L-`) or a one-shot targeted UPDATE. Prefer the former if the correction should stay available to the runtime; the latter if this is genuinely a one-time cleanup.

## Constraints

- **Sequencing is load-bearing: do not run the drain until WXYC/library-metadata-lookup#1241 lands.** WXYC/library-metadata-lookup#1237 (shipped as LML PR#1242, prod 2026-08-23) fixed only the *never-asked bound release* class — it re-asks Discogs for the release LML already bound when `artwork_checked_at IS NULL`. The covers for the rows in this ticket live under **different release ids** than the one LML bound: *Confield*'s cover is on release 1573110, not the bound 28138, which is not in the cache at all. Only the sibling-pressing rung (#1241) crosses pressings. Re-enriching before it lands re-derives the same wrong answer for that class.
- **A premature drain also destroys the measurement.** Re-enrichment stamps a fresh `updated_at`, and `updated_at` is the only available proxy for artwork-write time — it is what makes the before/after table above readable at all. Once overwritten, the `A-`-growth signal cannot be reconstructed.
- **Data safety.** Never overwrite a correct `R-` cover. The selector must be positive-match on `A-`/`L-`, not negative-match on "not `R-`" — the 574 + 191 `other` rows are legitimate Apple `mzstatic` URLs that decode to nothing. Run a `SELECT` with the exact predicate and eyeball the count before any `UPDATE`.
- The decode is pure string manipulation over an existing column — no Discogs API calls needed to *classify*. API calls are only needed to *re-resolve*.
- Rows that re-resolve to nothing must be left as-is rather than nulled: a wrong image is bad, but silently deleting artwork is a visible regression for a row that at least renders something. Decide this explicitly and record the decision.

## Acceptance criteria

- [ ] A provenance decoder lives in shared code (`@wxyc/metadata` alongside `filterSpacerGif` is the natural home) with unit tests covering: an `R-` cover, an `A-` artist image, an `L-` label logo, an Apple `mzstatic` URL, a legacy non-base64 Discogs URL, and a null.
- [ ] The prod counts in the table above are **re-measured** at implementation time and recorded on this ticket — they are as of 2026-08-20 and the `A-` cohort is growing.
- [ ] `album_metadata`: a drain selects `A-`/`L-` rows and re-enriches them; per-row outcomes are counted (`healed` / `still_wrong` / `no_match`) and reported.
- [ ] `library`: a correction path exists and is exercised against the 1,082 `A-` + 629 `L-` rows.
- [ ] Post-drain re-measurement shows the `A-`/`L-` counts materially reduced, with the residue explained (expected: albums with genuinely no cover anywhere in the cache).
- [ ] Autechre — *Chiastic Slide* renders `R-30797` in **both** tables, and the other four Autechre label-logo rows in `album_metadata` are resolved.
- [ ] If `updateArtworkUrl`'s guard is amended, the #720 race contract it protects is re-stated in the docstring and still holds.

## Notes for implementer

The decode, as SQL, for the selector — reassemble before decoding, or base64 rejects the `/` separators and the `.` extension:

```sql
convert_from(
  decode(
    translate(
      replace(
        regexp_replace(artwork_url, '^.*/w:[0-9]+/', ''),
      '/', ''),
    '-_', '+/') || repeat('=', (4 - length(...) % 4) % 4),
  'base64'),
'UTF8')
```

Verify the padding arithmetic against a known row before trusting a count — the blob length varies. A handful of rows will decode to garbage; treat a decode that does not match `^s3://discogs-database-images/[RLA]-` as unclassified rather than as wrong.

## Related

- WXYC/Backend-Service#1209 — drained the `artwork_url IS NULL` cohort from the same upstream LML bug (this ticket is its non-null complement)
- WXYC/Backend-Service#1946 — HITL prod execution of #1209's drain; the procedural model for this one
- WXYC/Backend-Service#720 — added the `updateArtworkUrl` race-guard that forecloses correction
- WXYC/Backend-Service#637 — the original `library.artwork_url` backfill
- WXYC/Backend-Service#1629 — `metadata_status` cannot distinguish an artist-only match from a real release match; its option 2/3 would let the worker label this class **at write time**, complementing this ticket's retroactive content selector
- WXYC/library-metadata-lookup#1241 — the sibling-pressing rung; **blocks the drain**
- WXYC/library-metadata-lookup#687 — removed the label rung on a false premise, moving failures to the artist rung
- WXYC/library-metadata-lookup#1237 — the narrower never-asked fix that shipped

jakebromberg:
Adds `jobs/artwork-provenance-remediation`, the one-shot drain for `wxyc_schema.album_metadata.artwork_url` rows whose stored image is provably a Discogs **artist image** or **label logo** rather than a release cover.

This is the exact complement of `flowsheet-artwork-repair` (#1209 / #1946). That drain healed the rows LML's `_resolve_fallback_artwork` bug left **null**; its `artwork_url IS NULL` predicate could not, by construction, see the rows the same bug left **wrong and non-null**. dj-site's card catalog reads this table, so five Autechre entries render the Warp Records logo today.

Closes #2258 for the `album_metadata` half. The `library` half stays open — see "What this does not do".

## How a wrong row is identified

Discogs' imgproxy URLs carry their origin S3 key as a base64url blob split across the path segments:

```
https://i.discogs.com/<sig>/rs:fit/g:sm/q:90/h:300/w:299/
  czM6Ly9kaXNjb2dz/LWRhdGFiYXNlLWlt/YWdlcy9MLTE4NjYt/MTIzMzE5MzU1Ny5q/cGVn.jpeg
  -> s3://discogs-database-images/L-1866-1233193557.jpeg
```

`R-` release cover, `A-` artist image, `L-` label logo. `classifyArtworkProvenance` lives in `@wxyc/metadata` alongside `filterSpacerGif`, and is tested against **verbatim prod URLs** of each class rather than synthesized ones — a hand-rolled fixture would only pin this decoder's own assumptions about what Discogs serves.

Two decisions worth a reviewer's attention:

- **The selector is a positive match on `A-`/`L-`, never a negative match on "not `R-`".** The 191 Apple `mzstatic` rows are legitimate covers that decode to nothing at all; a negative selector would sweep every one of them into an overwrite.
- **The SQL narrows coarsely (`LIKE '%i.discogs.com%'`) and the tested decoder decides.** A base64 decode written a second time in SQL — with its own padding arithmetic, no test, and no shared identity with the decoder that classifies LML's *replacement* — sitting on the far side of a 7,950-row overwrite is the wrong trade. One scan, auditable SQL, one tested rule.

## What it writes

Only `artwork_url` and `updated_at`, and only when the fresh answer is strictly better.

| outcome | condition | writes? |
|---|---|---|
| `healed` | LML resolved artwork that is not an `A-`/`L-` image | yes |
| `still_wrong` | LML resolved another `A-`/`L-` image | no |
| `no_match` | LML resolved no artwork (or only a `spacer.gif`) | no |
| `raced` | the guarded UPDATE matched zero rows | no |
| `error` | the LML call threw | no |

Four rules bound the blast radius. #2258 asked that the first two be decided explicitly and recorded:

1. **Never null a row out.** A wrong image is bad; a blank tile on a row that at least rendered something is a visible regression.
2. **Never write a lateral answer.** Swapping a Warp logo for a photo of Autechre fixes nothing and spends the row's `updated_at`, which #2258 relies on as its only proxy for artwork-write time.
3. **Narrow write.** The other nine columns came from the canonical writer under conditions this drain has not re-measured, and `discogs_url` in particular is already *correct* on these rows — LML bound the right release and then failed to find that release's cover.
4. **Exact-value race guard.** The WHERE pins the `artwork_url` the selector read, not just `album_id`, so a row a live enrichment healed between the scan and the write falls out of the predicate instead of being overwritten with a staler answer. Idempotent and order-independent in both directions.

## `budgetMs: null` is load-bearing

This is the only class-5 caller that suppresses `X-Caller-Budget-Ms`, and it is the one substantive departure from the donor job.

Class 5's default header caps LML's effective search budget at ~4s — right for an ordinary backfill, wrong for a population *defined by* LML having failed to resolve it once already. Answering these rows means a cold cross-pressing resolution measuring 4–20s on prod; under the cap LML returns `degraded: deadline_exceeded` with `artwork: null`, which this drain would score `no_match` and leave wrong. That is precisely the failure #1914 documented for the enrichment-worker. The pilot below was measured headerless, so the job reproduces those conditions rather than running against a variable the pilot never exercised. Pacing is unaffected — it comes from the shared `BACKFILL_LML_*` ceiling (1 concurrent, 20/min), not the header.

## Prod re-measurement, 2026-08-24

#2258's acceptance criteria ask for the counts to be re-measured at implementation time. These supersede the ticket's 08-20 table (read-only, from `wxyc-ec2`):

| | rows | `R-` | `A-` | `L-` | Apple |
|---|---|---|---|---|---|
| `album_metadata` | 41,524 | 33,383 | **973** | **6,977** | 191 |
| `library` | 50,761 | 48,431 | 1,121 | 629 | 580 |

Drain population: **7,950 rows**, ~6.6h at the default 20/min.

## New: 79% of the wrong rows already have a correct cover in `library`

Cross-checking the two tables on `album_metadata.album_id = library.id`:

| `album_metadata` | `library` | rows | |
|---|---|---|---|
| label | **release** | 5,764 | 72.5% |
| label | absent | 1,000 | 12.6% |
| artist | **release** | 517 | 6.5% |
| artist | absent | 227 | 2.9% |
| artist | artist | 204 | 2.6% |
| label | label | 172 | 2.2% |
| *(6 more, <25 each)* | | 66 | 0.8% |

**6,281 of 7,950 (79%) have a correct `R-` cover sitting in `library` for the same album.** The two tables were written by different writers at different times from different LML answers.

This is (a) strong independent corroboration that LML *can* resolve these covers — it demonstrably did — and (b) an unexploited optimization: a copy-from-`library` path would cut the drain from 7,950 LML calls to 1,669 (~1.4h instead of ~6.6h).

**Deliberately not implemented here.** It is a second heal source with different failure modes, and the pilot validated the LML path only; `library`'s `R-` proves the image is *a* release cover, not that it is *this album's*. Worth its own measurement and its own ticket rather than an unmeasured branch bolted onto a 7,950-row overwrite.

## Why running is the measured call, not a premature one

#2258 gates the drain on LML having a path to the covers. Of the two relevant LML fixes, only the first has shipped:

- **LML#1237** (PR#1242, prod 2026-08-23) re-asks Discogs for the bound release when `artwork_checked_at IS NULL`. A read-only pilot of 120 sampled `L-` rows against prod carrying this fix resolved **120/120** to real release covers — zero nulls, zero artist images, zero errors. Two independent 60-row samples (lowest-`album_id`, and md5-pseudo-random across all 6,977); 95% upper bound on the failure rate ≈ 2.5% by the rule of three. The random sample should have contained ~17% of rows from the harder `lib=absent`/`lib=label` classes in proportion, and still came back 60/60.
- **LML#1241**, the sibling-pressing rung, is merged **dark** behind `LML_RESOLVE_SIBLING_PRESSING_ARTWORK=false`. It covers rows whose cover lives under a *different* pressing.

The pilot says the dominant cohort heals without #1241, and a row that does not heal is left byte-identical — so a re-run after the flip picks it up at no cost.

## What this does not do

**`library`'s own 629 `L-` + 1,121 `A-` rows.** Both `library.artwork_url` writers hard-guard `IS NULL` (#720), so that half needs a deliberate correction path rather than a re-enrichment drain, and no pilot has measured it. #2258 stays open for it.

## Verification

- `npm run test:unit` — **8,245 passed** (26 new for the job, 24 for the decoder)
- `npm run typecheck`, `npm run format:check`, `npm run check:lml-callers`, `npm run check:docs` — all pass
- `eslint` on changed paths — 0 errors, 4 warnings, the same `security/detect-object-injection` baseline the donor job carries
- **Docker build not verified locally** — both this Dockerfile and the donor's fail identically at `npm install` with `E401 Unauthorized` fetching `@wxyc/shared` from GitHub Packages (expired local token). CI has a valid `NPM_TOKEN`.

Deploy registration needs nothing extra: `deploy-base.yml` reads `job-type: one-shot` from the workspace's own `package.json`, and turbo picks the new `jobs/*` workspace up as affected.

Closes #2258


jakebromberg:
## Update: self-review findings addressed, plus a CI heap bump

Ran the branch through an adversarial review pass. It could not construct a false positive in `classifyArtworkProvenance` (any garbage the split picks up is *prepended* to the decoded bytes, so the `^s3://discogs-database-images/` anchor fails — misses are possible, and every miss leaves the row untouched) and confirmed the race guard is sound (`album_id` is the PK, so the guarded UPDATE matches at most one row). It surfaced five findings; four are fixed in `03ca40cf`, one is recorded rather than changed.

**Fixed:**

1. **A rotated `LML_API_KEY` would have burned the whole run and exited 0.** `LmlAuthError` was caught generically and counted as `error`, so a rejected bearer would pace all 7,950 rows through at 20/min, emit 7,950 Sentry events carrying no aggregate signal, and report success — the silent-stall shape #1094 was filed against, on a job whose per-run cost is 40× the donor's. Now aborts on the first one with the bearer fingerprint, matching the handling already tested in `flowsheet-metadata-backfill`.
2. **A write error escaped `runRemediation` past the `finished` log.** The rows were never at risk (idempotent write, failed row re-selects); the run's accounting was — and those counts are what #2258's acceptance criteria ask for. Writes are now isolated per-row, and `finished` moved into a `finally`.
3. **`selectWrongProvenance` was applied at one call site, not at the boundary that writes.** `rows` is a public entry point that reached the writer unfiltered, so the property the design rests on — an Apple cover can never enter the drain — depended on `job.ts` remembering. Now enforced inside `runRemediation`; the provenance counters lost their catch-all `else` in the same move.
4. **Dry run now reports the `A-`/`L-` split** it already claimed to. A single total can't be reconciled against the ticket's per-class counts, which is the check that would catch a selector regression.

**Recorded, not fixed:** nothing verifies the fresh lookup binds the same release the row's `discogs_url` points at, so a row can carry a cover from release Y beside metadata describing release X. A guard on release-id agreement would reject exactly the rows LML#1241's sibling-pressing rung exists to heal — it returns a cover from a *different* pressing by construction. The rationale, what bounds the residue (rows are left byte-identical unless the answer is strictly better; 79% of the population already carries a correct `R-` cover in `library` from the same lookup path), and how to quantify the rest are in `remediate.ts` and the README.

## The CI heap bump in `4e04db3b` needs a reviewer's eye

This PR's Lint step OOMed twice at exit 134 while every other open PR passed. `projectService: true` builds one TypeScript program per tsconfig, so this job's workspace — the 51st under `jobs/*` — is what crosses 8 GiB. A cold-cache run on this branch peaks at a **measured 9.37 GiB RSS**.

Raised to 12288, which leaves ~2.6 GiB over that peak and ~4 GB for the OS on a 16 GB runner. The `//lint` note now carries the number rather than a hunch, and says plainly that there is about one more workspace of room and that the next trip needs the type-aware rules scoped or the lint job split, not another bump.

It is a repo-wide CI change riding in a feature PR, which I'd normally avoid — but this PR is what turns it red, and a PR has to leave CI green. Happy to split it out if you'd rather it land on its own.

**CI green** on `4e04db3b`: lint-and-typecheck, unit-tests (8,250), integration-tests, auth-tables-doc-drift.

jakebromberg:
## The wrong-album gap: measured at zero, and the guard I proposed would have been harmful

I said a release-identity check would reject the rows LML#1241's sibling-pressing rung exists to heal. **That was wrong**, and neither the review nor I checked it. In LML's [`_bind_resolved_release`](https://github.com/WXYC/library-metadata-lookup/blob/main/lookup/artwork.py#L207-L222):

```python
artwork_url = await _resolve_fallback_artwork(discogs_service, release.release_id, ...)
return DiscogsSearchResult(
    release_id=release.release_id,     # bound release X
    release_url=release.release_url,   # bound release X
    album=release.album_title,         # bound release X's title
    artwork_url=artwork_url,           # may come from sibling Y
)
```

Binding and artwork-sourcing are orthogonal — the sibling rung never moves `release_url`. And `to_match_result()` already puts the bound release's **title** on the wire, next to the artwork this job was reading and throwing the rest of away. The check I called expensive was free.

### Free pre-check: is release identity even a usable signal?

Decoded the stored cover's release id and compared it to the bound release across all 41,534 artwork rows. On healthy rows (`R-` cover, `discogs_url` present): **99.74% agree, 0.26% disagree**. Clean signal — disagreement is rare enough to be informative.

Two things fell out that I didn't know:

- **48% of the drain population has no `discogs_url` at all**, vs 19.6% of healthy rows. A row with no bound release is **2.4× more likely** to carry wrong artwork — which makes sense mechanically, since the cascade had nothing to work from. A release-id guard could only ever cover 52% of the population.
- The title check covers 100%, because `library.album_title` is in the enumeration's `WHERE`.

### Read-only probe, 240 rows, stratified

60 from each of `label`/`artist` × has-`discogs_url`/hasn't. The `artist`-without-`discogs_url` cell is a **census** — its pool is exactly 60. Run from `wxyc-ec2`; remote temp files cleaned up.

| | |
|---|---|
| Discogs release title matches the catalog exactly | **238 / 240** |
| **Wrong-album bindings** | **0** |
| Covers sourced from a sibling pressing | **0** |
| Healed to a real `R-` cover | 232 / 240 |

Per-stratum heal: `label+bound` 60/60, `label-nobound` 60/60, `artist+bound` 58/60, `artist-nobound` 54/60 → **99.6% population-weighted** (the strata deliberately over-weight the small artist cells, so the raw 96.7% understates it).

The two non-exact titles were a same-album format variant (`Pork Soda` vs `Pork Soda + 2 [10-inch single]`) and one row Discogs returned no title for. The 8 that didn't heal were left byte-identical, as designed — 4 resolved to another artist image, 4 to no artwork.

### The finding that changed the decision

**23 of the 119 sampled rows carrying a stored `discogs_url` bind a *different* release today than when they were written.** 22 of those 23 still match the catalog title **exactly**, and all 23 healed to a real cover — they are different pressings of one album, and today's binding is simply a better-covered one.

So the guard I was contemplating would have **skipped 19% of the rows that heal correctly**. It would have been worse than no guard.

### What changed in the code

No guard. The write stays narrow and ungated. Instead the job **counts** title agreement (`title_agreed` / `title_diverged`, each divergence logged with both titles) — the same question asked at 7,950 rows instead of 240, and the thing that catches a regression in LML's matching before the drain writes confident wrong covers. That asymmetry is the point: a Warp logo reads as "no art available"; a wrong cover asserts something false.

`titlesAgree` elides apostrophes where it maps every other punctuation mark to a space. A test caught that — an apostrophe sits *inside* a word, so spacing it splits `Amnesiac's` into `amnesiac s` and reports divergence against `Amnesiacs`. Catalog and Discogs disagree about apostrophes constantly; a counter that fires on that measures typography, not identity.

The false rationale is removed from the README and replaced with the measurement in both `remediate.ts` and the README. **Correction to my earlier comment above:** it never actually reached `remediate.ts` — that edit silently no-op'd on an em-dash mismatch I didn't verify at the time.

**CI green** on `7b7937e4`: lint-and-typecheck, unit-tests (8,264), integration-tests.

PR #2284, opened 2026-08-27
ci: allowlist BETTER_AUTH_TRUSTED_ORIGINS in set-ec2-env-var

### LINKED ISSUE #2283: Add orchestrator origin to prod BETTER_AUTH_TRUSTED_ORIGINS (+ allowlist the key in set-ec2-env-var.yml)
## Problem

[auto-dj-orchestrator#27](https://github.com/WXYC/auto-dj-orchestrator/issues/27) AC#4 requires the orchestrator's origin to be present in `BETTER_AUTH_TRUSTED_ORIGINS` in **staging and prod**. Staging is done. **Prod is not**, and two things block it — neither of which is the hosting decision that AC#4 was previously assumed to wait on.

[BS#1644](https://github.com/WXYC/Backend-Service/issues/1644) was closed on 2026-07-15 with its own trusted-origin acceptance criterion ("Orchestrator origin present in `BETTER_AUTH_TRUSTED_ORIGINS` (staging + prod)") never ticked. The service-account half of that ticket shipped; this half did not. This issue carries the remainder.

### Blocker 1 — the ops workflow cannot push this key

[`set-ec2-env-var.yml`](https://github.com/WXYC/Backend-Service/blob/main/.github/workflows/set-ec2-env-var.yml) resolves values through a hardcoded allowlist (`case "$SECRET_NAME"`), and hard-fails on anything not listed:

```
*) echo "::error::Unsupported secret_name '$SECRET_NAME'. Add it to the resolve step's env block." ; exit 1 ;;
```

`BETTER_AUTH_TRUSTED_ORIGINS` is not in it. Without that, changing this var on the live host means either a full image rebuild or a hand-edit of `~/.env` over SSH — exactly what this workflow exists to avoid.

### Blocker 2 — the upsert is a whole-value overwrite, and the current value is not readable from CI

The SSH step drops the existing line and appends a new one:

```bash
grep -v "^${KEY}=" "$ENV_FILE" > "$TMP" || true
printf '%s=%s\n' "$KEY" "$VALUE" >> "$TMP"
```

So the secret must hold the **complete** comma-separated list, not just the entry being added. The workflow deliberately never echoes values (`::add-mask::`, verify step prints only a length), so **the current prod value cannot be read from CI** — it has to be read off the host once and carried into the GH secret by hand.

Getting this wrong is a production outage, not a no-op: dropping `https://dj.wxyc.org` fails the origin check on every DJ sign-in.

## Current prod value (probed, not assumed)

An untrusted `Origin` is rejected with `403 INVALID_ORIGIN` *before* credentials are evaluated; a trusted one falls through to `401 INVALID_EMAIL_OR_PASSWORD`. Probing `POST https://api.wxyc.org/auth/sign-in/email` with a non-existent `@example.invalid` address therefore maps the trusted set without touching a real account:

| Origin | Result |
|---|---|
| `https://dj.wxyc.org` | **TRUSTED** (401) |
| `http://localhost:3000` | **TRUSTED** (401) |
| `https://feat-md-interface.dj-site.pages.dev` | **TRUSTED** (401) |
| `http://localhost:8090` (orchestrator, staging value) | untrusted (403) |
| `https://auto-dj.wxyc.org` (proposed prod value) | untrusted (403) |
| `https://wxyc.org` | untrusted (403) |
| `https://probe-abc123.wxyc-dj.pages.dev` | untrusted (403) |

Because `http://localhost:3000` and `https://dj.wxyc.org` are trusted *simultaneously*, `BETTER_AUTH_TRUSTED_ORIGINS` must be **explicitly set** in prod rather than falling through to `FRONTEND_SOURCE` — `FRONTEND_SOURCE` is read as a single origin by `oidc-login-page.ts`, `url-rewrite.ts`, and `provision-user.ts`, and `buildLoginPage` rejects a comma outright at startup.

The probe is consistent with a three-entry list, but **probing can only confirm origins you think to guess — it cannot prove the list has no fourth entry.** Read the real value off the host before writing:

```bash
grep '^BETTER_AUTH_TRUSTED_ORIGINS=' ~/.env
```

## The value to add, and why it is not the deploy host

`AUTH_TRUSTED_ORIGIN` is a **separate config var** from `ORCHESTRATOR_PUBLIC_URL` ([`src/config.ts:10` and `:30`](https://github.com/WXYC/auto-dj-orchestrator/blob/main/src/config.ts)). It is a header the orchestrator asserts on its own server-to-server calls ([`token-manager.ts:84`](https://github.com/WXYC/auto-dj-orchestrator/blob/main/src/backend/token-manager.ts)) — no browser enforces it, and nothing checks it against where the service actually runs. **It therefore does not have to equal the deployment URL, and should not.**

Decoupling is also the safer choice. Pointing it at a PaaS-generated hostname (`…up.railway.app`) would grant better-auth trusted-origin status — including as a redirect/`callbackURL` target — to a name on a **shared multi-tenant domain**. Delete or rename that service and the subdomain returns to the provider's pool, where whoever claims it next inherits the trust. A hostname under `wxyc.org`, which WXYC controls, has no such failure mode.

**Proposed:** `https://auto-dj.wxyc.org` — a name we own, stable across any hosting decision in [auto-dj-orchestrator#29](https://github.com/WXYC/auto-dj-orchestrator/issues/29), and the likely real public URL anyway (dj-site's virtual switch and the Arduino both need a stable endpoint). Confirm before pushing; the orchestrator's `AUTH_TRUSTED_ORIGIN` must be set to the identical string.

## Desired end state

- `set-ec2-env-var.yml` supports `key=BETTER_AUTH_TRUSTED_ORIGINS` with `restart_target=auth`.
- The prod GH secret `BETTER_AUTH_TRUSTED_ORIGINS` holds the current host value **plus** the orchestrator origin.
- The prod origin check accepts the orchestrator origin; `https://dj.wxyc.org` and the existing entries still pass.

## Where

- `.github/workflows/set-ec2-env-var.yml` — resolve-step env block + `case` arm
- [`shared/authentication/src/auth.definition.ts:151`](https://github.com/WXYC/Backend-Service/blob/main/shared/authentication/src/auth.definition.ts) — `BETTER_AUTH_TRUSTED_ORIGINS || FRONTEND_SOURCE || 'http://localhost:3000'`, unioned with `CORS_PREVIEW_ORIGINS`
- [`docs/env-vars.md`](https://github.com/WXYC/Backend-Service/blob/main/docs/env-vars.md) — existing prose for this var

## Constraints

- **Restart target is `auth`.** better-auth builds `trustedOrigins` once at auth-service startup. `apps/backend` resolves its own CORS from `FRONTEND_SOURCE` only (`apps/backend/middleware/cors.ts:71`), so the backend container does not need the restart.
- **Do not put the Pages preview wildcard in this var** — it belongs in `CORS_PREVIEW_ORIGINS`, which `auth.definition.ts` unions on top of whichever primary rung won.
- Least privilege: this list grants better-auth origin trust. Add exactly one entry; do not widen it to a wildcard under `wxyc.org`.

## Acceptance criteria

- [ ] `BETTER_AUTH_TRUSTED_ORIGINS` accepted by `set-ec2-env-var.yml` (allowlist + usage example + the two hazards documented inline)
- [ ] Current prod value read off the host and recorded before any write
- [ ] Prod GH secret holds the existing entries **plus** the orchestrator origin
- [ ] Workflow run completes with `restart_target=auth` and the auth healthcheck passes
- [ ] Post-push probe: orchestrator origin returns `401`, not `403 INVALID_ORIGIN`
- [ ] Regression probe: `https://dj.wxyc.org` still returns `401` (no entry dropped by the overwrite)

## Note — unrelated finding, not in scope here

`https://probe-abc123.wxyc-dj.pages.dev` is **untrusted** in prod, so the dj-site Cloudflare Pages preview wildcard is not in effect. `docs/env-vars.md` documents the production preview value as `https://*.wxyc-dj.pages.dev`, but the one preview host that *is* trusted is `feat-md-interface.dj-site.pages.dev` — a different Pages project domain (`dj-site.pages.dev`, not `wxyc-dj.pages.dev`), and an individually pinned host rather than a wildcard match. Either `CORS_PREVIEW_ORIGINS` is unset in prod or it carries a pattern that cannot match the real preview domain; either way preview deploys currently depend on someone hand-adding each branch host. Worth its own ticket.

## Related

- [auto-dj-orchestrator#27](https://github.com/WXYC/auto-dj-orchestrator/issues/27) (AC#4 — the consumer)
- [auto-dj-orchestrator#26](https://github.com/WXYC/auto-dj-orchestrator/issues/26) (deploy epic)
- [BS#1644](https://github.com/WXYC/Backend-Service/issues/1644) (closed; this is its unfinished trusted-origin AC)
- [BS#1107](https://github.com/WXYC/Backend-Service/issues/1107) (established the fail-closed CORS contract this var participates in)

jakebromberg:
Workflow-only change: adds `BETTER_AUTH_TRUSTED_ORIGINS` to the resolve step's allowlist in `set-ec2-env-var.yml`, plus a usage example.

## Why

The resolve step indexes secrets through a hardcoded `case` and hard-fails on anything not enumerated:

```
*) echo "::error::Unsupported secret_name '$SECRET_NAME'. Add it to the resolve step's env block." ; exit 1 ;;
```

`BETTER_AUTH_TRUSTED_ORIGINS` wasn't there, so there was no way to change the trusted-origin list on the live host short of a full image rebuild or hand-editing `~/.env` over SSH. [auto-dj-orchestrator#27](https://github.com/WXYC/auto-dj-orchestrator/issues/27) AC#4 needs the orchestrator origin added to it in prod.

It's a list of public origins, not a credential, but it rides the same secrets allowlist as `CREATE_AUTO_DJ_USER` / `AUTO_DJ_EMAIL` / `DONATE_*` for the reason already documented there — the resolve step can only index what it enumerates.

## The two hazards documented inline

Both are silent and both break production, so they're recorded at the declaration rather than left to whoever runs the workflow next.

**Whole-value overwrite.** The SSH step drops the existing line and appends a new one:

```bash
grep -v "^${KEY}=" "$ENV_FILE" > "$TMP" || true
printf '%s=%s\n' "$KEY" "$VALUE" >> "$TMP"
```

The secret must therefore hold the *complete* comma-separated list. A partial value doesn't error — it silently removes trust from whatever it omitted, and dropping `https://dj.wxyc.org` fails the origin check on every DJ sign-in. The workflow never echoes values (`::add-mask::`, verify prints only a length), so the current value has to be read off the host first: `grep '^BETTER_AUTH_TRUSTED_ORIGINS=' ~/.env`.

**First-time set shadows `FRONTEND_SOURCE`.** `auth.definition.ts:151` resolves `BETTER_AUTH_TRUSTED_ORIGINS || FRONTEND_SOURCE || 'http://localhost:3000'` — first rung wins outright, no union. In an environment where this var is currently unset, better-auth is trusting `FRONTEND_SOURCE`, and the value pushed here must be a superset or those origins lose trust. (Express-level CORS is unaffected: `apps/auth/app.ts:58` consults `FRONTEND_SOURCE` first and only falls back to this var.)

## `restart_target=auth`

better-auth builds `trustedOrigins` once at auth-service startup. `apps/backend` resolves its own CORS from `FRONTEND_SOURCE` only (`apps/backend/middleware/cors.ts:71`), so the backend container doesn't need the restart. The usage example encodes this.

## Verification

Prettier clean. No behaviour change for any existing key — the diff is one env entry, one `case` arm, one comment block, and one usage line; every existing arm is untouched.

The prod value this unblocks was mapped by probe rather than assumed, and the post-push check is symmetric (orchestrator origin must flip `403 INVALID_ORIGIN` → `401`, *and* `https://dj.wxyc.org` must still return `401`, proving the overwrite dropped nothing). Details and the full probe table are in #2283.

Closes #2283


jakebromberg:
Review pass applied in 6476b9e. All five findings verified against the code and fixed — the diff is mostly documentation, so that is where the defects were.

**Express-CORS parenthetical was false where it matters most** (medium). I had written "Express-level CORS is unaffected" unconditionally. `resolveCorsOrigin` skips an empty var name and *continues* rather than stopping (`cors-origin.ts:108-110`), so on a host with `FRONTEND_SOURCE` unset the auth service is fail-closed per BS#1107 — no `Access-Control-*` headers at all — and setting this variable flips it to serving credentialed ACAO for every entry in the list. That is exactly the deploy shape the second-position fallback was added for. Now stated conditionally, with a "check which shape the target host is in first."

**No normalization warning** (medium). `auth.definition.ts:154-157` only trims and drops empties, so `https://dj.wxyc.org/` survives as an entry that can never equal a real `Origin` header — and it fails silently: value looks set, container starts, healthcheck passes, sign-in breaks. `resolvePublicCorsOrigins` normalizes via `new URL().origin` and warns on each fixup; this path gets none of it. Added as a third ⚠️ block.

**`::add-mask::` claim was wrong for the multi-origin case** (low) — which is this variable's entire reason to exist. It registers the exact string handed to it, so for a comma-joined list the masked token is the whole list, not its members. Corrected, along with the trailing misattribution about the verify step (it reports a length for every key by design, masking or not).

**Restart caveat** (low). Added a pointer to the one already in `docs/env-vars.md`, with the reason it bites harder here: prod auth comes back with no Docker healthcheck, no explicit heap ceiling under the 450m cap, and Sentry events unattributed to a release.

**`docs/env-vars.md:52` was stale** (low). This repo records allowlist membership for every other key that has it — and line 157 goes out of its way to record *non*-membership — so silence there left the workflow as the only place to learn it. Now documented, including that `gh workflow run` executes `main`'s copy, so the allowlist entry must be merged before the key is accepted.

**The ambiguity you flagged is real and now resolved.** "Append to the current value" and "don't list the preview wildcard here" do conflict if one is already present. Resolved in favour of the overwrite rule: a wildcard already in the value is a pre-existing misplacement to move deliberately in its own change, never to drop as a side effect of this push. (It does not arise in prod today — `CORS_PREVIEW_ORIGINS` is unset there and the value holds no wildcard; see #2283.)

No functional change in this commit — the `env:` entry and `case` arm are untouched.

PR #2285, opened 2026-08-27
Auth role-model coherence: grant matrix as typed data with chain invariants; one admin predicate

### LINKED ISSUE #2282: Auth role-model coherence (2/2): grant matrix as typed data with chain invariants; explicit org-admin grants; one admin predicate
## Problem

`shared/authentication/src/auth.roles.ts` uses a two-layer model that reads as one layer, and the org documents a role hierarchy the middleware does not enforce. Concrete trap ([auth.roles.ts:49-54](https://github.com/WXYC/Backend-Service/blob/a3366583/shared/authentication/src/auth.roles.ts#L49-L54)): `stationManager` is built as `{ ...adminAc.statements, bin, catalog, flowsheet }`. The spread is better-auth's **fixed org-administration key set** (`organization`, `member`, `invitation`, `team`, `ac`), not a wildcard. Add a new statement key, grant it to `dj` and `musicDirector`, trust the spread for `stationManager` — and a plain DJ gets 200 while the station manager gets **403**, surfacing only when a real SM opens the page. Nothing existing catches it: `tests/unit/authentication/auth.roles.test.ts` is spot checks, not a matrix. `requirePermissions` ([auth.middleware.ts:193-224](https://github.com/WXYC/Backend-Service/blob/a3366583/shared/authentication/src/auth.middleware.ts#L193-L224)) is a pure per-role check with no hierarchy fallback — the documented "member < dj < musicDirector < stationManager" is folklore.

Additionally, the role-alias layer exists **six times** with three disagreeing policies:

1. `normalizeRole` + `systemRoleMap` (`auth.roles.ts:79-88`) — exact-match + `admin`/`owner`. Its `role in WXYCRoles` check **walks the prototype chain**: a JWT with role `'toString'` resolves and `requirePermissions` crashes calling `.authorize` on `Object.prototype.toString` (TypeError → 500).
2. shared's `roleToAuthorization` — case-fold + `station_manager`/`music-director`/`user`.
3. `device-authorization.ts:30` — `Object.hasOwn(WXYCRoles, ...)`, aliases **deliberately** excluded (a stray `owner`/`admin` member row must not QR-approve). Stays as-is.
4. `shared/authentication/src/auth.definition.ts:78` (**not** `apps/auth/`) — SQL `member.role IN ('admin', 'owner', 'stationManager')` inside `hasOtherAdminMembership`, the **revocation-side twin** of the admin-flag grant.
5. `apps/auth/provision-user.ts:56` — `ADMIN_SYNC_ROLES = ['stationManager', 'admin', 'owner']`, **two-thirds dead**: its own validation at `:74` rejects `admin`/`owner` first. That validation is itself `role in WXYCRoles` — the same prototype-chain flaw, so `provisionUser({ role: 'toString' })` passes validation today.
6. `scripts/backfill-missing-org-members.ts:47` — the same set as a `Set`, plus its own `mapUserRoleToMemberRole` and a comment citing `systemRoleMap`.

## Desired end state

Backend-Service owns the **only** grant matrix, as plain data: totality (every station-domain key decided for every role) enforced **by the type system**; monotonicity along shared's `ROLES` chain enforced by a unit test; the spread replaced by an explicit org-admin block pinned against the library; the alias layer collapsed onto shared's `canonicalizeRole` with an exhaustive BS-side pin; the admin-granting predicate defined exactly once. Runtime stays flat — the hierarchy becomes a checked invariant on the data, never a runtime fallback (preserving the `flowsheet: ['manage']` anti-role-proxying rationale). No endpoint's 200/403 outcome changes except two named, deliberate deltas (alias widening; prototype-key 500 → 403 bug fix).

## Steps (TDD order is load-bearing: pin first, refactor second)

0. **Prerequisite: bump `@wxyc/shared` to `^5.0.0`** in the four manifests that pin it (`apps/auth`, `apps/backend`, `shared/authentication`, `shared/lml-client`) + `npm install`. Steps 6-7 import `ROLE_ALIASES` from the installed dist (`jest.unit.config.ts` resolves `@wxyc/shared` via workspace `node_modules`), so no red-green cycle is executable before this.
1. **Pin the current construction — all four roles — and know which layer each pin guards.** The unit suite does **not** run real better-auth: `jest.unit.config.ts:47-48` maps `better-auth/plugins/access` and `better-auth/plugins/organization/access` to hand-written mocks whose `adminAc.statements` is a hardcoded literal, so a unit-level deep-equal against `adminAc.statements` compares the mock to itself. Two mechanisms: **(a)** the unit pin — each role's `.statements` deep-equals its current value (`member`/`dj`/`musicDirector` = exactly their three custom keys; `stationManager` = `adminAc.statements` + the three custom keys) plus `authorize()` spot checks — guards the *data and the step-3 refactor* under mock semantics. **(b)** a **mock-sync check** guards *library drift* — and because `scripts/` is the one directory this repo neither lints (`eslint.config.mjs:35` global-ignores `scripts/**`) nor typechecks, structure it as: a pure exported comparator `compareAccessModules(real, mock)` (deep-equals `defaultStatements` + `adminAc.statements`; runs the full (role, key, action) matrix through both modules' `createAccessControl`/`newRole().authorize` asserting `.success` agreement) unit-tested in `tests/unit/scripts/` (precedent: `ci-env-surface-parity.test.ts`); plus `scripts/check-better-auth-mock-sync.ts` as a thin `tsx` runner that resolves real better-auth via `createRequire` rooted at `shared/authentication/` (better-auth is not a root dep — a bare import from `scripts/` works only by undeclared hoisting) and **exits non-zero on any import/resolution failure**. Wire as npm alias `check:better-auth-mock-sync`, its own hard-fail line in `.husky/pre-push` (one command per line, no `&&`), and the CI lint job.
2. **Restructure `auth.roles.ts` as a two-block matrix.** `keyof AccessControlStatement` is 8 keys (the statement spreads `defaultStatements`), so naive full totality forces 20 boilerplate `[]` cells. Instead: **Block 1 `WXYC_GRANTS`** — typed-total over the WXYC key domain only, keyed on **shared's `WXYCRole`** (keying on `ImplementedRole` is circular — it derives from `WXYCRoles`, which is now constructed *from* the matrix; TS degrades that to `any` and silently destroys the guarantee): `{ [R in WXYCRole]: { [K in Exclude<keyof AccessControlStatement, keyof typeof defaultStatements>]: readonly ... } }`, with `export type ImplementedRole = keyof typeof WXYC_GRANTS` re-derived. The `_AssertLocalRolesAreShared` assertion at `:71` becomes redundant and goes, with the stale `:68-70` comment ("shared includes `admin`" — false). Missing a key on any role = compile error: adding `reviews: ['read']` without deciding all four roles cannot build. **Block 2 `ORG_ADMIN_GRANTS`** — the explicit better-auth org-administration set, `stationManager` only, pinned deep-equal against `adminAc.statements` by step 1. Construct roles as `newRole({...stripEmpty(WXYC_GRANTS[role]), ...(role === 'stationManager' ? ORG_ADMIN_GRANTS : {})})` — the empty-strip keeps `.statements` byte-identical (better-auth treats `[]` and absent the same for `.success` but not for `role.statements` shape/error strings; no `[]` cells exist today — the strip is for future denials).
3. **Remove the spread**, keeping the step-1 pin green. Preserve the `flowsheet: manage` and not-mirrored-to-shared rationale comments; add one explaining the explicit list and what the pin guards.
4. **Monotonicity test**: chain from shared's `ROLES`; per station-domain key, `WXYC_GRANTS[lower] ⊆ WXYC_GRANTS[higher]` for each adjacent pair. Kills the DJ-200/SM-403 inversion class structurally.
5. **Runtime parity test**: every (role, key, action) across both blocks — `authorize` succeeds iff the data grants it; `[]`-decided keys must fail. (Runs under the mock; the mock-sync check certifies the mock's `.success` semantics.)
6. **Collapse the alias layer with an exhaustive pin.** `normalizeRole` delegates to shared's `canonicalizeRole` (name + `ImplementedRole` return type kept); `systemRoleMap` deleted. Two deliberate behavior deltas, both pinned: **widening** (accepts case variants + `station_manager`/`music_director`/`music-director` — reaches `admin-flag-sync.ts:29`, i.e. the global `auth_user.role='admin'` grant; unreachable in stored data once step 7 fixes `provision-user`'s validation) and **narrowing** (prototype keys: `'toString'`/`'constructor'` now 403 instead of 500). In `tests/unit/authentication/shared-type-compatibility.test.ts` (the existing cross-repo witness): deep-equal shared's `ROLE_ALIASES` against a BS-local pinned copy of the complete map; a **fold guard** asserted against `normalizeRole` — accepts `'STATION_MANAGER'`, `' admin '`; rejects `'station-manager'`, `'station manager'`, `'user'`, `'administrator'`, `'toString'`, `'constructor'`, `'__proto__'` — so a shared-side change to table *or* pre-lookup fold turns BS CI red on the lockfile bump. Plus `admin-flag-sync` tests for `'admin'` and one variant string.
7. **Unify the admin-granting predicate.** Export `grantsAdminFlag` from `admin-flag-sync.ts` **and from the `@wxyc/authentication` barrel** (`shared/authentication/src/index.ts` doesn't export that module today). (a) `hasOtherAdminMembership` drops the SQL IN-list **and the `.limit(1)`** (after the filter moves to TS, the limit returns one arbitrary row — a user with both a `dj` and `stationManager` row would wrongly lose the flag); TS-filters all rows via `grantsAdminFlag`; symmetry test includes the multi-membership case. (b) `provision-user.ts` replaces `ADMIN_SYNC_ROLES` with `grantsAdminFlag`; **narrow `:74` to `Object.hasOwn(WXYCRoles, role)`** (the `in` check passes `'toString'` today); `tests/unit/auth/provision-user.test.ts:76-89`'s hand-built `jest.mock('@wxyc/authentication', ...)` factory gains `grantsAdminFlag` (named exception to "pre-existing tests unmodified"). (c) `backfill-missing-org-members.ts` replaces its `Set` with `grantsAdminFlag`; fix its `systemRoleMap` comment. Grant/revoke symmetry test: every pinned accepted string and near-miss agrees between `grantsAdminFlag` and the `hasOtherAdminMembership` filter.
8. **Docs**: CLAUDE.md's "Roles (hierarchical): ..." line; `docs/authentication.md` — the permissions matrix table, the numbered `requirePermissions` flow (step 4 "Validate role exists in `WXYCRoles`" becomes false), and the "role mismatch gotcha" section (now carries the alias story + emitted-vs-accepted distinction). Cite/fold the `docs/auth-convergence-proposals` branch (2026-02-14, unmerged): this initiative operationalizes its proposal 1 (admin as separate axis — since implemented by `admin-flag-sync`) and keeps the `admin`/`owner` alias as read-side normalization only.

## Verification traps (do not re-learn)

- New tests are **unit tests** — `npm run test:unit` + CI's unit-tests job are the gate. **`npm run ci:testmock` is structurally blind here**: integration-only `testMatch`, and integration runs `AUTH_BYPASS=true`, under which `requirePermissions` returns before the permission block.
- **Do not touch `tests/mocks/authentication.mock.ts`** — these tests import `auth.roles.ts` directly. The better-auth mocks are also unmodified; the unit suite runs *under* them, which is exactly why the mock-sync check exists.
- `auth.definition.ts` passes `roles: WXYCRoles` to `organizationPlugin` with no separate `ac` — better-auth's own org operations resolve against these role objects; byte-identical `.statements` (step-1 pin + empty-strip) is what keeps their behavior, including error strings, unchanged.

## Constraints

- **Blocked by** WXYC/wxyc-shared#409 (needs `ROLE_ALIASES`/`canonicalizeRole` from published 5.0.0).
- The dj-reviews internal surface (approved plan, unimplemented) must start **after** this merges — its `reviews: ['read']` is the first key added under typed totality.
- No migrations; auto-deploy on merge. Run typecheck + lint + test:unit locally before pushing.

## Acceptance criteria

- [ ] No `...adminAc.statements` spread; `WXYC_GRANTS` typed-total (keyed on shared `WXYCRole`), `ORG_ADMIN_GRANTS` explicit SM-only; empty-strip adapter; all four roles' `.statements` byte-identical to pre-refactor
- [ ] Monotonicity, runtime-parity, adminAc-pin, exhaustive `ROLE_ALIASES` deep-equal + fold-guard (accept/reject/prototype cases), and admin-flag-sync tests green in the unit suite
- [ ] Mock-sync check: comparator unit-tested in `tests/unit/scripts/`, `tsx` runner hard-failing on resolution errors, `check:better-auth-mock-sync` alias, own pre-push line, in CI lint job
- [ ] `normalizeRole` delegates to shared; `systemRoleMap` gone; `grantsAdminFlag` exported from the barrel and consumed by all three copies (SQL IN-list + `.limit(1)` removed; `ADMIN_SYNC_ROLES` deleted; provision-user `:74` narrowed to `Object.hasOwn`; backfill Set deleted + comment fixed); grant/revoke symmetry incl. multi-membership case green
- [ ] `device-authorization.ts` untouched except a comment naming its divergence deliberate
- [ ] Pre-existing auth unit tests pass unmodified except the two named carve-outs (deleted-export tests; provision-user mock factory)
- [ ] `docs/authentication.md` (matrix, flow step 4, gotcha section) + CLAUDE.md updated

## Related

- WXYC/wxyc-shared#409 — companion issue (blocker): role identity + alias table + matrix deletion, shared 5.0.0
- WXYC/wxyc-shared#254 — Ticket Giveaways epic: its future `giveaways`/`venues` grants land in `WXYC_GRANTS` here, not in shared
- WXYC/archive#89 — archive's @wxyc/shared pin (fan-out context)

jakebromberg:
Closes #2282. Companion to WXYC/wxyc-shared#409 (PR #410, shipped as @wxyc/shared 5.0.0).

## What

**The grant matrix becomes data.** `auth.roles.ts` now holds two blocks instead of four hand-built `newRole` calls:
- `WXYC_GRANTS` — station domain, **typed-total**: every role must decide every key (`[]` = explicit denial), so adding a statement key without deciding it for all four roles is a **compile error**. Keyed on shared's `WXYCRole` (keying on `ImplementedRole` would be circular now that it derives from the matrix — TS would degrade it to `any` and silently drop the guarantee); `ImplementedRole` is re-derived from `WXYC_GRANTS`.
- `ORG_ADMIN_GRANTS` — better-auth's org-administration keys, stationManager only, **written out explicitly**. This replaces `...adminAc.statements`, which reads as "stationManager gets everything" and is in fact a fixed library-owned set conferring no custom key. That's the trap: grant a new key to `dj` and trust the spread for `stationManager`, and a plain DJ gets 200 while the station manager gets 403.

**The chain becomes a checked invariant, not folklore.** A monotonicity test walks `@wxyc/shared`'s `ROLES` and asserts each role's grants ⊇ the role below it. `requirePermissions` stays a pure per-role check with no fallback — that's what keeps `flowsheet: ['manage']` a real action rather than a proxy for "MD or above". Roles are built through an empty-strip adapter so `.statements` stay byte-identical (better-auth distinguishes absent from `[]` in `role.statements` and error strings, though not in `.success`); pinned for all four roles *before* the refactor and green across it.

**One alias table, one admin predicate.** `normalizeRole` delegates to shared's `canonicalizeRole`. Two deliberate deltas, both pinned: it accepts more strings (case variants, `station_manager`), and it **stops** accepting prototype keys — `role in WXYCRoles` walked the prototype chain, so a JWT with role `'toString'` resolved and crashed `requirePermissions` with a **500** instead of a 403 (`provision-user.ts:74` had the same flaw; narrowed to `Object.hasOwn`). Because that widening reaches the global admin flag, `grantsAdminFlag` is exported as the single definition and replaces verbatim copies in `hasOtherAdminMembership`, `provision-user.ts`, and `backfill-missing-org-members.ts`. **`hasOtherAdminMembership` drops its `.limit(1)`** with the SQL role filter — with the predicate in TS, a limit returns one arbitrary membership and a user holding both a `dj` and a `stationManager` row would wrongly lose the flag.

**A tripwire that actually trips.** The unit suite runs against hand-written better-auth mocks (`jest.unit.config.ts` maps the real ESM), so its role pins compare the mock to itself and can never see a library upgrade. `check:better-auth-mock-sync` closes that: a pure comparator in `shared/authentication` (linted, typechecked, unit-tested — `scripts/` is none of those) plus a thin `tsx` runner resolving better-auth via `createRequire` rooted at the workspace that declares it, hard-failing on any resolution error. Wired into pre-push + CI. Verified by deliberately perturbing the mock: it fails with a readable diff, then passes on restore.

## Behavior

No unintended HTTP-layer change; all pre-existing auth tests pass unmodified except two named carve-outs (deleted-export tests in shared; `provision-user.test.ts`'s hand-built mock factory gaining `grantsAdminFlag`). Accepted semantic deltas: the alias widening, its deliberate mirror on the revocation path, and the prototype-key 500 → 403 bug fix.

## Verification

`typecheck` + `lint` (0 errors) + `test:unit` (**8474 passed / 481 suites**) + `format:check` + `check:auth-tables-doc` + `check:better-auth-mock-sync` all green locally. Note `ci:testmock` is **structurally blind** here — integration-only `testMatch`, and integration runs `AUTH_BYPASS=true`, where `requirePermissions` returns before the permission block; the unit suite is the real gate.


jakebromberg:
Addressed all six `/code-review` findings. Each was verified empirically before fixing — three turned out to be genuine holes in the guarantees this PR claims, so the fixes matter more than the diff size suggests.

**1. A fifth copy of the admin-granting set, on a live grant path** (`apps/auth/app.ts:567`). The boot reconciler's `findUsersMissingAdminFlag` still filtered `member.role IN (...)` in SQL. That path exists *specifically* to retry a membership hook that failed — so a narrow copy there meant a membership the widened hook grants on could never be repaired if the hook died. Routed through `grantsAdminFlag`; `docs/authentication.md`'s "four prior copies" corrected to five.

**2. The cast silently disabled action validity for the whole matrix.** `as Parameters<typeof newRole>[0]` instantiated better-auth's generic at its index-signature constraint. **Verified:** `flowsheet: ['read', 'write', 'mange']` compiled clean. Monotonicity structurally cannot catch this — it only notices a typo when a *lower* role spells the action correctly, so a typo on stationManager alone (or down a whole column) ships as a 403. `StationGrants`/`OrgAdminGrants` now derive permitted actions from `AccessControlStatement`; cast removed. **Re-verified:** same typo now fails with `Type '"mange"' is not assignable... Did you mean '"manage"'?`

**3. The monotonicity invariant skipped exactly the case it exists for.** `allStatementKeys` was a literal list, so a newly added key was exempt at the one moment the check matters. **Verified** by injecting the real inversion (`roster` granted to `dj`, denied to `stationManager`): monotonicity passed **green**; only the pin test failed — and the pin test is precisely what you'd update when adding a legitimate key, so the guard meant to survive that update didn't fire. Keys and the parity action universe now derive from `statement` via exported `STATEMENT_KEYS`/`STATION_KEYS`/`STATEMENT_ACTIONS`. **Re-verified:** monotonicity now fails on that injection on its own.

**4. The parity comparator reported zero findings while a real `success`-level divergence existed.** It never sent an empty request or empty action list. **Measured against the installed library:** `authorize({})` and `authorize({resource: []})` are both `false` in better-auth and were both `true` in the mock. That's the shape the comparator most needs to agree on, since `[]` is how the matrix spells an explicit denial. Mock corrected, comparator extended. **Re-verified:** it now surfaces all **36** divergences it previously missed.

**5. The provision-user mock restated `grantsAdminFlag`** in the file whose subject is that restatements drift — and the restatement was already wrong (lowercased the input, then tested a list containing camelCase `stationManager`: a branch that could never match). Now wires in the real predicate via `requireActual`.

**6. `selectMemberRole` — filed as #2286, named in place, deliberately not fixed here.** It mints a non-deterministic, org-unscoped station role into the JWT: the same multi-membership hazard this PR removed from `hasOtherAdminMembership`. Fixing it changes which role real users receive in their token, so it deserves its own PR, review, and deploy to bisect against. A `KNOWN GAP (BS#2286)` comment marks it so it isn't silently inherited.

All checks green on `df15dacc`: typecheck, lint (0 errors), **8474 unit tests / 481 suites**, format, auth-tables-doc, mock-sync, and Integration-Tests.

jakebromberg:
## Review follow-ups

Four review rounds since `df15dacc`, four commits. Everything correctness-affecting was fixed on the branch; three findings were out of this PR's scope and are filed rather than carried:

- **#2286** — `selectMemberRole` mints a non-deterministic, org-unscoped station role into the JWT. Marked `KNOWN GAP` in place; fixing it changes which role real users receive in their token.
- **#2287** — `syncAdminRoles` never revokes, so a failed *revocation* hook strands `auth_user.role='admin'` forever. The boot sweep's comment now states that scope honestly instead of claiming to retry any failed hook.
- **#2288** — `grantsAdminFlag` / `normalizeRole` fail closed on better-auth's comma-joined multi-role `member.role` (`"dj,stationManager"` is a storable value that 403s the user on every request). Pre-existing and it fails in the safe direction, but this PR is what makes the predicate canonical, so it is where the gap became worth naming. Needs a prod row count before a fix is chosen.

### Fixed on the branch

- **`624c4e0c`** — the mock-parity comparator only sent single-resource, array-form requests at the default connector, so it could not see that the mock ignored `authorize`'s second argument and threw on the `{ actions, connector }` object form. `dj.authorize({flowsheet:['read','write']}, 'OR')` was `success` in the library and failure in the mock — a unit test could assert 403 while production returned 200, behind a green check. Mock rewritten as a structural copy of `role().authorize`; comparator widened to the object form, multi-resource shapes, and `undefined`/`AND`/`OR`/`or`.
- **`91e0f848`** — constructed roles shared their action arrays with the matrix, and both are exported. Measured: one `WXYC_GRANTS.dj.flowsheet.push('manage')` flipped `dj.authorize({flowsheet:['manage']})` to true in-process, granting every DJ the operator tier behind `POST /flowsheet/shows/:id/force-end`. `buildRole` now clones, pinned by identity assertions.
- **`cd91ce47`** — the "adding a statement key is a compile error" guarantee had a hole: `StationKey` subtracted `defaultStatements`, so a station key *named* like a library key (`member`, for a roster gate) was subtracted back out and skipped by all three invariant suites while `ORG_ADMIN_GRANTS.member` silently supplied better-auth's actions under that name. The station domain is now its own block with two compile-time assertions closing both routes in.
- **`d596af2e`** — `STATEMENT_ACTIONS` held better-auth's own array instances, so an importer could mutate the library's module state process-wide; the arrays are now copied before use.

Also corrected several comments that the reviews disproved: `satisfies` (not the rebuilt construction) is what enforces action validity; `organizationLimit` does not bound `hasOtherAdminMembership`'s query; `@wxyc/shared` 5.0.0 has no `RESOURCES`/`ROLE_PERMISSIONS` to mirror to; and the predicate previously existed in five places, not four.

Full CI green on every pushed commit. Final: 481 suites / 8482 unit tests.

PR #2294, opened 2026-08-27
docs+lint(privacy): PII field registry, column comments, dead Cognito spec removal, restricted-real-name rule

### LINKED ISSUE #2293: PII conflation safeguards: field registry, column comments, restricted-real-name lint, dead Cognito spec
## Problem

Since `a0cd1979` (2025-12-31) collapsed the old `djs` table into `auth_user`, better-auth's required, innocuously-named `auth_user.name` column has been a hidden second copy of DJs' legal names — dj-site provisioning fills it with `realName || username`, so every ordinary `.name` read on a DJ row is a latent PII leak to a public surface (BS#1286, BS#1288, `2a37bbc6`, BS#1393, BS#2281 all trace back to that one structural fact). The doctrine that would have prevented each recurrence existed only as prose in scattered docblocks — several of which had already gone stale and were themselves propagating the wrong mental model.

## What this adds

Track 1 (labels and warnings) + Track 3a (ESLint rule) of the DJ real-name PII safeguards program. No behavior change: documentation, a DDL-only migration, comment rewrites, dead spec removal, and a new lint rule.

- **`docs/pii.md` — the PII field registry.** One table classifying every column in the DJ-name family: `auth_user.real_name`, `auth_user.email`, and tubafrenzy `DJ_NAME` as PII; `auth_user.name`, `auth_user.dj_name`, `flowsheet.dj_name`, `shows.legacy_dj_name`, `shows.dj_name_override` as explicitly non-PII, so the confusion has a written answer instead of being re-litigated per incident. Each row carries contents, why the column exists, allowed read sites, and its enforcement mechanism (or an honest "none today"). Includes a short DJ-name-vs-real-name history section and an `@rule`-annotated Enforcement section. Registered as one bullet in CLAUDE.md's Topic guides router.
- **The doctrine lives in the database too.** A DDL-only `COMMENT ON COLUMN` migration puts the warning on the five confusable columns, so a reader who reaches the schema through `psql`/`\d+` or a GUI — never opening the repo — still sees which column is the legal name and which is the public handle.
- **`wxyc/restricted-real-name` ESLint rule.** Flags `realName`/`real_name` as both member-expression access and object-literal property keys, outside an explicit allow-list of the legitimate legal-name sites. A new PII read now requires an allow-list edit in the same PR — it converts "someone remembered the doctrine" into "CI failed."
- **Docblock sweep across 7 files.** The mirror-forwarding doctrine comments now describe the end state (`real_name` is the sole legal-name carrier; `auth_user.name` is a derived display handle/username, backfill pending) and each points at `docs/pii.md`. A stale doctrine comment on exactly these columns is how the last incident propagated.
- **Dead Cognito-era spec removed whole** from `apps/backend/app.yaml`: the `real_name`-bearing `NewDJ` schema, the `/djs/register` paths, and the bare `/djs` GET that `$ref`s `NewDJ` in an `allOf`. All unreachable since `a0cd1979` — `routes/djs.route.ts` mounts only `/bin` and `/playlists` — and deleting the schema without the GET would have left Swagger-UI serving a dangling `$ref`.
- **The plan document itself** ships here, so the program is reviewable from inside the repo.

## Scope and follow-on

This is the prevent-recurrence layer only. The full program is `plans/dj-name-pii-safeguards.md` (added in this change): Track 0 stops the one live mirror writer, Track 2 unscrews the data structures (preserve-first `name → real_name` audit, a `databaseHooks.user` derivation choke point, provision-route validation change, one-shot backfill job), and Track 3b adds the sentinel wire spec. Each ships separately. The stored-data remediation on `flowsheet.dj_name` (BS#2281) is a separate, already-planned track and is not in scope here.

jakebromberg:
Tracks 1 (labels and warnings) + 3a (ESLint rule) of `plans/dj-name-pii-safeguards.md`, which ships in this PR. No behavior change — documentation, a DDL-only migration, comment rewrites, dead spec removal, and a new lint rule.

Closes #2293

### 1. `docs/pii.md` — the PII field registry

One table, one row per column in the DJ-name family: classification, contents, why the column exists, allowed read sites, and the enforcement mechanism (or an honest "none today", for `auth_user.email`). PII entries: `auth_user.real_name`, `auth_user.email`, tubafrenzy `DJ_NAME`. The non-PII entries are the point — `auth_user.name`, `auth_user.dj_name`, `flowsheet.dj_name`, `shows.legacy_dj_name`, `shows.dj_name_override` are listed explicitly so the classification question has a written answer instead of being re-derived (wrongly) per incident. Plus a short DJ-name-vs-real-name history section, and an Enforcement section using the `<!-- @rule id=… enforced-by=… -->` annotation form.

`auth_user.name`'s row is deliberately hedged: classified non-PII **post-backfill**, with an instruction not to treat a given row's `name` as authoritative for either purpose until `jobs/auth-user-name-backfill` (Track 2d) has run.

Registered as one bullet in CLAUDE.md's Topic guides router — one line only, since CLAUDE.md is already past its own doc-budget alarm threshold and the prose belongs in `docs/pii.md`.

### 2. `0157_dj-name-pii-column-comments.sql` — the doctrine, in the database

Five `COMMENT ON COLUMN` statements on `auth_user.real_name`, `auth_user.name`, `flowsheet.dj_name`, `shows.legacy_dj_name`, `shows.dj_name_override`. DDL-only, no `ALTER TABLE`. This is the copy that reaches someone who arrives via `psql`/`\d+` or a GUI and never opens the repo.

**The header documents a drizzle-kit 0.31.10 bug found while writing this migration**, because the next person to need a DDL-only migration will hit it. `drizzle-kit generate --custom` fails *silently and every time* when `meta/applied-hashes.json` is present: `preparePrevSnapshot` takes `snapshots[snapshots.length - 1]` off a directory listing sorted **alphabetically**, not by migration index, so `applied-hashes.json` (`'a'` sorts after `'0'`–`'9'`) is picked as "the previous snapshot", fails `pgSchema.parse`, and drizzle-kit exits 0 having written nothing — no error, no `.sql`, no journal entry, just the malformed-data notice and then silence.

This is a stronger claim than what `docs/migrations.md` currently records. That file describes the move-aside workaround as a just-in-case measure for plain `generate`, where the malformed file is merely skipped; under `--custom` it is mandatory. The header carries the full recipe (move `applied-hashes.json` aside, generate, move it back, hand-write the DDL, check the journal `when` against `previous + 1`, `drizzle:freeze-hashes`, `lint:migrations`), plus notes on why `COMMENT ON` needs no `IF NOT EXISTS` guard (idempotent, last write wins) and why it is PG14-safe by construction (plain SQL, unchanged since PG 7.2 — no release-notes check required under the `dev-prod-pg-version-skew` rule).

With the workaround applied, `meta/0157_snapshot.json` was written correctly, so the "duplicate the predecessor's snapshot" fallback was not needed here — the header says so explicitly rather than leaving the reader to guess which branch was taken.

### 3. Docblock sweep — 7 files, comment-only

`apps/backend/services/flowsheet.service.ts`, `apps/backend/routes/internal.route.ts`, `jobs/flowsheet-etl/job.ts`, `jobs/flowsheet-etl/fetch-legacy.ts`, `jobs/flowsheet-etl/backfill-legacy-ids.ts`, `jobs/legacy-dj-name-remediation/job.ts`, `shared/database/src/dj-name.ts`. Each carried some version of the `realName || name` mirror-forwarding claim or the "`auth_user.name` holds the real name" reasoning; each now describes the end state (`real_name` is the sole legal-name carrier, `auth_user.name` is a derived display handle/username with the backfill pending) and points at `docs/pii.md`.

**Comment-only, verified**: every hunk in these seven files is inside a comment or docblock — no executable line changed, which is why the full unit suite passing is a meaningful check on this section rather than a formality.

`schema.ts`'s `legacy_dj_name` docblock and `shared/legacy-mirror/` are deliberately untouched — they belong to the sibling Track 0 PR (#2292), which is where the mirror's behavior actually changes.

### 4. `apps/backend/app.yaml` — dead Cognito-era spec removed whole

The `real_name`-bearing `NewDJ` schema, the `/djs/register` POST/PATCH paths, and the bare `/djs` GET with its `cognito_user_name` query param. All unreachable since `a0cd1979` — `routes/djs.route.ts` mounts only `/bin` and `/playlists`.

**No dangling `$ref`**: the bare `/djs` GET `$ref`s `NewDJ` inside an `allOf`, so deleting the schema without also deleting that block would have left Swagger-UI resolving a reference to nothing. Both went together; `grep -n "NewDJ\|cognito_user_name\|djs/register" apps/backend/app.yaml` returns no matches, and the YAML still parses.

Note that `app.yaml` is Swagger-UI documentation; `wxyc-shared/api.yaml` is the schema-first source of truth. The corresponding `dj_name` description fixes there are a separate cross-repo PR.

### 5. `wxyc/restricted-real-name` ESLint rule

`eslint-rules/restricted-real-name.cjs`, registered in the merged `wxycLocalRules` object alongside the two existing local rules, scoped to `apps/**/*.ts`, `shared/**/*.ts`, `jobs/**/*.ts`. Flags `realName`/`real_name` as **both** member-expression access (`dj.realName`, `row['real_name']`) and object-literal property keys (`{ realName: … }`) — either shape alone misses about half the sites, and the originating incident was a construction site, not an access site.

The allow-list is baked into the rule module rather than passed through `eslint.config.mjs` options, specifically so `RuleTester` can exercise "an allow-listed file is valid" directly instead of asserting on a config the tests can't see.

Two deliberate non-features, both documented in the rule's own docblock so a future reader doesn't mistake them for oversights:

- **No `user.name` clause.** The incident shape is alias-bound (`dj.name` at the mirror call site), so an identifier-keyed matcher on `name` would have caught none of the historical cases — and `name` is a column on at least three tables here, so it would fire constantly on code with nothing to do with DJ identity. Claiming coverage the rule can't deliver is the false-assurance failure the plan indicts elsewhere; that half of the risk is closed by Track 0 and the registry.
- **`tests/**` is outside the scope.** At least seven unit fixtures legitimately construct `realName` objects, per-file allow-list entries for them would rot, and a fixture literal cannot reach a wire. Precedent both ways exists in this config (`source-tagged-constraint` scopes narrowly; `no-bare-array-in-sql-template` deliberately includes tests), so the exclusion rests on its merits, not on habit.

Known gaps are documented rather than papered over: destructuring (`const { realName } = user`, an `ObjectPattern`) and type-only declarations (`TSPropertySignature`) aren't inspected. Neither appears at a non-allow-listed site today, verified by grep.

**How allow-list completeness was proven**: `npm run lint` across the full tree finishes with **0 errors and zero `restricted-real-name` findings**. Since the rule is active over `apps/**`, `shared/**`, and `jobs/**`, a single missing allow-list entry would have surfaced as an error on a legitimate existing PII site. Zero findings against the real tree is the evidence that the allow-list is complete — not an assertion, an observation. 19 rule tests at `tests/unit/database/restricted-real-name.rule.test.ts` cover both AST shapes, computed and non-computed access, allow-listed exact paths, the `jobs/auth-user-name-backfill/` prefix entry, and the out-of-scope shapes above.

## Local gate

Run on the rebased branch (`origin/main` @ `d8306903`):

```
npm run format:check   → All matched files use Prettier code style!
npm run typecheck      → clean (all 5 workspaces, tsc --noEmit)
npm run lint:migrations → Migration journal validation passed (155 entries, 0 warning(s)).
npm run lint           → ✖ 962 problems (0 errors, 962 warnings)   [exit 0; all 962 pre-existing, 0 from the new rule]
npx jest --config jest.unit.config.ts --testPathIgnorePatterns "tests/unit/jobs/flowsheet-etl"
                       → Test Suites: 476 passed, 476 total / Tests: 8361 passed, 8361 total   [exit 0]
npx jest --config jest.unit.config.ts tests/unit/jobs/flowsheet-etl
                       → Test Suites: 9 passed, 9 total / Tests: 197 passed, 197 total   [exit 0]
```

The `flowsheet-etl` suites were run separately out of caution for a known import-time `process.exitCode` quirk; on this branch they exited 0 with all 197 tests passing, so nothing was masked. The full suite (all 485 files, 8558 tests) was re-run green after the fix commit below.

### One CI failure, fixed in `39a38909`

**The gate above passed and the migration still failed to apply** — worth recording, because the gap is reusable. The first CI run failed both `Integration-Tests` (at "Initialize database") and `Migration Dry-Run (prod-shaped data)` with `42P01 relation "wxyc_schema.auth_user" does not exist`.

This repo has two schemas, and the hand-written DDL guessed the wrong one for `auth_user`: it is declared with a bare `pgTable(...)` in `schema.ts`, so it lives in **`public`**, while `flowsheet` and `shows` are `wxyc_schema.table(...)`. Drizzle's own generated SQL carries both forms side by side — 0021 has `REFERENCES "public"."auth_user"("id")` two lines away from `ALTER TABLE "wxyc_schema"."shows"` — so the convention was established; only the hand-written statements missed it.

`lint:migrations` cannot catch this class of error: it validates the journal, not the DDL. Nothing in the local gate applies a migration, which is exactly why a green gate and a broken migration coexisted. The fix commit therefore does three things rather than one:

1. Corrects the two `auth_user` statements to `"public"."auth_user"`.
2. Adds the qualification trap to the migration header's recipe — which tables live in which schema, how to tell from `schema.ts`, and the explicit statement that `lint:migrations` will not catch a wrong guess.
3. Names the commands that *do* catch it (`npm run db:start`, or `node dev_env/init-db.mjs` against a throwaway Postgres) so the next person has a local check that covers the gap.

Verified by applying the full chain against a throwaway `postgres:18.0-alpine`: all 155 migrations applied, and `col_description()` confirms the five comments landed on `public.auth_user.{real_name,name}` and `wxyc_schema.{flowsheet.dj_name,shows.legacy_dj_name,shows.dj_name_override}` — the only column comments in the database. `drizzle:freeze-hashes` re-run afterwards (0157's entry is the only hash that moved).

CI is green on the fix: all six jobs pass, including the migration dry-run against prod-shaped RDS data.

## Related

- #2292 (`bugfix/mirror-djhandle-pii`) — Track 0 of the same plan: stops the live mirror writer at `http-mirror.ts:512` from putting legal names into tubafrenzy's public `DJ_HANDLE`. **Independent of this PR** — neither blocks the other, and the two touch disjoint files by design (that PR owns `shared/legacy-mirror/` and `schema.ts`'s `legacy_dj_name` docblock; this one leaves both alone). They can land in either order.
- `plans/dj-name-pii-safeguards.md`, added here, is the full program: Track 2's derivation hook, preserve-first backfill, and provision-route change, plus Track 3b's sentinel wire spec, all ship separately.


jakebromberg:
Addressed a review finding plus a companion plan amendment, pushed as `19fc2190`:

- **Finding**: `docs/pii.md`'s `real-name-restricted-lint` `@rule` annotation body was 2009 chars, over the repo's 900-char compress-candidate threshold — `check:doc-rules` was surfacing a new warning that violated this PR's own acceptance criterion. Compressed the body to a one-line pointer at `eslint-rules/restricted-real-name.cjs`'s docblock, which already carries the full rationale, allow-list, and known gaps. Reproduced the warning first, then confirmed it's gone and the 9 other (pre-existing, unrelated) warnings are unchanged.
- **Plan amendment**: `plans/dj-name-pii-safeguards.md`'s 2a preserve-first predicate — also duplicated verbatim in 2d's backfill precondition gate — false-positives on rows where `name` is merely the resolved on-air handle (per BS#2297), which would copy a handle into the `real_name` PII column instead of a legal name. Added a handle exemption (`trim(name) IS DISTINCT FROM resolveDjDisplayName(dj_name)`) to both copies, noting the handle-is-real-name edge case is exempt by construction, same as the flowsheet scrub.

`check:doc-rules`, `format:check`, `lint:migrations`, and the `restricted-real-name` rule suite all pass; CI is green on the pushed commit.

jakebromberg:
Simplify pass follow-up, pushed as `aa0013d6` (CI green, both runs):

- **Migration recipe relocated**: the 95-line drizzle-kit `--custom` recipe moved out of `0157`'s SQL header into a new **docs/migrations.md** section; the migration's hash re-frozen since editing the .sql changes its SHA-256.
- **Rule cleanup**: `restricted-real-name.cjs` drops its flat-config-only compat fallbacks and garbled docblock, and its message text now matches the sibling rules' shape.
- **Allow-list hygiene**: the unexercised `derive-user-display-name.ts` allow-list entry is removed, with its test fixture converted to prove the removal took effect.
- **Plan amendment**: the 2a procedure now makes the backfill job's TypeScript predicate the sole arbiter — the reviewed manual SQL sources its id list from the job's dry-run output instead of re-deriving the predicate in SQL text (eliminating the dual-language predicate drift risk).

Issue #2313, opened 2026-08-29
Convert flowsheet-etl and the tubafrenzy webhook's dj_name writes from a re-derived COALESCE to the canonical resolveShowDjName helper


jakebromberg:
## Problem

Two live writers still re-derive the `flowsheet.dj_name` resolution chain in SQL instead of calling the canonical `@wxyc/database` helpers (`resolveShowDjName` / `resolveDjDisplayName` / `showDjNameOverride`, `shared/database/src/dj-name.ts`):

- `jobs/flowsheet-etl/job.ts:121` — `SET dj_name = COALESCE(u.dj_name, s.legacy_dj_name)`, scoped `dj_name IS NULL`.
- `apps/backend/routes/internal.route.ts:195` — the same two-column `COALESCE` on the tubafrenzy webhook path.

Both predate `dj_name_override` (BS#1321) and omit the literal-`"Anonymous"` filter (BS#1286), so both can write a `dj_name` value the canonical helper would never produce — on a show carrying an override, or a DJ whose stored handle is literally `"Anonymous"`.

This is exactly the anti-pattern `shared/database/src/dj-name.ts`'s own docstring names as the reason the helper was extracted in the first place: "`jobs/flowsheet-april-gap-import` shipped a `COALESCE(auth_user.dj_name, shows.legacy_dj_name)` copy that predated `dj_name_override` (BS#1321) and omitted the literal-'Anonymous' filter (BS#1286)."

## Why this matters now

WXYC/Backend-Service#2281 (the `flowsheet-dj-name-scrub` PII scrub) had to design around these two writers rather than fix them: its `verifyComplete`/`verifyScrub` divergence check is explicitly bounded to `id <= <drain high-water mark>` specifically because these two writers can produce a non-canonical value on a row *after* the scrub's drain has already passed it, and an unbounded check would then fail nondeterministically. `jobs/flowsheet-dj-name-scrub/README.md`'s "Verification" section states plainly: "divergence regrows at the rate those two writers run... this job is not complete without it" (i.e. without this conversion).

In other words: BS#2281 removes the existing pollution once, but as long as these two writers exist, new pollution of the same two shapes (missing override, un-filtered "Anonymous") keeps entering the table at whatever rate `flowsheet-etl` and the tubafrenzy webhook run.

## Desired end state

[...]

Issue #2312, opened 2026-08-29
flowsheet-dj-name-scrub: re-render show_start/show_end messages around the recomputed handle instead of degrading to nameless


jakebromberg:
## Problem

`jobs/flowsheet-dj-name-scrub` (BS#2281) has a `message` rewrite pass (`rewriteMessage` in `jobs/flowsheet-dj-name-scrub/orchestrate.ts`) that removes PII embedded in marker `message` text. For `show_start` and `show_end` rows, when the embedded name is a roster real name, the pass degrades the message to a nameless form:

- `Start of Show: ${name} joined the set at ${t}` -> `Start of show: ${t}`
- `End of Show: ${name} left the set at ${t}` -> `End of show: ${t}`

That degraded wording matches what `startShow`/`endShow` themselves already emit when the DJ name is unresolvable, so it introduces no new shape into the corpus — but it also discards attribution that is recoverable. Unlike `dj_join`/`dj_leave` (where the joining guest's identity genuinely is not recoverable from `shows` — it belongs to a different DJ than the show's `primary_dj_id`), the correct handle for a `show_start`/`show_end` row IS available: it is the same `shows` join the scrub's main pass (`loadMainPage`, the `dj_name` recompute) already used for that exact row, one table over.

`rewriteMessage`'s docstring explains why it does not re-render around a name the main pass changed for a *non-PII* reason (a per-show override added after the fact) — that is a documented, deliberate omission. It does not address this PII case, which was raised in review and deliberately deferred rather than implemented in that PR.

## Desired end state

For a `show_start` or `show_end` row whose stored `message` embeds a roster real name, the message pass re-renders the message around the row's RECOMPUTED handle (the same value the main pass just wrote to `dj_name`) instead of degrading to the nameless form — i.e. `Start of Show: ${recomputedHandle} joined the set at ${t}` / `End of Show: ${recomputedHandle} left the set at ${t}` — preserving attribution instead of discarding it. The nameless degrade remains the fallback when the recompute itself resolves to `null` (no override, no usable handle, no legacy handle) — that case is unchanged from today's behavior.

## Where

- `jobs/flowsheet-dj-name-scrub/orchestrate.ts` — `rewriteMessage`, `MESSAGE_TEMPLATES.show_start` / `.show_end`, the `message` pass's `decideForMessagePass` / `runPass('message', ...)` wiring.
[...]

PR #2316, opened 2026-08-29
Scrub historical denormalized flowsheet.dj_name to the current PII-safe resolution policy

### LINKED ISSUE #2281: Scrub historical denormalized dj_name values to match the current resolution policy
## Problem

`flowsheet.dj_name` is a denormalized snapshot of the DJ display name, written onto each flowsheet row at play time. The resolution helper that feeds it is correct today — `resolveDjDisplayName` takes only the handle and returns `null` for null, empty, whitespace, and the `Anonymous` literal, following #1286 (epic #1288).

Rows written before that change retain values resolved under the older chain, so the stored column no longer matches the policy that governs new writes. `shows.legacy_dj_name`, backfilled from tubafrenzy, is a second store with the same mismatch: `resolveShowDjName` reads it as the third step of its chain, and the helper being correct does not make the values it reads conform.

The distinction that matters here is between a resolution helper being correct and the stored data conforming to it. Comments describing the chain as PII-safe are accurate about the code path and were not intended as claims about the contents of these two columns.

## End state

Both columns hold only values that the current resolution policy would produce, and no derived index carries a stale value. Specifically:

- `flowsheet.dj_name` contains a handle, or `NULL`. No value that is an `auth_user.name` rather than an `auth_user.dj_name`.
- `shows.legacy_dj_name` likewise.
- `flowsheet.search_doc` reflects the scrubbed values. This needs no separate step: it is a `STORED GENERATED` column whose expression includes `dj_name`, so the scrub's `UPDATE` recomputes it per row.
- A regression guard exists so a future denormalization cannot reintroduce the mismatch.

## Files and paths

- `apps/backend/services/search.service.ts` — `DJ_NAME_EXPR`, `buildDjNameMatch`; `search_doc` is the index that needs rebuilding.
- `apps/backend/utils/flowsheet-projection.ts` — `projectFlowsheetEntry` passes `dj_name` through unchanged, so every entry-serving endpoint reflects the column directly.
- `apps/backend/services/flowsheet.service.ts` — `resolveShowDjName`, and the show-heading path that reads `shows.legacy_dj_name`.
- `shared/database/src/dj-name.ts` — `resolveDjDisplayName`, the current policy; the scrub should reproduce its semantics rather than restate them.
- Migration for the scrub itself, following the numbering after 0083.

## Constraints

- **Batch the updates.** ~2.6M rows, each recomputing a generated tsvector column. Migration 0053 already wedged on locks once (#511, which forced splitting the DDL from the backfill); this is not a single transaction.
- **Data safety.** This rewrites historical rows in the flowsheet table. Run a `SELECT` with the same `WHERE` clause and record the counts before any `UPDATE`, take a backup, and make the migration reversible or at minimum replayable.
- **Define the scrub as a recompute, not as an equality match.** For each row, recompute what the current chain would produce from current inputs (post-scrub `legacy_dj_name`, current `auth_user.dj_name`, `dj_name_override`) and overwrite where the stored value differs. Do not scrub by pattern-matching what "looks like" a real name — but equally, do not define it as "rows whose value equals an `auth_user.name`": the largest affected cohort belongs to legacy shows with **no `auth_user` row at all**, so a rule joining through `auth_user` cannot see it. This is the same blind spot that led the #1393 post-run verification to report an empty residue class; see that issue's closing comment.
- **Cover all entry types.** The #1393 remediation filtered `entry_type IN ('show_start','show_end','dj_join','dj_leave')`. Search serves `entry_type = 'track'`. Those sets are disjoint, so track rows have never been re-resolved and are the bulk of the work here.
- Where a DJ has since set a handle, prefer that handle over `NULL` — it preserves attribution, which is the reason the column exists.
- `NULL` is already a supported value on this column by contract, and consumers render it as `Unknown DJ`, so nulling is safe for clients.
- **No separate reindex step.** An earlier revision of this issue asked for one. `flowsheet.search_doc` is `STORED GENERATED` over an expression containing `dj_name` (`shared/database/src/schema.ts:1295`), so updating `dj_name` regenerates it in the same write — migration 0054 relies on exactly this: "the `dj_name` UPDATE recomputes the `search_doc` generated column (`dj_name` is in the expression) and dj-name search becomes correct". One pass fixes both the displayed value and what the tsvector matches, with no intermediate state where they disagree. A `VACUUM`/`REINDEX` afterwards is bloat hygiene after ~2.6M updates, not correctness.

## Acceptance criteria

- [ ] A count, from a query rather than an estimate, of affected rows in each of the two columns.
- [ ] Migration scrubs `flowsheet.dj_name`, preferring a current handle where one exists and `NULL` otherwise.
- [ ] Migration scrubs `shows.legacy_dj_name` on the same basis.
- [ ] Spot-check confirming `search_doc` regenerated for touched rows (expected automatically, given the generated column — verify rather than implement).
- [ ] Test asserting no served `dj_name` equals any `auth_user.name`, run against a fixture containing a user whose `name` and `dj_name` differ and one with `dj_name` unset.
- [ ] Existing `resolveDjDisplayName` / `resolveShowDjName` unit tests still pass unchanged — this issue does not alter the helpers.

## Related

- `2a37bbc6` (2026-06-08, in PR #1374) — the commit that actually removed the `auth_user.name` fallback, and therefore **the cutoff date for "pre-fix rows" in this issue**.
- #1286, epic #1288 (2026-06-03) — centralized the resolution helper but deliberately kept the real-name arm (`resolveDjDisplayName(djName, name)`), so it is *not* the cutoff. An earlier revision of this issue cited it as such; a scrub using 2026-06-03 would leave five days of rows behind.
- #1393 — the previous remediation of this data, marker-rows-only; this issue is the rest of it.
- #1060 — dropped `flowsheet_dj_name_trgm_idx` in migration 0083; relevant background on how this column is indexed.
- Consumers `WXYC/website` `/airplay-search` and `/playlists/archive` were removed pending this work, which also leaves `wxyc.info/playlists/searchPlaylists` and `wxyc.info/playlists/radioWeek` without successors at the 2026-09-07 tubafrenzy cutover. Restoring those pages depends on this issue.

Maintainers: see the draft security advisory on this repository for the full assessment, measured extent, and a suggested interim mitigation.

---

_Milestone 1 moved from 2026-08-31 to **2026-09-07** on 2026-08-28 — 24 scheduled DJs still lack a usable Backend account and the classic flowsheet UI is their only surface. Engineering readiness was not the gate. See [WXYC/wiki#125](https://github.com/WXYC/wiki/issues/125)._

jakebromberg:
Closes #2281

## What this is

A one-shot, **dry-run-by-default** job (`jobs/flowsheet-dj-name-scrub`) that recomputes historical `flowsheet.dj_name` to the current PII-safe resolution chain, rewrites the marker `message` text that embeds a DJ's real name, and ships a prerequisite SSE age-guard so the drain doesn't push millions of events at live clients.

`flowsheet.dj_name` is a denormalized snapshot written at play time. The write path has been correct since `2a37bbc6` (2026-06-08), but snapshots frozen under two superseded policies are still in the table, still served raw by every public flowsheet endpoint, and — through `GET /flowsheet/search` — still *matchable*: `search_doc` carries `dj_name` at weight B and `buildDjNameMatch` ILIKEs the column, so a real name is a working search key on a public, unauthenticated endpoint.

- **Cohort A — the `auth_user.name` era.** Between `a0cd1979` (2025-12-30) and `2a37bbc6`, every writer fell back to better-auth `name`, which dj-site's admin provisioning filled with the DJ's real name.
- **Cohort B — the tubafrenzy `DJ_NAME` era.** The 2026-04-27 backfill behind migration 0053 froze `COALESCE(u.dj_name, s.legacy_dj_name, u.name)` onto ~2.6M rows while `legacy_dj_name` held `DJ_NAME` (the full real name) rather than `DJ_HANDLE`.

**Cohort C is deliberately out of scope**: real names DJs themselves typed into tubafrenzy's free-text `DJ_HANDLE`. That is a policy question, not a correctness bug. **Sampling after this job will still show name-shaped residue — that is expected, not a failed run.**

## Design

Page rows out, compute the expected value **in TypeScript** with the canonical `@wxyc/database` helpers (`resolveShowDjName` / `resolveDjDisplayName` / `showDjNameOverride`), write back only what differs. Parity with the serving path is by construction — there is no second copy of the chain to drift. Re-deriving the chain in SQL is the documented anti-pattern (`shared/database/src/dj-name.ts:1-13`); a scrub that exists to remove a helper-vs-data divergence must not reintroduce it.

Idempotency comes from comparing against the **recomputed** value (`IS DISTINCT FROM`), never from `IS NULL`. That is the specific defect that left the previous remediation (#1393) incomplete: its live UPDATE was scoped `dj_name IS NULL`, so a row already holding a polluted value was never corrected, and all four of its query sites filtered marker types while `GET /flowsheet/search` serves `entry_type = 'track'` — the remediated set and the searchable set were disjoint.

### Scope is per entry type, and the `show_start` split is load-bearing

| `entry_type` | Reconciled against | Action |
| --- | --- | --- |
| `track` | `resolveDjNameForShow` → `resolveShowDjName` | Recompute |
| `show_end` | same (BS#2068) | Recompute |
| `show_start`, `primary_dj_id IS NOT NULL` | `startShow` — override → handle, **no legacy fallback** | Recompute against *that* chain |
| `show_start`, `primary_dj_id IS NULL` (legacy) | `jobs/flowsheet-etl` (tubafrenzy codes 1 and 9) | Recompute via the shows chain |
| `dj_join` / `dj_leave` | — | **PII-null probe only.** Never recomputed from the shows join |
| `talkset`, `breakpoint`, `message` | — | **Never touched, under any pass** |

Applying `startShow`'s no-legacy chain to the legacy cohort resolves `null` for every one of those rows — they have no `primary_dj_id`, so there is no user row to read — wiping what migration 0053 wrote. `flowsheet.service.ts` documents that exact outcome as the bug BS#2068 fixed on `show_end`: *"the old form resolved `null` for the ENTIRE legacy cohort (2,813 of production's 2,814 open shows on 2026-08-21)"*. A test pins the non-regression.

The `talkset`/`breakpoint`/`message` exclusion is the most dangerous failure mode in the job: those rows are deliberately NULL, and under a bare `IS DISTINCT FROM <shows chain>` every one becomes a candidate and gets **newly populated** — a PII scrub that invents DJ names on rows that never had one. It is a named constant with tests in both tiers rather than an implicit consequence of a WHERE clause.

### Three intentional-NULL classes

A `NULL dj_name` after this run is the fix, not a regression, and two deployed jobs would reverse it — `jobs/legacy-dj-name-remediation` and `jobs/flowsheet-dj-name-backfill` both fill `dj_name WHERE dj_name IS NULL` from the shows join, which re-attributes a guest DJ's row to the primary DJ. Both now refuse at runtime and have had their root Dockerfiles removed, so `Manual Build & Deploy` cannot produce an image either; a test fails if one is restored without a deliberate decision. Deliberately no override env var — reviving one means reverting the guard under review, which is the point.

## Scope reduction you should see rather than discover

**The issue asks for `shows.legacy_dj_name` to be scrubbed too. This PR does not scrub it.** The evidence is that BS#1393 already rewrote that column from tubafrenzy `DJ_NAME` to `DJ_HANDLE`, so what remains there is Cohort C. But the recompute reads `legacy_dj_name` as a direct chain input with no PII check of its own, so if that premise is wrong for any subset of shows, this job actively writes real names onto every in-scope row of those shows.

Rather than leave that asserted, the job ships a startup pre-flight that counts `shows.legacy_dj_name` values which are themselves roster real names, runs in **both** dry-run and execute mode before the first pass, and warns loudly (structured log + Sentry) when the count is non-zero. Read it before `--execute`.

## Change-class provenance

The first prod dry run reported 1,826,070 track rows differing from the canonical value — 86.7% of everything scanned. That is consistent with Cohort B, but "consistent with" is a story, and 1.8M rows on the live on-air table deserve better than a story. Every would-be write is now classified by *why* it differs, with counts and an id-only sample per class.

Two classes decide whether to proceed. `stored_is_roster_real_name` **plus** `stored_is_superseded_legacy_name` are the PII-removal evidence, and they must be read together: the first is keyed on `auth_user` and therefore cannot see the largest affected cohort — legacy shows with no `auth_user` row at all — which is the same blind spot that made #1393's post-run verification report an empty residue class. The second closes it with no new query, off the `legacy_dj_name` already joined in for the recompute. `recomputed_null_non_pii` should be near zero; a non-trivial count means the job is erasing handles it cannot justify as PII removal.

`recomputed_is_roster_real_name` is a different signal — what a write would itself **write**, not what it removes. It is counted and warned on, never a hard failure: correcting it means widening into Cohort C, so failing the run would be a permanent red over a condition this job has no in-scope way to fix.

## SSE fan-out guard — extracted to #2317, and a deploy prerequisite

**Split out of this PR: #2317.** That guard is a behavior change to a live production broadcast path, it stands on its own, and it should deploy *before* this job ever runs rather than alongside it — so it gets its own review and its own revert handle. Nothing here depends on it at build or test time; the dependency is operational. **#2317 must be merged and deployed before the first `--execute` run of this job.**

`filterMetadataUpdate` broadcast on **any** flowsheet UPDATE landing in a terminal `metadata_status`, with no age guard and no `entry_type` check, and historical `track` rows are almost all terminal. This drain would have emitted one projected `liveFs:update` per row to every `/events/stream` client on every backend instance, for hours. (`jobs/flowsheet-metadata-backfill` has been paying a smaller nightly version of the same cost.)

The guard, its cost, and its `SSE/UpdateSuppressed` counter are described in #2317.

## Verification bound

`verifyScrub` re-scans read-only and counts rows that would still be written, **bounded to `id <= <drain high-water mark>`**. The bound is correctness, not caution: `jobs/flowsheet-etl/job.ts:121` and `apps/backend/routes/internal.route.ts:195` both still `SET dj_name = COALESCE(u.dj_name, s.legacy_dj_name)`, both predate `dj_name_override` (BS#1321) and omit the literal-`'Anonymous'` filter (BS#1286), so an unbounded check would fail nondeterministically on any row they touch after the drain passed it. **Divergence regrows at the rate those two writers run** — converting them is tracked as #2313, and this work is not complete without it.

## Run procedure

Dry-run is the default and writes nothing; `--execute` writes; passing both is a hard error.

1. Confirm #2317's SSE guard is deployed, and run outside peak listening hours — `touch_flowsheet_watermark` advances `last_modified_at` on every statement, so every polling client gets a `200` instead of a `304` for the whole drain.
2. Dry run. Read the per-pass counts and the change-class breakdown, and the `shows.legacy_dj_name` pre-flight.
3. `--execute`.

Full runbook, the WAL/NOTIFY budget, the resume cursors, and the known limits are in `jobs/flowsheet-dj-name-scrub/README.md`.

## Why this PR is still large

At ~4,400 lines it is over the repo's 1,000-line guideline, and it does not decompose further along any seam that leaves both halves coherent. The two slices that *were* separable are out: #2317 (the SSE guard, ~400 lines) and, still here because its refusal message is only meaningful once the scrub exists, the reversal guards on the two superseded jobs (~350).

What remains is one job plus its tests: ~1,830 lines of source (1,569 of it `orchestrate.ts`), ~1,920 of tests, ~370 of docs and plan. Splitting it by pass — `dj_name`, then `message`, then orphan — would land a partial privacy fix that reads as a complete one: the `dj_name` pass alone leaves the DJ's real name still rendered in the marker `message` text. Splitting scaffolding from logic ships a Dockerfile that builds nothing. Neither is a better review artifact than one coherent job.

Nothing here is live on merge: the job is one-shot, registers no cron, and is dry-run-by-default. The thing that needs sequencing is the deploy of #2317, not the merge of this.

## Follow-ups

- #2312 — re-render `show_start` / `show_end` messages around the recomputed handle instead of degrading to the nameless form. Deferred deliberately: it widens the message pass from "remove PII" to "restore attribution".
- #2313 — convert the two live SQL re-derivers to the canonical helper.


jakebromberg:
Rebased to drop the SSE age guard, now #2317. The two PRs are disjoint rather than stacked — this branch no longer contains those commits, so either can merge first and neither blocks the other's review. The dependency is operational only: **#2317 must be deployed before the first `--execute` run of this job**, which is stated in the run procedure and in `jobs/flowsheet-dj-name-scrub/README.md`.

Diff here drops from ~4,810 to ~4,440 lines. Still over the 1,000-line guideline; the PR body now says which slices were separable, which were not, and why splitting the job itself by pass would produce a partial privacy fix that reads as a complete one.

PR #2317, opened 2026-08-29
Age-guard the liveFs:update SSE broadcast so a bulk flowsheet UPDATE can't flood live clients

jakebromberg:
## The bug

`filterMetadataUpdate` broadcasts a `liveFs:update` on **any** `flowsheet` UPDATE landing in a terminal `metadata_status`. It has no age guard and no `entry_type` check, and historical `track` rows are almost all terminal.

So any bulk UPDATE over `flowsheet` emits one projected `liveFs:update` per row, to every `/events/stream` client, on every backend instance, for the entire run. `jobs/flowsheet-metadata-backfill` has been paying a smaller nightly version of this cost for months.

The insert path received exactly this guard in BS#2131 (`LIVE_FS_INSERT_MAX_AGE_HOURS`, default 24h). The update path never did — `metadata-broadcast.ts` and `docs/env-vars.md` both stated that omission was deliberate under the BS#1534 doctrine. That was true when written; it is not any more.

## The fix

An `add_time` age guard on the update path, mirroring the insert sibling, under its own knob: **`LIVE_FS_UPDATE_MAX_AGE_HOURS`** (default 24h).

Deliberately a separate knob rather than sharing the insert one — the two paths have different volume profiles, and an operator narrowing one for the duration of a bulk run should not silently move the other.

A new `matchTerminalTrackUpdate` helper holds the structural match, shared by `filterMetadataUpdate` and `isAgeSuppressedUpdate` so the two can never disagree about what the guard applies to — the same shape `matchTrackInsert` already gives the insert path. The `entry_type === 'track'` check inside it is a **no-op today**: every enrichment writer already gates on `entry_type = 'track'`, and `metadata_status` defaults to `'pending'` for marker rows, so no non-track row reaches a terminal status. It exists so a future writer that does cannot start pushing marker rows onto a stream whose payload is typed as a track row and which dj-site patches into a track cache.

The two age resolvers now share one `parseMaxAgeMs` body rather than a hand-kept copy. The `0`-is-rejected rule is the load-bearing part, and two copies of it is precisely how one path ends up quietly accepting `0` as "disabled" — which here would classify every update as historical and take the whole `liveFs:update` feed dark station-wide.

## What it costs

This is a narrowing, not a pure win, and the docstring and env-vars entry both say so: an enrichment landing on a row older than the threshold no longer emits its "refetch me" signal, so a client viewing a historical range loses a live patch it would otherwise have received and must refetch on its own. Enrichment of anything played in the last day — the case that actually matters to a `/live` viewer — is unaffected.

Same posture as the insert guard otherwise: **fails open** on a missing or unparseable `add_time`, warn-and-default on any bad value with a warn-once latch per distinct raw value, and "now" measured from the CDC event's own database-clock `timestamp` rather than `Date.now()`.

## Observability

Suppressions are counted as `SSE/UpdateSuppressed`, dimensioned by `Topic` plus a dimensionless companion (the org's alarm-input convention). Kept as its own metric rather than folded into `SSE/InsertSuppressed` because the two answer different questions — a spike here means a bulk UPDATE is running, or the threshold is misconfigured, which is a distinct operational condition from a bulk import.

Expect the counter to climb for the duration of a deliberate bulk UPDATE. **That climb is the guard working.** A spike with no such job running means the threshold is misconfigured.

## Why this is its own PR

Extracted from #2316 (the BS#2281 `flowsheet.dj_name` scrub), which is what surfaced it: that job is a multi-hour drain over ~2.6M rows and without this guard it would push millions of unsolicited events at live connections. But nothing here depends on that job — this is a behavior change to a live production broadcast path, it stands on its own, it should deploy before the scrub runs rather than alongside it, and it deserves its own review and its own revert handle. It also fixes the smaller nightly cost `flowsheet-metadata-backfill` is paying today, regardless of whether the scrub ever runs.

## Related

- #2316 — the dj_name scrub, whose README lists this guard's deployment as a precondition for its first live run
- #2281 — the underlying issue


jakebromberg:
Split out of #2316 so it can be reviewed and deployed on its own. #2316's README lists this guard's deployment as a precondition for the dj_name scrub's first live run, so this should merge and deploy first — but nothing in #2316 depends on it at build or test time, and this fix stands on its own regardless: `jobs/flowsheet-metadata-backfill` has been paying a smaller nightly version of the same unguarded fan-out already.

Issue #2233, opened 2026-08-21
POST /flowsheet/join: explicit `intent`, 409 on someone else's open show, and takeover


jakebromberg:
## Problem

`joinShow` (`apps/backend/controllers/flowsheet.controller.ts:824`, branch at `:888`) routes purely on whether the newest show is open:

```js
if (current_show?.end_time !== null || latestEntryIsShowEnd) { startShow(...) }
else { addDJToShow(...) }   // ← silent co-host join
```

So a DJ going live while someone else's show is open is silently made a co-host of *their* show. `primary_dj_id` stays with the departed DJ, `on_air` reports the wrong name, and — because `leaveShow` (`:912`) only calls `endShow` when `dj_id === primary_dj_id` — the new DJ has no way out. See WXYC/Backend-Service#2232 for the incident.

## Desired end state

| request | behavior |
|---|---|
| no `intent` | different DJ's show open → **409**. Otherwise unchanged. |
| `intent: "join"` | today's `addDJToShow` co-host path |
| `intent: "takeover"` + `expected_show_id` | `endShow(current_show)` then `startShow(dj_id, …)` |

## Routing rules — order matters, existing guards must not weaken

1. BS#1861 `latestEntryIsShowEnd` and BS#2065 `closeShowFromTerminalShowEndMarker` run **first, unchanged**. A show they prove closed still routes to `startShow` with no prompt.
2. `isDjAlreadyActiveOnShow` (`:897`) keeps its no-op-duplicate 200 and is checked **before** the 409 — a DJ re-pressing their own toggle must never be prompted.
3. Only a genuinely-open show belonging to someone else reaches the 409.
4. Unknown `intent` → 400. `intent: takeover` without `expected_show_id` → 400. `expected_show_id` not matching `getLatestShow()` → re-409 with fresh details, nothing closed.

`takeover` composes the existing `endShow` + `startShow` in that order, so `endShow`'s `WHERE end_time IS NULL` CAS still serializes two racing takeovers — the loser gets its existing 400 rather than creating a second show.

## 409 body

```js
throw new WxycError('A show is already on air', 409, {
  code: 'show_already_open',
  details: { show: { id, dj_name, start_time } },
});
```

[...]

Issue #2232, opened 2026-08-21
[Epic] Go Live must not silently co-host — takeover prompt, sign-out close, operator close


jakebromberg:
Incident-driven epic. On 2026-08-20 `dj ronnabyte` started show `1951164` on dj-site at 11:03 AM PDT and signed out without ending it. Nine hours later the show was still open and had silently absorbed three more DJs as co-hosts — `dj bubblegum`, `DJ boy`, `dj onomatopoeia` — because `joinShow` routes on nothing but "is the newest show open." Its `primary_dj_id` stayed with the departed DJ, so `GET /flowsheet` reported `on_air: dj ronnabyte` all evening while someone else was actually broadcasting, and `GET /flowsheet/djs-on-air` disagreed with it. No co-host could end the show; the only fix was a manual DB remediation.

The full plan follows. It went through six rounds of `/review-plan` and every code citation in it has been verified against the repos.

---

# Go Live must not silently co-host

> Revision 8. **§6's premise in revision 7 was wrong and is corrected below** — the tubafrenzy "admin open-shows page" it cited is unreachable dead code (`FixRadioShowListServlet` has no `@WebServlet` and no `web.xml` mapping; the `/jsp/openShows.jsp` it forwards to does not exist in the repo). The real capability is different, and more permissive. §5 also gained two mechanism-level fixes, one of which would have silently reproduced the incident.
>
> Revision 7. Adds two mechanisms discovered after the incident: **§5 sign-out ends the show** (Jake's proposal, and the root-cause fix — Ronnabyte *did* sign out) and **§6 an open-shows admin surface** (parity with a tubafrenzy tool the migration dropped, which dies with turndown on 2026-08-31). The plan now covers four complementary mechanisms and is really an epic with children; see Delivery.
>
[...]

PR #2325, opened 2026-08-29
test(flowsheet): flag-ON integration coverage for takeover, and the six join sites it breaks

### LINKED ISSUE #2309: Flag-ON integration coverage for POST /flowsheet/join takeover, and the four join sites it breaks
## Problem

[#2308](https://github.com/WXYC/Backend-Service/pull/2308) shipped the `intent` / `expected_show_id` routing for `POST /flowsheet/join` behind `FLOWSHEET_TAKEOVER_ENABLED`, default OFF, with unit coverage only. The integration half of [#2233](https://github.com/WXYC/Backend-Service/issues/2233)'s acceptance criteria is not there, because turning the flag on in the integration environment is a prerequisite change with its own blast radius rather than a tail on that PR.

## Why it was split

With the flag off, `POST /flowsheet/join` co-hosts exactly as it always has, so every existing integration spec is valid unchanged. Turning it on converts four sites that deliberately join a *secondary* DJ onto an already-open show into 409s:

- `tests/integration/flowsheet.spec.js:339`
- `tests/integration/flowsheet.spec.js:1424`
- `tests/integration/flowsheet.spec.js:1449`
- `tests/integration/metadata.spec.js` (five sites join `getTestDjId()` with the secondary token)

Each needs auditing individually — a silent 409 skips the co-host setup and surfaces as a confusing assertion failure several lines later, not at the join.

## Desired end state

`tests/integration/flowsheet-takeover.spec.js`, a sibling of `flowsheet-join-after-tubafrenzy-signoff.spec.js`, running against real Postgres with the flag on:

- Seed an open BS-native show with a primary DJ and one active co-host, mirroring the incident's shape.
- A takeover closes it: `end_time` set, `show_end` marker written, all `show_djs` deactivated, a `dj_leave` for the co-host — and opens exactly one new show whose `primary_dj_id` is the caller.
- `end_time` equals the show's last logged `add_time`, not `now()`.
- `GET /flowsheet` `on_air` and `GET /flowsheet/djs-on-air` **agree** afterward. That divergence is what users actually saw.
- Two concurrent takeovers produce exactly one new show; the loser gets `endShow`'s CAS 400.
- The tubafrenzy sign-off assertion belongs here, against the mirror helper, rather than in the dj-site E2E where tubafrenzy is a mock.

## Also in scope

- `FLOWSHEET_TAKEOVER_ENABLED=true` in the integration service env (`dev_env/docker-compose.yml` CI profile and `.github/workflows/test.yml`'s service env), not just the jest runner's — the suite talks to a separately spawned backend over HTTP.
- Give `tests/utils/flowsheet_util.js`'s `join_show` a **throw on non-2xx**. #2308 gave it the optional extra-fields object but deliberately left it returning unasserted responses: with the flag off the 409 it guards against cannot occur, and the throw would change behavior for ~30 existing call sites. Once the flag is on in that environment, it stops being optional.

## Constraints

- The integration environment should mirror production, where this ships off. Flipping it on there is a decision about test coverage, not about the rollout — say so in the PR.
- The four/nine sites above must be audited individually, not blanket-patched with `{ intent: 'join' }`. Some of them may be asserting something the 409 makes untrue.

## Related

- [#2232](https://github.com/WXYC/Backend-Service/issues/2232) (epic), [#2233](https://github.com/WXYC/Backend-Service/issues/2233), [#2308](https://github.com/WXYC/Backend-Service/pull/2308)

jakebromberg:
Closes #2309

Adds the flag-ON integration half of #2233's acceptance criteria. #2308 shipped the `intent` / `expected_show_id` routing for `POST /flowsheet/join` with unit coverage only; this turns the flag on **in the integration environment** and adds `tests/integration/flowsheet-takeover.spec.js`, which drives the takeover branch end-to-end against real Postgres.

## The flag stays OFF in production

`FLOWSHEET_TAKEOVER_ENABLED=true` is set in exactly two places, both of which describe the **integration test environment**:

- `dev_env/docker-compose.yml`, the `ci` profile's `backend` service
- `.github/workflows/test.yml`, the "Start services" step

It is set on the **backend service's own runtime env**, not the jest runner's — the suite talks to a separately spawned process over HTTP, so the runner's environment is invisible to the code under test. That distinction is the whole reason this needed its own PR.

**Nothing here changes the rollout.** The flag's default is still OFF, production still ships it OFF, and no user-facing behavior changes. This is a decision about test coverage only. `.env.example` and `apps/backend/config/flowsheetTakeover.ts` are untouched, and the dormant default is what every non-CI environment continues to get.

## The issue's site list was wrong — six sites, not four

The issue predicted four breaking call sites and named `metadata.spec.js` as five of them. That list was written by reading the code. This PR's list was produced empirically: turn the flag on, give `join_show` its throw, run the unmodified suite, and record what actually fails.

The result differs from the prediction in both directions.

### `metadata.spec.js` does not reproduce — zero sites

All nine `join_show` calls there pass **unmodified**. Each is a `beforeEach` that joins `getTestDjId()` — which is `global.secondary_dj_id` — paired with an `afterEach` that ends that same DJ's show. The secondary DJ is therefore always opening and closing *its own* show, never landing on a show someone else has open. The intent contract governs exactly one branch (a show is open, and it isn't the caller's), and these calls never reach it. `mirror-http.spec.js:34` is the same `beforeEach`/`afterEach` shape and is likewise untouched.

### Two real sites the issue missed

Neither appears in the issue, and both are genuine:

- `flowsheet.spec.js`'s "Join Show > Properly Formatted Request" — a **raw `POST /flowsheet/join`**, not a `join_show` call, so a grep for the helper misses it
- `mirror-http.spec.js`'s "guest-DJ leave does not sign off" test

### Per-site audit

All six are in the same routing branch, and all six resolved to `{ intent: 'join' }`. That uniform outcome is a result of the audit, not a substitute for it — a blanket patch would have been correct here **by coincidence**, and would have been wrong the moment one site turned out to be asserting something the 409 makes untrue. Each was checked against what its enclosing test actually asserts:

| # | Site | Verdict | Why |
|---|---|---|---|
| 1 | `flowsheet.spec.js:245` — "Join Show > Properly Formatted Request" (raw POST) | `intent: 'join'` | This *is* the co-host join the `describe` block exists to exercise. The secondary joins the primary's genuinely-open show, which is precisely the #2233 branch. **Missed by the issue.** |
| 2 | `flowsheet.spec.js:350` — "Leave Show" `beforeEach` | `intent: 'join'` | Establishes the co-host the block's tests need in order to have a secondary DJ hit `/flowsheet/end`. See the finding below. |
| 3 | `flowsheet.spec.js:1438` — "returns multiple DJs when multiple are on air" | `intent: 'join'` | The test's entire point is two DJs on air simultaneously. A 409 here leaves one DJ and the assertion fails several lines later for an unrelated-looking reason. |
| 4 | `flowsheet.spec.js:1466` — "Retrieve Playlist Object" `beforeEach` | `intent: 'join'` | The test asserts both DJs appear in `show_djs` *and* that a `dj_join` marker exists, so this has to be a real co-host join. |
| 5 | `mirror-http.spec.js:265` — "endShow mirror shape guard on guest-DJ leave (BS#1119)" | `intent: 'join'` | The test contrasts a guest *leave* against a primary *end*; B must actually be a co-host for the distinction to exist. **Missed by the issue.** |
| 6 | `flowsheet-join-after-tubafrenzy-signoff.spec.js:187` — first join | `intent: 'join'` **on the first join only** | See "option (c)" below. |

### The strongest finding: a `beforeEach` that never checked its own setup

`flowsheet.spec.js`'s "Leave Show" `beforeEach` claims to join a second DJ as a co-host. **None of its three tests depend on that co-host existing:**

- "Properly formatted request" — the primary ends the show; the co-host is irrelevant
- "Forbidden when `dj_id` does not match" — primary's token, `dj_id: 1000`, expects 403 before any membership check
- "No Active Show Session" — the primary ends the show first, so `showMemberMiddleware` sees an empty member list and returns 400 whether or not the secondary ever joined

So with the flag on and no throw, that join would 409, the "co-host" would never join, the `beforeEach`'s stated setup would be a lie — and all three tests would still pass green. The suite would have reported success over a silently degraded fixture. **The throw on non-2xx is what surfaced this**; it is the reason that change belongs with this one rather than being deferred.

### `flowsheet-join-after-tubafrenzy-signoff.spec.js`, option (c)

Only the **first** join needed the intent. The two re-joins after it are unaffected, because they hit `isDjAlreadyActiveOnShow`'s no-op-200 guard in `joinShow` — branch (c) — which returns before the intent check is ever reached. Marking them would have implied a contract they never touch.

## `end_time` is the last logged track's `add_time`, not `now()`

`resolveShowEndInstant` derives the close instant from `MAX(add_time)` floored at `start_time`. The test proves this with a **real 1200 ms wait** rather than a hand-set timestamp:

1. Log a track and read back its database-assigned `add_time`.
2. Wait 1200 ms.
3. Take over, and assert `shows.end_time` is **byte-equal** to that track's own `add_time`.
4. Assert `end_time` predates the takeover call by more than 1000 ms.

Step 4 is what makes step 3 load-bearing. Byte-equality alone would not exclude a coincidental `now()` that happened to land on the same value; the measured gap does. The `show_end` marker is asserted at the same derived instant.

## Also covered

- **`on_air` / `djs-on-air` agree afterward.** This is #2232's actual user-visible symptom — the divergence where `on_air` named the departed owner while `djs-on-air` named whoever had shown up. Asserted positively (both name the taker) and negatively (neither still names the primary or the co-host).
- **The old show is closed properly**: `end_time` set, one `show_end` marker, all `show_djs` deactivated, and exactly one `dj_leave` — for the co-host only, since `endShow` skips the primary whose departure the `show_end` marker already records.
- **Exactly one new show opens**, owned by the caller.
- **The tubafrenzy sign-off**, against the real mirror helper rather than the dj-site E2E where tubafrenzy is a mock. Correlated to the *old* show's `legacy_show_id`, which guards the subtle failure #2308 calls out: both mirror taps read the same `res.locals.mirrorData` key, so a mis-wired end tap would sign off the show that just started.
- **Two concurrent takeovers** produce exactly one new show; the loser gets `endShow`'s compare-and-set 400. Driven with two genuinely parallel requests rather than a timing guess.

## Verification

Six suites, clean database, flag on:

```
Test Suites: 6 passed, 6 total
Tests:       111 passed, 111 total
```

`flowsheet-takeover`, `flowsheet-join-after-tubafrenzy-signoff`, `flowsheet`, `metadata`, `djs`, `mirror-http`.

## Related

- #2232 (epic), #2233, #2308

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -574,6 +574,13 @@ jobs:
           # column was previously destructured away, with no gating coverage.
           CATALOG_TRACK_SEARCH_CTA_ENABLED: 'true'
           CATALOG_TRACK_SEARCH_DISCOGS_ENABLED: 'true'
+          # Flowsheet Go-Live Intent (BS#2233/BS#2308): same GHA-vs-compose
+          # parity as the two flags above (dev_env/docker-compose.yml).
+          # Default OFF in prod; CI flips it on so
+          # tests/integration/flowsheet-takeover.spec.js exercises the
+          # routing against real Postgres. Test-coverage decision, not a
+          # rollout.
+          FLOWSHEET_TAKEOVER_ENABLED: 'true'
           # SES has a 200-message/month quota. The auth service (host process
           # here) is the only sendEmail/sendOTPEmail caller; keep real sends
           # off regardless of whether AWS_ACCESS_KEY_ID etc. are ever set on
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -575,11 +575,13 @@ jobs:
           CATALOG_TRACK_SEARCH_CTA_ENABLED: 'true'
           CATALOG_TRACK_SEARCH_DISCOGS_ENABLED: 'true'
           # Flowsheet Go-Live Intent (BS#2233/BS#2308): same GHA-vs-compose
-          # parity as the two flags above (dev_env/docker-compose.yml).
-          # Default OFF in prod; CI flips it on so
-          # tests/integration/flowsheet-takeover.spec.js exercises the
-          # routing against real Postgres. Test-coverage decision, not a
-          # rollout.
+          # parity requirement as the flags above (dev_env/docker-compose.yml),
+          # but NOT the same rollout state — those are ON in production, and
+          # this one ships OFF and stays OFF there. CI flips it on only so
+          # tests/integration/flowsheet-takeover.spec.js exercises the routing
+          # against real Postgres. Test-coverage decision, not a rollout.
+          # Must be on the SERVICE env (this step), not the jest runner's: the
+          # suite drives a separately spawned backend over HTTP.
           FLOWSHEET_TAKEOVER_ENABLED: 'true'
           # SES has a 200-message/month quota. The auth service (host process
           # here) is the only sendEmail/sendOTPEmail caller; keep real sends
--- a/dev_env/docker-compose.yml
+++ b/dev_env/docker-compose.yml
@@ -266,6 +266,13 @@ services:
       # tests/integration/library-search-alias.spec.js exercises the LATERAL
       # against a seeded artist_search_alias variant ('Thee Oh Sees' → OHSEES).
       - CATALOG_SEARCH_ALIAS_ENABLED=${CATALOG_SEARCH_ALIAS_ENABLED:-true}
+      # Flowsheet Go-Live Intent (BS#2233/BS#2308): default OFF in prod: a
+      # caller must say `intent: "join"` or `"takeover"` before joining a show
+      # they don't belong to; an absent intent is a 409. CI flips it on so
+      # tests/integration/flowsheet-takeover.spec.js exercises the routing
+      # against real Postgres. This is a test-coverage decision, not a
+      # rollout — see docs/env-vars.md.
+      - FLOWSHEET_TAKEOVER_ENABLED=true
     ports:
       - '${CI_PORT:-8081}:8080'
 
--- a/dev_env/docker-compose.yml
+++ b/dev_env/docker-compose.yml
@@ -266,13 +266,16 @@ services:
       # tests/integration/library-search-alias.spec.js exercises the LATERAL
       # against a seeded artist_search_alias variant ('Thee Oh Sees' → OHSEES).
       - CATALOG_SEARCH_ALIAS_ENABLED=${CATALOG_SEARCH_ALIAS_ENABLED:-true}
-      # Flowsheet Go-Live Intent (BS#2233/BS#2308): default OFF in prod: a
-      # caller must say `intent: "join"` or `"takeover"` before joining a show
-      # they don't belong to; an absent intent is a 409. CI flips it on so
+      # Flowsheet Go-Live Intent (BS#2233/BS#2308). When ON, a caller must say
+      # `intent: "join"` or `"takeover"` before joining a show they don't
+      # belong to; an absent intent is a 409. CI flips it on so
       # tests/integration/flowsheet-takeover.spec.js exercises the routing
-      # against real Postgres. This is a test-coverage decision, not a
-      # rollout — see docs/env-vars.md.
-      - FLOWSHEET_TAKEOVER_ENABLED=true
+      # against real Postgres. This is a test-coverage decision, NOT a rollout:
+      # the flag ships OFF and stays OFF in production — see docs/env-vars.md.
+      # Overridable (like the three flags above) so a developer can reproduce
+      # the flag-OFF production configuration locally; with it off the takeover
+      # spec fails loudly at `.expect(409)` rather than passing vacuously.
+      - FLOWSHEET_TAKEOVER_ENABLED=${FLOWSHEET_TAKEOVER_ENABLED:-true}
     ports:
       - '${CI_PORT:-8081}:8080'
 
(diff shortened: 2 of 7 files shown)


jakebromberg:
One consequence of this change worth flagging for a reviewer, deliberately **not** fixed here.

Turning the flag on for the whole integration environment means every integration spec now runs flag-ON — so **flag-OFF, the configuration production actually ships, loses its integration coverage.** It keeps only the unit coverage in `tests/unit/controllers/flowsheet.joinIntent.test.ts`, which mocks the entire service layer.

That matters more than usual here because the flag-OFF contract is rollout-critical and documented as such in `docs/env-vars.md`: flag-OFF must never 400 or 409, because `auto-dj-orchestrator` ships `intent: "takeover"` before the flip and its `join()` throws on a response body with no show id. A regression that made the flag-OFF branch 409 on a co-host join, or 400 on an unrecognized `intent`, would pass CI green. The five specs edited here were exactly that regression net, and they were converted rather than duplicated.

I left it alone because closing it properly means either a second backend instance in the integration environment or a runtime config-override endpoint — both materially larger than this PR, and neither is what #2309 asked for. Raising it here rather than filing blind, since it may belong to the #2232 epic rather than a standalone ticket.

Two smaller things intentionally skipped, for the record:

- `mirror-http.spec.js` has a file-scoped `waitForMockRequests` helper that duplicates the sign-off poll loop in the new spec. Extracting it into `tests/utils/mock_api.js` would be a real DRY win, but it changes a shared utility's surface for two callers, and the two loops are tuned differently (50ms vs 100ms interval). Better as its own change than smuggled into a coverage PR.
- Several independent DB reads and the two `on_air`/`djs-on-air` HTTP reads could run under `Promise.all`. Measured upside is a few tens of milliseconds against a 24-second suite, and the sequential form keeps a failure legible — not worth it in test code whose value is legibility.

PR #2332, opened 2026-08-30
fix(flowsheet): floor force-end's ended_at at the show's last logged entry

### LINKED ISSUE #2315: force-end ended_at floor is start_time, so a show can close before entries it contains
## Problem

The `ended_at` override on `POST /flowsheet/shows/{id}/force-end` is bounded to `[start_time, now]`. The upper bound is right. The lower bound lets an operator close a show at an instant that falls **before** entries the show already contains.

Concrete: a show starts 02:00 and has entries logged through 05:00. An operator sends `ended_at: 02:30` — a typo, or a wrong belief about when the DJ actually left. It is accepted. The show now carries `[02:00, 02:30]` while owning rows stamped up to 05:00.

Two public reads then disagree with the row data:

- `getShowsInTimeWindow` (backing `GET /flowsheet/range`, which archive.wxyc.org reads) admits a closed show on `start_time < windowEnd AND end_time > windowStart`. A window covering 03:00–05:00 excludes the show while `GET /flowsheet` still serves its entries from that span.
- Any "which show does this entry belong to" derivation from the time interval now contradicts the `show_id` FK and `show_djs`.

This is the mirror image of the defect the endpoint's own description says the last-entry derivation exists to prevent. That description argues at length against stamping `now()` because it produces an interval overlapping windows the show did not air in; a floor of `start_time` produces an interval that *excludes* windows the show demonstrably did air in.

## End state

`ended_at`'s floor is `max(start_time, <last logged entry's add_time>)` rather than `start_time` — the same instant `resolveShowEndInstant` already computes for the no-override path — so the override can only ever move the close *later* than the derived answer, never earlier. Requests below that floor are a 400 naming the floor.

The `api.yaml` bound is updated in the same change so the contract keeps describing what the server does.

## Alternative worth considering before implementing

If there is a genuine operator need to close a show *before* some of its entries — mis-filed late rows, say — then truncating the show is the wrong remedy and the entries should be re-pointed instead (`jobs/flowsheet-show-split` already does that kind of work). Decide which of the two this is before changing the bound; if it turns out operators do need it, the right answer is a documented, separate affordance rather than a silently-permissive floor.

## Files

- `apps/backend/controllers/flowsheet.controller.ts` — `resolveForceEndInstant`, the `endedAt.getTime() < startedAt.getTime()` check
- `apps/backend/services/flowsheet.service.ts` — `resolveShowEndInstant` (already computes `MAX(add_time)` floored at `start_time`; reuse it rather than recomputing)
- `WXYC/wxyc-shared` `api.yaml` — the `ended_at` field description and the `force-end` 400, both of which currently state `[start_time, now]`

## Constraints

- `endShow(currentShow, endedAt)` treats its instant as one value for `end_time`, the `show_end` marker's `add_time`, its rendered message, and every co-host `dj_leave` marker. Whatever floor is chosen applies to all of them; don't move only the column.
- A show with **no** logged entries has no last-entry instant — the floor degrades to `start_time` there, which is correct and already what `resolveShowEndInstant` returns.
- Contract change means a `wxyc-shared` version bump and a publish, so sequence the two repos accordingly.

## Acceptance criteria

- [ ] `ended_at` below the last logged entry's `add_time` is rejected with a 400 that names the floor
- [ ] A show with no entries still accepts any `ended_at` in `[start_time, now]`
- [ ] `api.yaml`'s `ended_at` description and force-end 400 describe the new floor
- [ ] Integration coverage for: below-floor rejected, exactly-at-floor accepted, above-floor accepted, no-entries show accepted

## Related

Surfaced in review of [WXYC/wxyc-shared#415](https://github.com/WXYC/wxyc-shared/pull/415), the contract half. `ended_at` itself is added by [#2233](https://github.com/WXYC/Backend-Service/issues/2233) — this is a follow-up on the bound, not a blocker for it.

jakebromberg:
## Problem

The `ended_at` override on `POST /flowsheet/shows/:id/force-end` was bounded to `[start_time, now]`. The upper bound is right. The lower one let an operator close a show at an instant that falls **before** entries the show already contains — a show starting 02:00 with entries logged through 05:00 accepted `ended_at: 02:30`, and then carried the interval `[02:00, 02:30]` while owning rows stamped up to 05:00.

Two public reads then disagree with the row data:

- `getShowsInTimeWindow`, backing `GET /flowsheet/range` (which archive.wxyc.org reads), admits a closed show on `start_time < windowEnd AND end_time > windowStart`. A window covering 03:00–05:00 excludes the show while `GET /flowsheet` still serves its entries from that span.
- Any "which show does this entry belong to" derivation from the interval contradicts the entry's own `show_id` and `show_djs`.

This is the mirror image of the defect the endpoint's own description says the last-entry derivation exists to prevent. That description argues at length against stamping `now()` because it produces an interval overlapping windows the show did not air in; a floor of `start_time` produces an interval that *excludes* windows the show demonstrably did air in.

## The alternative, decided

The issue asked to settle first whether operators genuinely need to close a show *before* some of its entries — because if they do, a floor is the wrong remedy.

They do not, and the case that looks like it is the one this endpoint was built for. When an abandoned show hangs, `joinShow` routes every subsequent DJ into it as a silent co-host (the 2026-08-20 shape), so the hung show accumulates rows that aired after its real end. The thing that is wrong there is the *entries' ownership*, not the show's length: truncating the show around them trades one interval lie for another and leaves the rows pointing at a show that claims not to have been running. `jobs/flowsheet-show-split` already re-points them to the show that aired them. So the remedy is split-then-close, and the 400 is what tells an operator that is the situation they are in.

## What changed

`resolveForceEndInstant`'s floor is now `resolveShowEndInstant(show)` — the same `MAX(add_time)`-floored-at-`start_time` instant the no-override path already derives, reused rather than recomputed. An override may therefore only ever move the close *later* than the flowsheet's own answer, never earlier.

A show with nothing logged has no last-entry instant, and there `resolveShowEndInstant` returns `start_time` — so the old bound survives exactly where it was the honest one, which is also the shape of the entire legacy open-show backlog.

The rejection names the floor as an ISO instant, and names *which* of the two it is ("the show's last logged entry" vs "the show's start_time"), because `resolveShowEndInstant` collapses the two cases and "too early" without "earlier than what" leaves the operator bisecting.

The future check moves above the floor query so a mistyped year costs no round trip.

`ended_at` has no other reader anywhere in the repo (`grep` across `apps/`, `jobs/`, `tests/` finds only this resolver and its own tests), so the change is contained to one function.

## Acceptance criteria

- [x] `ended_at` below the last logged entry's `add_time` is rejected with a 400 that names the floor
- [x] A show with no entries still accepts any `ended_at` in `[start_time, now]`
- [x] `api.yaml`'s `ended_at` description and force-end 400 describe the new floor — [WXYC/wxyc-shared#425](https://github.com/WXYC/wxyc-shared/pull/425)
- [x] Integration coverage for: below-floor rejected, exactly-at-floor accepted, above-floor accepted, no-entries show accepted

## Tests

Unit (`tests/unit/controllers/flowsheet.operatorClose.test.ts`) — written first, both new rejection tests red before the change and green after. The mocked `resolveShowEndInstant` returns an instant 46 minutes after the fixture's `start_time`, and that gap is the point: it separates the two floors the endpoint used to conflate. `accepts an instant exactly at start_time` is replaced by the no-entries case, which is the only situation where that is still the right answer.

Integration (`tests/integration/flowsheet-open-shows.spec.js`) — three new fixture shows with a four-day gap between `start_time` and their last logged entry, which is the span the old bound accepted in full. `insertShow` grows an `entriesDaysAgo` option; every pre-existing probe stamps its entries *at* `start_time`, so none of them could tell the old bound from the new one. One show per 2xx case, since each close consumes it. All three are inserted before `current`, which must stay `max(shows.id)` for the `is_current` assertions.

## Verification

Local, all green:

- `npm run typecheck`, `npm run lint` (0 errors), `npm run format:check`
- `npm run test:unit` — 504 suites, 8905 tests
- `npm run ci:testmock` (Docker, isolated compose project and ports) — **115 suites, 1279 tests, 0 failures**

Three local reds along the way were all environmental and are worth recording: five suites failed at `require()` time for missing `jobs/*/dist/*.cjs` until `npm run build` ran in the fresh worktree; `auth-auto-membership` failed until `.env` carried `DEFAULT_ORG_SLUG`; and `library-search-alias` failed until `CATALOG_SEARCH_ALIAS_ENABLED=true`, which `.env.example` ships as `false`. None of them touch this diff.

## Related

Contract half: [WXYC/wxyc-shared#425](https://github.com/WXYC/wxyc-shared/pull/425). No publish is needed to consume it — the change is description-only and no generated type changes shape.

Closes #2315

### DIFF
--- a/apps/backend/app.yaml
+++ b/apps/backend/app.yaml
@@ -1310,11 +1310,35 @@ paths:
           schema:
             type: string
             enum: ['true']
+      requestBody:
+        required: false
+        content:
+          application/json:
+            schema:
+              type: object
+              properties:
+                ended_at:
+                  type: string
+                  format: date-time
+                  description: >
+                    Override the derived end instant. Bounded on both ends, or
+                    the request is a 400 naming the bound it missed: at or
+                    before now, and at or after the derived instant it replaces
+                    (the show's last logged entry, floored at `start_time`). An
+                    override can only ever move the close LATER than the
+                    flowsheet's own answer — closing a show before entries it
+                    still owns would sink its `show_end` marker below them in
+                    the flowsheet's global `add_time` order. A show that logged
+                    nothing has no last entry, and there the floor is
+                    `start_time`.
       responses:
         '200':
           description: The finalized show, with `end_time` set
         '400':
-          description: The id was not a positive integer, or the show is already ended
+          description: >
+            The id was not a positive integer, the show is already ended, or
+            `ended_at` was unparseable, in the future, or earlier than the
+            show's last logged entry.
         '403':
           description: 'Caller lacks the `flowsheet: manage` permission'
         '404':
--- a/apps/backend/controllers/flowsheet.controller.ts
+++ b/apps/backend/controllers/flowsheet.controller.ts
@@ -1178,14 +1178,33 @@ export const getOpenShows: RequestHandler<object, unknown, unknown, { window_hou
  * need an override — a DJ who stopped logging at 9pm and stayed on the air
  * until 11pm leaves a show whose flowsheet cannot say so, and the operator can.
  *
- * Bounded on BOTH ends because each bound protects a public read. Below
- * `start_time` yields `end_time < start_time`, an interval that cannot overlap
- * anything, which hides the show from `GET /flowsheet/range` on the very day it
- * aired (the same hazard `resolveShowEndInstant`'s own floor exists for). Above
- * `now` yields a show claiming to have ended in the future, which sorts a
- * `show_end` marker above every real entry on the public flowsheet. Out of
- * range is a 400, never a silent clamp: an operator who mistypes a date should
- * learn that, not get a different answer than they asked for.
+ * Bounded on BOTH ends because each bound protects a public read. Above `now`
+ * yields a show claiming to have ended in the future, which sorts a `show_end`
+ * marker above every real entry on the public flowsheet.
+ *
+ * The floor is the DERIVED instant itself, not `start_time` (BS#2315): an
+ * override may only ever move the close LATER than the flowsheet's own answer.
+ * `start_time` alone let a show close before entries it still owns — start
+ * 02:00, entries through 05:00, `ended_at: 02:30` accepted — and that interval
+ * is wrong in the mirror image of the way `now()` is. `getShowsInTimeWindow`
+ * admits a closed show on `start_time < windowEnd AND end_time > windowStart`,
+ * so a 03:00-05:00 window drops it while `GET /flowsheet` still serves its
+ * entries from that span, and any "which show does this entry belong to"
+ * derivation from the interval contradicts the `show_id` FK and `show_djs`.
+ * A show with nothing logged has no last-entry instant, and there
+ * `resolveShowEndInstant` returns `start_time` — so the old bound survives
+ * exactly where it was the honest one.
+ *
+ * Closing a show BEFORE some of its entries is therefore not something this
+ * endpoint offers. When the entries are the thing that is wrong — a later DJ's
+ * rows absorbed into an abandoned show, the 2026-08-20 shape — the remedy is to
+ * re-point them (`jobs/flowsheet-show-split`) and then close, not to truncate
+ * the show around them.
+ *
+ * Out of range is a 400, never a silent clamp: an operator who mistypes a date
+ * should learn that, not get a different answer than they asked for. The
+ * rejection names the floor, because "too early" without "earlier than what"
+ * leaves the operator bisecting.
  */
 const resolveForceEndInstant = async (raw: unknown, show: Show): Promise<Date> => {
   if (raw === undefined || raw === null) {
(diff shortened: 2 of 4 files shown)


jakebromberg:
Pushed a correction. Review caught that the change's central justification was factually wrong, and it was wrong in the issue too — I inherited it rather than checking.

**The wrong claim.** The docstring, both test suites, the contract PR and the PR body all said a show closed behind its own entries disappears from a `GET /flowsheet/range` window those entries fall in, while `GET /flowsheet` still serves them, leaving `entries[].show_id` unresolvable. It doesn't. `getShowsInTimeWindow` is a three-arm `or`, and the third arm admits any show an in-window entry references — added in BS#2062 (`b0c8a05f`, 2026-08-09), three weeks before #2315 was filed, precisely so `show_id` stays resolvable when the timestamps alone would drop the show. A window holding the show's entries still returns the show.

**What survives.** Narrower, still real: a window holding *none* of the show's entries but lying inside its true airtime — entries at 02:10 and 04:50, window 03:00–04:00 — where the range read reports the show did not air.

**The argument that should have led.** It was in the same paragraph all along, and it is the *same* defect as the `now()` one rather than a different one. `endShow` stamps the supplied instant on the `show_end` marker's `add_time` and on every co-host `dj_leave` marker, and `getEntriesByPage` orders globally by `add_time DESC, id DESC`. Where `now()` floats the sign-off above every real entry, an instant behind the show's own entries sinks it below them — the public flowsheet renders a show that signs off and then keeps playing.

The bound itself is unchanged; only its justification is. Three other findings addressed in the same commit:

- **Floor/ceiling can cross.** `resolveShowEndInstant` has no upper bound and legacy paths write `add_time` from source values, so a show with a future-stamped entry has an empty legal range and every override 400s naming a future floor. Documented rather than clamped: `Math.min(floor, now)` would re-admit exactly the truncation this bound refuses. The operator's answer is to omit `ended_at`.
- **The "which floor" discriminator was untested**, despite being an acceptance criterion. Now pinned in both directions, and each test is killed by a different mutation — verified by inverting the comparison (kills the last-entry test) and by forcing it true (kills the start_time test).
- **`apps/backend/app.yaml`**, the spec `/api-docs` actually serves, documented no request body for this endpoint at all and a 400 predating `ended_at`. Both now describe the floor.

Contract PR WXYC/wxyc-shared#425 carries the same correction. Re-verified locally: `test:unit` 8907 green, `ci:testmock` 115 suites / 1279 tests / 0 failures, lint 0 errors, format and typecheck clean.

PR #2340, opened 2026-08-31
Gate the dj_name scrub on the rows it would rewrite, not the shows that carry a real name

jakebromberg:
Follow-up to #2327, which gated `--execute` on the count of `shows.legacy_dj_name` values that are roster real names. That count is a proxy. Measuring what it stands for shows it is a bad one.

## The measurement

Production, 2026-08-30, over the 392 shows the current gate flags:

| | |
| --- | ---: |
| in-scope rows in those shows | 14,663 |
| rows the main pass would **skip** as `already_current` | **14,581 (99.4%)** |
| rows the pass would actually **change** | **82** |
| …of which write a roster real name | 82 |
| distinct shows holding those 82 rows | **3** |

Migration 0053 froze `flowsheet.dj_name` from the very `legacy_dj_name` the recompute now returns. For 99.4% of the flagged rows the recompute equals the stored value, so the pass skips them. Re-deriving a value that is already in the row writes nothing and leaks nothing.

## Why this matters beyond a tidier number

The gate was blocking a ~1.8M-row drain over rows the job provably does not touch. Worse, it made the remedy a question the job cannot answer — whether five DJs who went by their real name on air consented to that. Cohort C is out of scope by design and those rows are skips, so the consent question was never a precondition for anything this job does. It was an artifact of the proxy.

Same for the tubafrenzy angle: re-sourcing `legacy_dj_name` from `DJ_HANDLE` is a **no-op** (all 839 such shows already hold their upstream handle verbatim — BS#1393 copied it losslessly and `DJ_HANDLE` itself contains the real name). **Nothing here is gated by the 2026-09-07 turndown.**

## What changed

- **`loadPreflightRows`** loads the candidate shows' rows in `loadMainPage`'s exact join shape.
- **`countHarmfulRecomputes`** runs the real `recomputeDjName` and counts only rows that would CHANGE and whose new value is a roster real name. Reusing the pass's own function is deliberate — a probe that predicts the pass by re-derivation is a second implementation of the decision, and this file's thesis is that two copies of one decision drift.
- **`selectHarmCandidateShows`** reads *both* `shows`-level recompute sources. A `dj_name_override` holding a real name wins `resolveShowDjName` outright and the old gate never looked at it. The one remaining arm — a live `show_start` resolving through `auth_user.dj_name` — can only fire if one DJ's handle is another's legal name, which the pass-time `recomputed_is_roster_real_name` detector still covers.
- **`PREFLIGHT_MAX_CANDIDATE_SHOWS`** (5,000) refuses rather than probes an implausibly large cohort — the mirror image of the empty-index refusal, and it bounds the probe's single query.
- The polluted-show count is still computed, logged, and carried on the summary span. Context now, not a gate.

Unchanged from #2327: the refusal is still scoped to the `main` pass, still has no override env var, and the empty-roster-index refusal still blocks the whole run.

## What now blocks a live run

**Three junk `auth_user.real_name` rows.** That is the whole blocker, and it always was — this gate and the #2327 proxy it replaces are both downstream of the same bad data.

82 rows in 3 shows are counted, but **they are the scrub working correctly, not harm.** Verified by SHA-256 fingerprint against the 2025-10-13 tubafrenzy backup, for all three shows (legacy ids 115609, 134431, 163809):

| | Backend holds | tubafrenzy column |
| --- | --- | --- |
| `flowsheet.dj_name` (stored) | the legal name | `DJ_NAME` |
| `shows.legacy_dj_name` (recompute source) | the on-air handle | `DJ_HANDLE` |

Both columns match on all three shows. The recompute would replace a **legal name** with the DJ's **handle** — Cohort B, the exact remediation this job exists to perform. The probe scores it as harmful only because two junk `real_name` values (one is *two characters*) collide with those DJs' genuine short handles, pulling their shows into the candidate cohort.

Clean the three rows and the count reaches zero on its own: the shows leave the cohort, and the 82 rows get scrubbed correctly on the next run.

Do **not** clear it by teaching `buildPiiNameIndex` a name-shape heuristic — inside a PII gate that trades a visible false block for an invisible false pass.

## Checks

`format:check` clean, `lint` 0 errors, `typecheck` clean (workspaces + the job's own `tsc`), **520 suites / 9050 tests pass**. No values were printed during any of the production measurement — counts, shapes and booleans only.

Related to #2281.


jakebromberg:
Correcting this PR's original reading of the 82 rows, using the tubafrenzy backup archive at `/Volumes/Spark/wxyc-backups` (195 daily dumps, 2025-10-13 → today) as a baseline.

I described them as "3 distinct name-shaped values that match neither a roster real name nor any DJ's handle — plausibly a misattribution as well as a real-name write," and said the cohort should be read row by row. That was wrong, and it was wrong because I was reasoning from Backend alone about a legacy cohort Backend cannot see — the same `auth_user`-keyed blind spot that has now misled this issue four times.

The 2025-10-13 dump settles it by fingerprint, on all three shows: `flowsheet.dj_name` holds tubafrenzy's `DJ_NAME` (the legal name), `shows.legacy_dj_name` holds `DJ_HANDLE` (the handle). The pass would replace the legal name with the handle — Cohort B, exactly what this job is for. The gate is blocking the remediation it exists to enable.

Root cause is unchanged from what the gate already reports: three junk `auth_user.real_name` values, one of them two characters, colliding with legacy DJs' genuine short handles. Fix those rows and the count clears itself.

The PR's substance stands — gating on would-be writes is still right, and it is what narrowed 392 shows to 3 for inspection. Only my reading of what the 3 contained was wrong. Body updated; the README now carries the backup evidence.

No values were printed in this analysis — SHA-256 prefixes, lengths and token counts only.

Issue #2295, opened 2026-08-27
Albums with artwork but no streaming URLs are stamped enriched_match and never re-asked


jakebromberg:
## Problem

An album whose `album_metadata` row carries `artwork_url` but no streaming columns is stamped `metadata_status = 'enriched_match'` without LML ever being called, and without `album_metadata` ever being written. Every flowsheet row that links to that album then serves a terminal status with five null streaming URLs — permanently. Clients read `enriched_match` as "this album has enriched fields" and render five dead streaming buttons.

The proxy read path has the answers the whole time. For `Funkadelic — Soul Mate` (`Hardcore Jollies`, `album_id 39166`, flowsheet row `5312465`), `GET /proxy/metadata/album` returns all five:

```
spotifyUrl:      https://open.spotify.com/search/Funkadelic%20Soul%20Mate
appleMusicUrl:   https://music.apple.com/search?term=Funkadelic%20Soul%20Mate
youtubeMusicUrl: https://music.youtube.com/search?q=Funkadelic%20Soul%20Mate
bandcampUrl:     https://bandcamp.com/search?q=Funkadelic%20Hardcore%20Jollies
soundcloudUrl:   https://soundcloud.com/search?q=Funkadelic%20Soul%20Mate
```

while the V2 flowsheet feed for the same row returns `spotify_url`, `apple_music_url`, `youtube_music_url`, `bandcamp_url`, `soundcloud_url`, `discogs_url`, `release_year`, `artist_bio`, `genres` and `styles` all null, with `metadata_status: "enriched_match"` and `artwork_url` populated.

## Root cause

[`apps/enrichment-worker/precheck.ts:97`](https://github.com/WXYC/Backend-Service/blob/d8306903c90ee477e94cd10fd7e3d2e9821ad705/apps/enrichment-worker/precheck.ts#L97) skips the LML call when:

```ts
or(isNotNull(album_metadata.artwork_url), isNotNull(album_metadata.discogs_url)),
```

`finalizeFromCachedMetadata` then flips the flowsheet row to `enriched_match` and — by design, since the skip means there was nothing new to write — performs **no `album_metadata` write at all**.

[...]

Issue #1192, opened 2026-05-28
Apple Music search-URL fallback masks LML verification; lands on in-app search


jakebromberg:
## Problem

`SearchUrlProvider.getAppleMusicUrl` (added in #1185, commit e035f24) is invoked unconditionally from `metadata.service.fetchMetadata` whenever LML returns `apple_music_url: null`. The intent (per #1185 body) was the "Tragic Magic" case — *artist not in WXYC library at all → LML returns zero results → no artwork → no Apple button*. The implementation doesn't distinguish that case from LML's *strict-verification rejection* case: when iTunes Search returned candidates but they failed the 80/80 artist+track (+album) floor (WXYC/library-metadata-lookup#389 / #390 / #396), LML deliberately returns `None` so iOS doesn't mint a confidently-wrong link. The new BS fallback overrides that decision with `https://music.apple.com/search?term=<artist>%20<track>`, which Apple Music opens as its in-app search page.

The original LML floor was added with the explicit reasoning *"a missing Apple Music link is strictly better than a confidently-wrong one"* (commit https://github.com/WXYC/library-metadata-lookup/commit/42a6c5d). BS's new write-path fallback structurally reverses that decision for every LML rejection.

## Evidence

Surfaced on prod by Jake's iOS now-playing card for **Tamaryn / "Love Fade" / "The Waves"** (album_id 45717) — Apple Music app lands on its in-app search page for `Tamaryn Love Fade` instead of playing the track. The track IS on Apple Music (collection `1489079636`, also `6771783460`).

`album_metadata` snapshot (prod, 2026-05-28, all three rows minted today after the BS#1188 deploy):

| album_id | apple_music_url | updated_at | linked flowsheet rows |
|---|---|---|---|
| 51321 | `music.apple.com/search?term=Harumi%20Fire%20by%20the%20RIver` | 2026-05-28 19:12:35 | 182 |
| 27150 | `music.apple.com/search?term=The%20Replacements%20Theyre%20Blind` | 2026-05-28 18:45:39 | 31 |
| 45717 | `music.apple.com/search?term=Tamaryn%20Love%20Fade` | 2026-05-28 18:33:22 | 31 |

iTunes Search probe for `Tamaryn Love Fade` returns the correct track:
```
trackName=Love Fade artistName=Tamaryn collectionName=The Waves
trackViewUrl=https://music.apple.com/us/album/love-fade/1489079636?i=1489079641&uo=4
```
[...]

PR #2345, opened 2026-08-31
fix(flowsheet): synthesize streaming search URLs on the V2 flowsheet read path

### LINKED ISSUE #2339: V2 flowsheet read path never synthesizes streaming search URLs — same album greys out on the feed but not on the proxy
## Problem

The two album-metadata read surfaces degrade differently when LML has no verified streaming link, and the V2 flowsheet side is the one iOS users actually see.

`GET /proxy/metadata/album` synthesizes all five search URLs at request time when the resolved metadata lacks them ([proxy.controller.ts#L752-L757](https://github.com/WXYC/Backend-Service/blob/c03adcd4/apps/backend/controllers/proxy.controller.ts#L752-L757), via `searchUrlProvider.getAllSearchUrls`). That was the deliberate #1184/#1185 decision: iOS gets a clickable `open.spotify.com/search/…` / `music.apple.com/search?term=…` button instead of a grey one, and #1192's verified-rejection invariant was scoped to the *write* path only — "Synthesizing at request time doesn't poison persisted state."

The V2 flowsheet read projection does no such thing. [`apps/backend/utils/album-metadata-projection.ts`](https://github.com/WXYC/Backend-Service/blob/c03adcd4/apps/backend/utils/album-metadata-projection.ts) is a plain `coalesce(album_metadata.X, flowsheet.X)` per column plus the #1714 host guard — a NULL in both tables reaches the wire as NULL, and the hardwired iOS "Spotify" / "Apple Music" buttons grey out. This projection feeds all three per-play surfaces: `GET /flowsheet`, `/v2/flowsheet`, and the legacy `GET /playlists/recentEntries?v=2` (via `enrichPlaycutMetadata`).

So the same album, same play, renders differently depending on which endpoint the client hit: the detail proxy shows five tappable buttons, the flowsheet feed greys two of them.

## Why this is now the residual defect

#2295's drain (2026-08-30) filled the five streaming columns for the entire 4,442-row frozen cohort. But the drain follows the *write-path* policy from `enrich.ts`: it persists synthesized YouTube Music / Bandcamp / SoundCloud search URLs (derived from our own library text) and only *verified* Spotify links — never a synthesized Spotify or Apple URL (#1192). Post-drain, the remaining grey-button population on the V2 feed is exactly:

- `spotify_url` NULL wherever LML matched the album but Discogs/streaming had no verifiable Spotify link (persisted as `spotify_status = 'unresolved'`, eligible for the #1915 bounded re-ask), plus the 9 permanent `no_match` albums from the drain;
- `apple_music_url` NULL for essentially every album that has never passed through the proxy's fresh-LML branch, because nothing persists Apple links on this path by design.

This is the structural remainder of the original "all the links are greyed out" report behind #2295 — the part the enrichment write path cannot fix, because it is a read-path asymmetry.

## Desired end state

The V2 flowsheet surfaces degrade the same way `/proxy/metadata/album` does: when a streaming URL is absent after the coalesce, the response carries a synthesized search URL rather than NULL, for the same set of platforms the proxy synthesizes (all five, per #1185's "identical degradation behavior regardless of which branch served the request").

Equivalently acceptable if review lands there instead: a deliberate, documented decision that the flowsheet feed should *not* synthesize (e.g. the feed's buttons should reflect verified-only links), recorded in the projection header and the issue — but then the proxy and the feed disagree on #1185's stated principle, and that disagreement should be owned explicitly, not left as drift.

**Narrowed 2026-08-31 (owner decision, recorded in comments):** the fill is gated on the pre-fill coalesce already carrying at least one real streaming URL. Shipped iOS 3.2's inline gate (`PlaycutMetadataService`: `metadataStatus?.isTerminal == true || inline.streaming.hasAny`) skips the live `/proxy/metadata/album` fetch whenever any streaming URL is present, so an unconditional fill would re-open the BS#2103 empty-card bug for metadata-less rows — five search buttons and nothing else, permanently. Gating on existing streaming means the fill only upgrades grey Spotify/Apple buttons on rows that already short-circuit inline (which includes the entire post-#2295-drain cohort, since the drain persisted synthesized YouTube Music/Bandcamp/SoundCloud URLs) and can never suppress the proxy fallback. Rows with zero real streaming URLs keep serving NULLs.

## Where

- [`apps/backend/utils/album-metadata-projection.ts`](https://github.com/WXYC/Backend-Service/blob/c03adcd4/apps/backend/utils/album-metadata-projection.ts) — the single projection (BS#2103). Synthesis cannot live in the SQL itself (URL-encoding of artist/title belongs in TS), so the likely seam is post-query serialization in the two consumers:
- `apps/backend/services/flowsheet.service.ts` (`FSEntryFieldsRaw` consumers → `GET /flowsheet`, `/v2/flowsheet`)
- `apps/backend/services/playlist-proxy.service.ts` (`enrichPlaycutMetadata` → `GET /playlists/recentEntries?v=2`)
- `apps/backend/services/metadata/providers/search-urls.provider.ts` + `synthesizeSearchUrls` in `@wxyc/metadata` — the existing synthesis implementations to reuse; do not add a new inline copy (that is the #1257/#1242 disease)
- Inputs: the `/flowsheet` seam has them in hand — `FSEntryFieldsRaw` selects `artist_name` / `album_title` / `track_title`. The legacy seam does **not**: `enrichPlaycutMetadata`'s select carries only the projection + library fields, and the names available at `applyPlaycutMetadata` live on the `GroupedPlaycut` (`artistName` / `releaseTitle` / `songTitle`), which come from the tubafrenzy payload and can drift from the flowsheet columns (mojibake era, name variants). Preferred resolution: add `flowsheet.artist_name` / `flowsheet.album_title` / `flowsheet.track_title` to the `enrichPlaycutMetadata` select and synthesize from those, so both surfaces build their search URLs from identical text.

## Constraints

- **#32 post-launch hardening freeze**: the V2 flowsheet shape and the album-metadata serve path are listed critical surfaces on the [Post-launch service hardening board](https://github.com/orgs/WXYC/projects/32). Check the project's current state before implementing; if the freeze is still in effect this waits or gets an explicit exemption.
- **Wire contract**: no shape change — the fields exist and stay nullable; this only changes NULL → search-URL values. But shipped iOS 3.2 clients decode these fields on the legacy endpoint (BS#2103), so verify the app treats a search URL in `spotifyUrl`/`appleMusicUrl` the same way it does from the proxy today (it does for the proxy; confirm nothing keys "verified" semantics off the flowsheet variant).
- **Do not persist**: synthesis must stay request-time. Writing synthesized Spotify/Apple URLs into `album_metadata` re-opens #1192, and writing any synthesized URL from a read path would fight the #2295 forward-fix gate's semantics.
- **Host guard ordering**: `suppressMislabeledStreamingUrls` (#1714) must run *before* the fallback fills, so a suppressed mislabeled URL degrades to a synthesized search URL exactly as the proxy does ([proxy.controller.ts#L445-L453](https://github.com/WXYC/Backend-Service/blob/c03adcd4/apps/backend/controllers/proxy.controller.ts#L445-L453)).
- Per-request synthesis on the flowsheet list endpoints is O(rows) string building — trivial, but keep it out of the SQL and out of any cached-value writes (the #1089 rule on the proxy cache is a good model if caching ever enters this path).

## Acceptance criteria

- [ ] A V2 flowsheet entry whose album has `spotify_url`/`apple_music_url` NULL after the coalesce — and whose coalesce carries at least one real streaming URL — serves synthesized search URLs, byte-compatible in shape with the proxy's degradation (`open.spotify.com/search/…`, `music.apple.com/search?term=…`); a row with zero real streaming URLs serves NULLs, preserving the shipped-iOS live-proxy fallback
- [ ] Same for the legacy `?v=2` grouped payload via `enrichPlaycutMetadata`
- [ ] A verified persisted URL still wins (fallback only fills absent values)
- [ ] `suppressMislabeledStreamingUrls` runs before the fill; a suppressed URL degrades to a synthesized one
- [ ] No new inline copy of the synthesis logic — reuse `@wxyc/metadata` / `search-urls.provider.ts`
- [ ] No write to `album_metadata` or `flowsheet` from any read path
- [ ] Unit tests on both serialization seams; integration test pinning one entry through `GET /v2/flowsheet` with all-NULL streaming columns
- [ ] Projection header (`album-metadata-projection.ts`) updated to document where synthesis happens and why it is not in the projection itself
- [ ] The fill matches the proxy's degradation checks: falsy/blank-aware (a persisted `''` or whitespace value degrades to a synthesized URL exactly as the proxy's `if (!metadata.spotifyUrl)` branch does, not `??`-nullish), and no synthesis for a blank/whitespace-only `artist_name`
- [ ] Scope, corrected 2026-08-31: the fill covers the two read-GET seams only — `transformToIFSEntry` (which serves both V1 `GET /flowsheet` top-level fields and `/v2/flowsheet`) and `applyPlaycutMetadata` (legacy `?v=2`). The mutation/peek echoes (`projectFlowsheetEntry`) and the SSE `liveFs:update` channel (`pickClientFacingColumns`) do **not** route through `transformToIFSEntry` — the earlier "single producer" claim here was wrong — and deliberately stay unfilled: with the fill gated on existing streaming, the gate-relevant bit (`streaming.hasAny`) never differs between producers for the same row, only button decoration does. Documented in the projection header. dj-site does not consume entry-level streaming URLs (its `StreamingLinks` panel reads the proxy metadata feature only).

## Context

- #2295 — the frozen-cohort issue this is the residue of; its closing comment names this asymmetry as the part the drain structurally cannot reach
- #1184 / #1185 — why the proxy synthesizes Spotify/Apple at request time ("iOS doesn't show greyed buttons")
- #1192 — why synthesized Spotify/Apple URLs are never *persisted*; request-time synthesis explicitly exempted
- #2103 — why the projection is one definition; any fix must keep the derivation unforked across the three surfaces
- #1714 — the host guard that must order before the fallback
- #1257 — open write-path reconcile of `synthesizeSearchUrls` copies; this issue must not add another copy

## Related

- WXYC/Backend-Service#2295
- WXYC/Backend-Service#1257
- WXYC/Backend-Service#1192
- WXYC/Backend-Service#1185
- WXYC/Backend-Service#2103

jakebromberg:
## Summary

`GET /proxy/metadata/album` has synthesized a search-URL fallback for absent streaming links at request time since #1184/#1185, but the V2 flowsheet read path (`GET /flowsheet`, `GET /v2/flowsheet`, and the legacy `GET /playlists/recentEntries?v=2` grouped payload) served a bare `null` instead — the same album, same play, rendered differently depending on which endpoint the client hit.

**The fill is gated**, per the [owner decision on #2339](https://github.com/WXYC/Backend-Service/issues/2339#issuecomment-5484322149): it fires only when the post-host-guard, pre-fill values already carry **at least one streaming URL the client would actually see** — `wireUrl`'s own predicate (absolute http(s), no parser differential), so the gate bit and iOS 3.2's `inline.streaming.hasAny` are the same bit *by construction*. Shipped iOS 3.2's `PlaycutMetadataService` skips the live `/proxy/metadata/album` fetch when `metadataStatus?.isTerminal == true || inline.streaming.hasAny`, so a row that already has a real streaming URL short-circuits inline regardless of what we serve — the fill can only upgrade grey Spotify/Apple buttons there, and can never suppress a proxy fetch. A row with zero real streaming URLs keeps serving NULLs, so the live proxy fallback (which returns full metadata *plus* its own synthesized links) still fires. An unconditional fill would have re-opened #2103's empty-card bug for every metadata-less row, permanently, including the now-playing track.

This still covers the population the ticket exists for: the post-#2295-drain cohort carries synthesized YouTube Music / Bandcamp / SoundCloud URLs on every matched row, so those rows qualify and their grey Spotify/Apple buttons become clickable.

- Adds `fillSynthesizedSearchUrls` to the shared `album-metadata-projection.ts` module, reusing `SearchUrlProvider.getAllSearchUrls` exactly as `proxy.controller.ts` does. The gate lives there, once, so the two seams cannot drift.
- Moves `wireUrl` / `hasWireUrlParserDifferential` out of `playlist-proxy.service.ts` into that same module and imports them back. `wireUrl` is simultaneously the legacy seam's **output filter** and the read path's **presence predicate**; a second copy would let the fill's gate and the wire's key set drift apart. Same collapse-the-copies rule BS#889 applied to `synthesizeSearchUrls`. Only the output filters still differ per seam (`?v=2` re-runs `wireUrl` on the way out; `/flowsheet` emits verbatim).
- A value that fails the wire predicate is absent for the fill too: on a qualifying row it degrades to a synthesized URL rather than surviving to be dropped downstream with no fallback left to run; on a non-qualifying row it normalizes to `null`, so an unusable value can never reach a hardwired iOS button and a blank can never reach the V1 wire as `""`.
- Wires it into both read-GET serialization seams downstream of the projection — `transformToIFSEntry` (`flowsheet.service.ts`, serving both V1 `/flowsheet` top-level fields and `/v2/flowsheet`) and `applyPlaycutMetadata` (`playlist-proxy.service.ts`, legacy `?v=2`) — running after the BS#1714 host guard, so a suppressed mislabeled URL degrades exactly as an outright-missing one does. A row whose *only* streaming URL was host-guard-suppressed counts as zero-streaming and serves NULLs.
- The mutation/peek echoes (`projectFlowsheetEntry`) and the SSE `liveFs:update` channel (`pickClientFacingColumns`) do **not** route through `transformToIFSEntry` and deliberately stay unfilled. With the gate in place this is safe: the only bit 3.2 branches on (`streaming.hasAny`) can no longer differ between producers for the same row — only button decoration does. Documented in the projection header.
- The legacy seam's `enrichPlaycutMetadata` select gains `flowsheet.artist_name`/`album_title`/`track_title` so it synthesizes from the same text `/flowsheet` does, never from the tubafrenzy-mirror-derived `GroupedPlaycut` fields.
- Synthesis is request-time only — never persisted to `album_metadata` or `flowsheet` (#1192).

## Review fixes

The blocking review's findings, all addressed:

1. **Gate the fill at both seams** — done in the shared helper, not two copies.
2. **Control-field regression** — `hasRenderableInlineMetadata` now takes the pre-fill streaming snapshot explicitly instead of reading it back off the mutated payload, so a synthesized search URL can never count as renderable metadata for the `metadataStatus` control field. (With the gate this is belt-and-braces, but it is the reading that is correct by construction.)
3. **Blank-aware degradation** — the fill matches the proxy's falsy `if (!metadata.spotifyUrl)` branch: a persisted `''`/whitespace value degrades to a synthesized URL on a qualifying row rather than surviving a `??`-nullish check and then being dropped by `wireUrl` with no fallback left to run. Blanks normalize to `null` unconditionally, so one can never reach the V1 `/flowsheet` wire as `""`.
4. **Non-blank artist gate** — trim-aware, so a whitespace-only `artist_name` synthesizes nothing instead of five buttons opening `%20%20%20` searches.
5. **Set the five URL fields exactly once** — `applyPlaycutMetadata` computes the fill before writing any of the five keys, restoring byte-identical JSON key order with pre-#2339 output for unfilled rows.
6. **Test pins** — every arm of the gate is pinned at both seams (see below). The false docstring claiming the mutation/peek echoes project off the same `IFSEntry` is corrected in the test and in the projection header.

### Wire golden

Regenerated once, at the end, and reviewed line by line. Against `main`, exactly **two** rows change, and both qualify under the wireable test:

| Row | Wireable pre-fill streaming URL | Gains |
|---|---|---|
| 9005 Eiko Ishibashi & Jim O'Rourke | spotify / youtube / bandcamp / soundcloud | `appleMusicURL` |
| 9009 Chuquimamani-Condori | bandcamp (spotify was host-guard-suppressed) | spotify / apple / youtube / soundcloud |

Every other probe reverts to its exact pre-#2339 shape, with no key-order churn:

- **9012** "BS2103 Unenriched Artist" — no persisted streaming at all.
- **9011** Stereolab — scheme-relative `spotify_url`, bare-domain `bandcamp_url`. This is the row the re-review traced: under a non-blank gate it shipped zero streaming keys pre-PR and four synthesized ones post-PR, flipping `streaming.hasAny` false → true purely by synthesizing. Under the wire predicate it no longer qualifies.
- **9013** Hermanos Gutiérrez — raw tab / LF / space inside all three of `youtube_music_url`, `bandcamp_url`, `soundcloud_url`, i.e. WHATWG-vs-Foundation parser differentials. None is wireable, so the row does not qualify either.

The invariant is now pinned as a named unit test (`emits no streaming key for a row whose only persisted streaming URLs are unwireable`) rather than only as golden bytes, and in the fill-helper suite as a parameterized sweep over scheme-relative / bare-hostname / non-web-scheme / raw-tab / backslash / relative-path shapes. These are shapes the columns permit and no write path validates: `sanitizeLookupStreamingUrls` guards only spotify/apple (BS#1710), while `normalize-lookup.ts`, all three `enrich.ts` branches and `flowsheet-no-match-recheck/writer.ts` persist LML's youtube/bandcamp/soundcloud values verbatim.

`GOLDEN_SHA256` is re-pinned to `d8cc07bd3d720442cf6ea39f71d12ca641e9330e4145ac7ad6737b3708acfbb8`. **Follow-up required in `wxyc-ios-64`**: copy `tests/fixtures/recent-entries-v2-wire-golden.json` to `Shared/Playlist/Tests/PlaylistTests/Fixtures/bs2103-enriched-payload.json` and update the SHA there, per the two-step contract documented in the golden test header.

## Test plan

- [x] Gate pinned at both seams: zero-streaming terminal-but-empty `enriched_no_match` row (no fill, `metadataStatus` withheld, no streaming keys at all); a qualifying row with one verified `youtube_music_url` (Spotify/Apple synthesized, YT untouched); whitespace-only `artist_name`; persisted `''` on a qualifying row; a row whose only streaming URL is host-guard-suppressed; a row whose only streaming URLs are unwireable
- [x] Wire-golden cross-repo contract test regenerated once and re-pinned, every changed row verified against the gate
- [x] Integration tests updated on both arms — `GET /flowsheet` with a partially-enriched album vs a zero-streaming album; `?v=2` free-text and artwork-only rows reverted to their pre-#2339 (unfilled) expectations
- [x] `npm run lint`, `npm run format:check`, `npm run typecheck`, `npm run test:unit` (9,083 tests) all green locally
- [ ] CI

Closes #2339

### DIFF
--- a/apps/backend/services/flowsheet.service.ts
+++ b/apps/backend/services/flowsheet.service.ts
@@ -395,6 +395,13 @@ export type FSEntryRaw = {
  * this is the single producer of every IFSEntry that reaches the `/flowsheet`
  * (top-level fields) and `/v2/flowsheet` (`transformToV2`, nested `metadata`)
  * read paths, so guarding the two hardwired streaming URLs here covers both.
+ *
+ * It is NOT, however, the single producer of every flowsheet-shaped payload:
+ * the mutation/peek echoes (`projectFlowsheetEntry`) and the SSE `liveFs:update`
+ * channel (`pickClientFacingColumns`) project straight off the DB/CDC row and
+ * never come through here. They therefore do not get #2339's search-URL fill —
+ * see the header of `utils/album-metadata-projection.ts` for why that asymmetry
+ * is safe.
  */
 export const transformToIFSEntry = (raw: FSEntryRaw): IFSEntry => {
   // BS#1714 host guard, shared with the legacy recentEntries serializer
@@ -406,9 +413,14 @@ export const transformToIFSEntry = (raw: FSEntryRaw): IFSEntry => {
   // #2339: request-time-only fill for whatever the host guard just left
   // absent, mirroring `GET /proxy/metadata/album`'s degradation so the same
   // album doesn't grey out its streaming buttons on this endpoint alone. Runs
-  // AFTER the host guard so a suppressed mislabeled URL degrades to a
-  // synthesized one, same as an outright-missing one. Never persisted — this
-  // only touches the values below, not `raw` or any DB write.
+  // AFTER the host guard so a suppressed mislabeled URL degrades the same way
+  // an outright-missing one does. GATED: the fill only fires for a row whose
+  // post-guard values already carry at least one real streaming URL — shipped
+  // iOS 3.2 skips its live `/proxy/metadata/album` fetch on
+  // `inline.streaming.hasAny`, so filling a zero-streaming row would suppress
+  // the fallback that is the only thing serving it any metadata at all. See
+  // `fillSynthesizedSearchUrls` for the full rationale. Never persisted —
+  // this only touches the values below, not `raw` or any DB write.
   const { spotify_url, apple_music_url, youtube_music_url, bandcamp_url, soundcloud_url } = fillSynthesizedSearchUrls(
     {
       spotify_url: guardedSpotifyUrl,
--- a/apps/backend/services/flowsheet.service.ts
+++ b/apps/backend/services/flowsheet.service.ts
@@ -28,7 +28,11 @@ import {
   lastLoggedShowEntryOrderBy,
   lastLoggedShowEntryOrderBySql,
 } from '@wxyc/database';
-import { ALBUM_METADATA_PROJECTION, suppressMislabeledStreamingUrls } from '../utils/album-metadata-projection.js';
+import {
+  ALBUM_METADATA_PROJECTION,
+  suppressMislabeledStreamingUrls,
+  fillSynthesizedSearchUrls,
+} from '../utils/album-metadata-projection.js';
 import { getUpcomingShowsMapsCached } from './concerts.service.js';
 import { lookupCriticReviewsByAlbumIds } from './album-metadata-lookup.service.js';
 import { getConfig as getCriticReviewsConfig } from '../config/criticReviews.js';
@@ -397,7 +401,24 @@ export const transformToIFSEntry = (raw: FSEntryRaw): IFSEntry => {
   // (BS#2103) — see `suppressMislabeledStreamingUrls`. Applied once and
   // reused for both the top-level field and the nested `metadata` object
   // below.
-  const { spotify_url, apple_music_url } = suppressMislabeledStreamingUrls(raw);
+  const { spotify_url: guardedSpotifyUrl, apple_music_url: guardedAppleMusicUrl } =
+    suppressMislabeledStreamingUrls(raw);
+  // #2339: request-time-only fill for whatever the host guard just left
+  // absent, mirroring `GET /proxy/metadata/album`'s degradation so the same
+  // album doesn't grey out its streaming buttons on this endpoint alone. Runs
+  // AFTER the host guard so a suppressed mislabeled URL degrades to a
+  // synthesized one, same as an outright-missing one. Never persisted — this
+  // only touches the values below, not `raw` or any DB write.
+  const { spotify_url, apple_music_url, youtube_music_url, bandcamp_url, soundcloud_url } = fillSynthesizedSearchUrls(
+    {
+      spotify_url: guardedSpotifyUrl,
+      apple_music_url: guardedAppleMusicUrl,
+      youtube_music_url: raw.youtube_music_url,
+      bandcamp_url: raw.bandcamp_url,
+      soundcloud_url: raw.soundcloud_url,
+    },
+    { artist_name: raw.artist_name, album_title: raw.album_title, track_title: raw.track_title }
+  );
   return {
     id: raw.id,
     show_id: raw.show_id,
@@ -429,9 +450,9 @@ export const transformToIFSEntry = (raw: FSEntryRaw): IFSEntry => {
     release_year: raw.release_year,
     spotify_url,
     apple_music_url,
-    youtube_music_url: raw.youtube_music_url,
-    bandcamp_url: raw.bandcamp_url,
-    soundcloud_url: raw.soundcloud_url,
+    youtube_music_url,
+    bandcamp_url,
+    soundcloud_url,
     artist_bio: raw.artist_bio,
     artist_wikipedia_url: raw.artist_wikipedia_url,
     on_streaming: raw.on_streaming ?? null,
@@ -452,9 +473,9 @@ export const transformToIFSEntry = (raw: FSEntryRaw): IFSEntry => {
       release_year: raw.release_year,
       spotify_url,
       apple_music_url,
-      youtube_music_url: raw.youtube_music_url,
-      bandcamp_url: raw.bandcamp_url,
-      soundcloud_url: raw.soundcloud_url,
+      youtube_music_url,
+      bandcamp_url,
+      soundcloud_url,
       artist_bio: raw.artist_bio,
       artist_wikipedia_url: raw.artist_wikipedia_url,
       genres: raw.genres,
--- a/apps/backend/services/playlist-proxy.service.ts
+++ b/apps/backend/services/playlist-proxy.service.ts
@@ -86,6 +86,7 @@ import {
   ALBUM_METADATA_PROJECTION_WITHOUT_ARTWORK,
   suppressMislabeledStreamingUrls,
   fillSynthesizedSearchUrls,
+  type SynthesizableStreamingUrls,
 } from '../utils/album-metadata-projection.js';
 import type { ConcertDTO } from './concerts.service.js';
 import { attachUpcomingShows, attachCriticReviews, getOnAirDJName } from './flowsheet.service.js';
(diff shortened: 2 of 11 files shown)


jakebromberg:
Cross-repo follow-up for the golden re-pin filed: https://github.com/WXYC/wxyc-ios-64/issues/1032 (fixture bytes + `goldenSHA256` refresh per the two-step contract).

PR #2346, opened 2026-08-31
fix(search): emit nextCursor on the first page of /flowsheet/search

### LINKED ISSUE #2344: GET /flowsheet/search never emits nextCursor on the first page, capping the Previous Sets archive at one page of 50
## Problem

`GET /flowsheet/search` never emits `nextCursor` on the first page, so cursor pagination can never start and the Previous Sets archive at `dj.wxyc.org/dashboard/playlists` is **permanently capped at one page of 50 rows**.

The gate is in [`search.service.ts`](https://github.com/WXYC/Backend-Service/blob/main/apps/backend/services/search.service.ts):

```ts
const parsedCursor = cursor !== undefined && sort === 'date' ? parseCursor(cursor) : null;
const useCursor = parsedCursor !== null;
...
const nextCursor =
  useCursor && results.length === limit
    ? encodeCursor(results[results.length - 1].play_date, results[results.length - 1].id)
    : undefined;
```

`nextCursor` is conditioned on `useCursor` — on a cursor having been **passed in** — rather than on there being a next page. The first request of any session sends no cursor, so `useCursor` is false, so no `nextCursor` comes back. dj-site's infinite query then stops dead:

```ts
// lib/features/playlist-search/api.ts
infiniteQueryOptions: {
  initialPageParam: null,
  getNextPageParam: (lastPage) => lastPage.nextCursor ?? undefined,
},
```

`undefined` means "no further pages", `hasMore` goes false, and the scroll handler in `Results.tsx` never calls `loadNextPage`. Page 2 is unreachable through the UI.

### Measured

Against a local stack seeded with 12 track rows, `limit=5`:

| Request | `results` | `nextCursor` |
|---|---:|---|
| `?q=&page=0&limit=5&sort=date&order=desc` | 5 (a **full** page) | **`null`** |
| same + `&cursor=<hand-built from row 5>` | 5 | `'2026-08-30 14:14:00+00_5'` |

So everything downstream of the first page is correct — `parseCursor`, the `(add_time, id)` predicate, the ORDER BY tie-break, and `encodeCursor` all work. The forward chain simply has no first link.

### How it reached a user

Reported in Slack 2026-08-31 by a DJ using the archive to read back sets:

> when i try to sort by date, i can see some of the results but not all. when i look at the sets from today, i can see all of mollie and sulli's, but only part of jackson's because the results cut off then (idk if that is due to the # of results you can view at any given time)

The 50-row wall lands mid-DJ, so it reads as data loss rather than as pagination. It is also **self-contradicting on screen**: `Results.tsx` renders its `{total} results` footer precisely when `!hasMore`, so the user sees 50 rows above a footer claiming thousands.

## Desired end state

A full first page returns a `nextCursor`, and the client pages through the archive by scrolling.

## Suggested approach

Condition `nextCursor` on there being a next page, not on the request having carried a cursor — i.e. drop `useCursor &&` and emit whenever `sort === 'date'` and `results.length === limit`. `sort` still gates it, because the compound `(add_time, id)` predicate only has index support for the date sort (the reason `useCursor` ignores a cursor on other sorts in the first place). Do not widen it to non-date sorts as a drive-by.

Watch the interaction with offset mode: `offset = useCursor ? 0 : page * limit`. dj-site always sends `page: 0`, so today the default listing is offset-0 forever — that is the same bug seen from the other side. Emitting the cursor fixes both without touching the offset path, which other callers may still use.

## Acceptance criteria

- [ ] A full first page with `sort=date` returns a `nextCursor`; a partial final page does not.
- [ ] Non-date sorts still return no `nextCursor`.
- [ ] A regression test walks **at least three** pages from a cold start (no hand-built cursor) and asserts no duplicate and no skipped `id` across the boundaries — the first-page link is exactly what had no coverage.
- [ ] Verified against dj-site: scrolling the Previous Sets list past 50 rows loads more.
- [ ] The `{total} results` footer no longer appears under a truncated list.

## Related

- WXYC/dj-site#1299 — the epic whose default-listing work surfaced this
- WXYC/dj-site#1324 — the other half of what that DJ asked for (talksets / hour markers in the listing)
- WXYC/Backend-Service#2335 — widens the same endpoint; land order does not matter, but both touch `searchFlowsheet`
- WXYC/Backend-Service#1681 — `COUNT_CAP`, which is why `total` can read `10000+` under a 50-row list

jakebromberg:
Closes #2344.

## The bug

`nextCursor` was gated on `useCursor` — on a cursor having been passed **in** — rather than on there being a next page:

```ts
const nextCursor =
  useCursor && results.length === limit
    ? encodeCursor(/* last row */)
    : undefined;
```

The first request of any session carries no cursor, so it never handed back the link to the second. dj-site's `getNextPageParam` read `undefined`, `hasMore` went false, and the Previous Sets archive at `dj.wxyc.org/dashboard/playlists` was permanently capped at one page of 50 rows. Everything downstream of the first page already worked — `parseCursor`, the `(add_time, id)` predicate, `encodeCursor`. The forward chain simply had no first link.

## Why the ORDER BY changed too — the non-obvious half

Widening the emit gate on its own would have been wrong, and this is the substance of the change.

Page 0 and cursor mode compiled **different** `ORDER BY` clauses. Cursor mode ordered `add_time <dir>, id <dir>`, matching the compound cursor predicate. Page 0 ordered by `add_time` alone.

`add_time` is not a total order on this table. The service's own comment says why: batch-imported legacy entries all carry one import timestamp, so tie groups are large and routinely straddle a page boundary. Under the untied clause Postgres is free to return such a group in any order — harmless while the page is only ever addressed by `OFFSET`, but not once the page's last row becomes the next page's cursor. That row is then an *arbitrary* member of its tie group, and the rest of the group is either re-served (duplicate) or stepped over (skipped).

So the fix is two halves that only work together:

1. Apply the `id` tiebreaker whenever cursor pagination can apply (`sort === 'date'`), cursor passed in or not — the first page emits a cursor now, so the first page has to be totally ordered.
2. Emit `nextCursor` off that now-deterministic last row.

Non-date sorts keep the untied clause. They neither emit nor accept a cursor, so a tiebreaker there would buy nothing and only add sort work.

### Plan cost

Unchanged in kind. The single-column `flowsheet_add_time_idx` still drives the scan and Postgres adds an Incremental Sort over each timestamp group — the same plan `getEntriesByPage` (BS#2133) and `fetchRecentRows` (BS#2132) already run for their `desc(add_time), desc(id)` reads. `schema.ts` records that trade-off measurement on the index itself.

## Two footguns preserved deliberately

- **`sort === 'date'` stays in the emit gate.** Cursor mode is date-only by design — `parseCursor` is consulted only for the date sort. A cursor emitted under any other sort would be silently ignored on the way back in and the client would re-fetch page 0 forever.
- **A final page holding exactly `limit` rows emits a cursor** and costs one extra request that returns zero rows. That is correct and standard; the alternative is an over-fetch or an exact count, and the count on this endpoint is deliberately capped (`COUNT_CAP`, BS#1681) precisely because exact counts are what it cannot afford.

The `LIMIT`/`OFFSET`-vs-`LIMIT` split is untouched, so offset-mode callers are unaffected apart from now receiving a cursor they are free to ignore.

## Tests

New unit suite `tests/unit/services/search.service.first-page-cursor.test.ts`:

- A full first page with no cursor emits `nextCursor` (desc and asc); a short page and an empty page do not; a full page under `sort=artist`/`song`/`dj` does not; a full page in offset mode past page 0 does.
- SQL-level assertions on the compiled `ORDER BY`: the `id` tiebreaker is present for date sorts in both directions and with or without an inbound cursor, and absent for the three non-date sorts.
- A cold-start walk over a nine-row fixture whose page boundaries fall **strictly inside** tie groups. The fake executor returns a tie group in the *opposite* order when the compiled `ORDER BY` lacks the tiebreaker — a legal ordering under `add_time <dir>` alone — so the walk fails both ways against the old code: no cursor at all with the original gate, and one duplicated plus one skipped row if the gate is widened without the ordering change.

New integration spec `tests/integration/flowsheet-search-cursor-walk.spec.js` walks the same shape end to end against Postgres in both sort directions, asserting no duplicate and no skipped `id` across three pages from a cold start. The expected sequence is read back from Postgres under the ordering the contract promises rather than transcribed into the test.

One existing unit test (`omits nextCursor when cursor is not provided (offset mode)`) asserted the bug and is inverted.

`docs/playlist-search/README.md` gains a "Where the chain starts" note recording the emit rule and the ordering requirement it implies.

## Verification

Local: `format:check`, `typecheck`, `lint`, `npm run test:unit` (521 suites / 9,061 tests, all green), and the Docker CI-mock integration tier — 111 suites passing including the new spec. The seven reds in that run are environmental, not from this diff: six are `Cannot find module .../dist/*.cjs` because the local flow skips the workspace `npm run build` that CI's Build step runs before the integration tier (all six pass after building), and `auth-auto-membership` needs the default-org bootstrap that only the `--full` profile enables.

Client-side verification against dj-site (scrolling past 50 rows) is the remaining acceptance item and needs this deployed; no dj-site change is required once `nextCursor` flows.


jakebromberg:
## Changes since review

Rebased onto `main` (currently `9cc01d6`) and pushed one follow-up commit, `c0af65b`. CI green: unit, lint-and-typecheck, integration.

### The cursor's precision — and a correction to the premise

The review's headline finding was that the emitted cursor is floored to milliseconds, because `transformRow` runs `.toISOString()` on a JS `Date` while `flowsheet.add_time` is `timestamptz DEFAULT now()` and carries microseconds on every live-inserted row. The consequence is real and worth stating precisely, because it is what the new code guards against: a cursor naming `floor_ms(T)` is strictly less than the boundary row's `T`, while `parseCursor` binds it back at full `::timestamptz` precision. Ascending, `(T, id) > (floor_ms(T), cursor_id)` then holds for the boundary row **itself** and every page re-serves its predecessor's last row. Descending, the boundary row is correctly excluded but anything inside the open interval `(floor_ms(T), T)` fails `< floor_ms(T)` and is stepped over.

**That is not what production does today, and the reason is worth having on the record.** drizzle-orm's postgres-js driver installs a transparent parser over OID 1184 — `postgres-js/driver.ts`, `transparentParser`, applied to 1184/1082/1083/1114/1182/1185/1115/1231 at `drizzle()` construction — so that its own column mappers can do the converting. A **raw** `db.execute` never receives a `Date` at all. Verified against the CI database on this branch's code:

```
{"id":7118,"isDate":false,"val":"2019-04-01 12:00:00.123456+00"}
```

So `play_date` was already reaching clients as Postgres's own text rendering, at full microsecond precision, and the cursor inherited it. Before the fix, the two microsecond walk specs added here **passed**; only the format assertion failed. Full pre-fix run against a backend image built at `ffcaf69`:

```
✓ walks every page in desc order with no duplicate and no skipped id
✓ walks every page in asc order with no duplicate and no skipped id
✕ the cursor names the boundary row at full precision, not its floored play_date
    Expected pattern: /\.\d{6}Z$/
    Received string:  "2019-03-02 12:00:03.0005+00"
✓ walks microsecond-precision rows in desc order with no duplicate and no skipped id
✓ walks microsecond-precision rows in asc order with no duplicate and no skipped id
```

I still made the change, for three reasons that stand without the bug:

1. **The correctness rides on a dependency's private decision.** Nothing in this repo asks for the transparent parser, no test below the integration tier can observe it, and `SearchResultRow.play_date` was typed `Date` — which is what every unit mock supplies, so the whole unit suite asserts the shape that *would* be broken. A drizzle upgrade, or a rewrite of this query onto the typed builder, silently reintroduces the defect with no failing test.
2. **The documented format was wrong.** `docs/playlist-search/README.md` says the encoding is `${ISO_timestamp}_${id}`. `2019-03-01 12:00:00+00` is not an ISO-8601 timestamp; it round-trips only because V8's `Date.parse` is lenient and Postgres accepts its own output. The token is now `2019-03-01T12:00:00.000000Z`.
3. It is one selected expression.

`CURSOR_TIME_EXPR` selects `to_char(add_time AT TIME ZONE 'UTC', 'YYYY-MM-DD"T"HH24:MI:SS.US"Z"')` as `cursor_time` alongside `play_date`, and the emit reads that. `play_date` is unchanged — it is the `SearchResult` response contract — and `cursor_time` never reaches a returned object (pinned by a test). `parseCursor` needed no change. `SearchResultRow.play_date` is corrected to `Date | string` with the mechanism written down at the type.

### Three simplifications

**Boolean trio collapsed.** `cursorEligible` is hoisted above the two things that depend on it, `sort === 'date'` is written once, and `useCursor` is gone in favour of `parsedCursor !== null` at both call sites. `offset` keeps its exact prior value including in the count-failure fallback, which `search.service.count-cap.test.ts:146` pins.

**The cost comment named the wrong index.** It said `flowsheet_add_time_idx`; the index this query rides is the partial `flowsheet_track_add_time_idx` (migration 0050), whose `WHERE entry_type = 'track'` predicate is this query's own — 0050's header says as much. 0144's `flowsheet_add_time_idx` is the unpartitioned ASC index built for `GET /flowsheet/range`. The cited measurements (BS#2132, BS#2133) are DESC-only, unfiltered and small-group, so they evidence the plan **shape**, not this query, and the comment now says that. `docs/playlist-search/README.md` step 2 claimed this index avoids the in-memory sort, which the added tiebreaker partly undoes; that section now carries the amendment and names what the remaining Incremental Sort is over.

**The hand-rolled fake Postgres is deleted** — the fixture executor, `orderRows`, and the two cold-start walk tests, ~112 lines. Its hostile branch was unreachable in a passing run, it keyed on the database mock representing `flowsheet.id` as the literal string `'id'`, and the `ORDER BY` assertions above it already pin the same defect at its cause. In its place: a compiled-predicate assertion in both directions, which closes a real gap — nothing in `tests/unit/` asserted that the compound `WHERE` predicate compiles at all — plus assertions that the data query selects the microsecond expression and that the cursor is built from it rather than from `play_date`. Both fail against the pre-fix service.

Also folded in: the miscounted "Four pages of three" comment (removed with the walk it described), the controller's stale "cursor mode + full page" note, the two `new Set(seen).size` assertions that could not fail, `cursor` added to the degraded-count Sentry `extra` (in cursor mode `page` is always 0, so a count failure mid-walk was unlocatable), and the duplicated emit-gate test — the inverted one in `search.service.test.ts` is kept.

### Verification

`npm run build`, `format:check`, `typecheck`, `lint`, `npm run test:unit` (523 suites / 9,113 tests, all green), and the Docker CI-mock integration tier: 117 suites passing, 1,309 tests. One red, `auth-auto-membership`, which needs the default-org bootstrap only the `--full` profile enables — unrelated to this diff and the same one noted in the PR description. The six `dist/*.cjs` module-resolution reds from that earlier note are gone; they were the missing workspace build, which CI runs and the local flow does not.

### One loose thread, not fixed here

`GET /flowsheet/search` serves `play_date` as `"2019-03-01 12:00:00+00"` — Postgres text, not the ISO-8601 string the `SearchResult` contract and `transformRow`'s `instanceof Date` branch both imply. Clients cope because `new Date(...)` parses it, and changing it is a response-shape change with its own blast radius, so it stays out of this PR. Worth its own ticket.

Issue #2365, opened 2026-09-05
Epic: passcode-gated DJ self-signup for holiday breaks (server)


jakebromberg:
## Why

During holiday breaks WXYC suspends the regular schedule and fills shows by signup. An inactive DJ who walks into the control room with no account cannot log their show at all — account creation is admin-only end to end, and no manager is on site.

This epic adds a **passcode-gated self-signup path**: a short code on a sticky note in the control room, which a DJ enters to create their own account at `dj` role. Every account made this way is flagged for manager review.

## The governing constraint: availability, not secrecy

This feature runs during weeks when nobody is watching. **Every control must fail toward "wait a few minutes," never "the control room is locked out until someone drives in."**

That constraint is load-bearing and easy to violate. Three separate mechanisms had to be redesigned during planning because each converted a legitimate user's mistake into an outage:

- a per-IP rate limiter — every DJ shares one IP, the control-room computer, so three fumbled codes would lock the room out for 15 minutes
- an auto-revoking failure counter — a failed attempt matches no passcode row, so the only implementable version revoked *every* active code on a few wrong guesses from anywhere on the internet
- a use cap that counted typos — validation ran after the use was claimed, so 25 fumbled usernames would burn the code

**Check any new counter or limit against this before adding one.**

## Design decisions worth knowing before you start

- **The passcode is encrypted, not hashed** (AES-256-GCM), because it is meant to be read back: a manager reveals the current code and phones it to a stranded DJ rather than rotating, since rotating under the two-active-row cap can invalidate the note the rest of the room is using. The key sits beside `DB_PASSWORD` in the EC2 `.env`, so this protects **dumps and snapshots, not host compromise**.
- **`disableSignUp: true` stays on** for both `emailAndPassword` and `emailOTP`. This is a separate endpoint that happens to create users, not better-auth's signup.
- **Self-signed DJs get `dj` immediately.** `member` has `flowsheet: ['read']` only, so a lower role could not do the thing they walked in to do. The review flag, not a reduced role, is what bounds the risk — and between signup and review, the passcode is sufficient to obtain flowsheet write access. That is the deliberate trade.
- **Review state is three `auth_user` columns with no derived boolean**, and the predicate is exported once and shared between the roster filter and the badge, so the two cannot drift.

## Sub-issues

Ship in dependency order. The first two define the schema and the module surface everything else consumes, so they land sequentially before anything fans out.

## dj-site side

[...]

Issue #2363, opened 2026-09-05
Station signup: jobs/station-signup-attempt-prune for 30-day attempt-log retention


jakebromberg:
## Problem

`station_signup_attempt` grows without bound. Rows are retained 30 days for audit (they answer "who revealed the code?" and "what did the attack look like?"), then need deleting.

## Desired end state

A `jobs/station-signup-attempt-prune` workspace deleting attempt rows older than 30 days, consuming `pruneSignupAttempts` from the lifecycle module.

**This is not one file.** `jobs/*` is an npm workspace (`package.json:67-71`). The nearest shape, `jobs/flowsheet-ghost-row-sweep/`, carries `job.ts`, `package.json`, `tsconfig.json`, `tsup.config.ts`, and a README, plus a root `Dockerfile.flowsheet-ghost-row-sweep` and deploy/cron wiring. Follow it.

## Cron registration

Register in `docs/ops-cron-scheduling.md` under **"Excluded / DB-only (verified — don't try to 'space' these)"** (line 46), beside `metadata-no-match-digest` — **not** in the slot table. That table exists to space LML-hitting crons, and a pure-Postgres prune is not a scheduling constraint.

## Acceptance criteria

- [ ] Workspace with the full `jobs/*` footprint
- [ ] Root Dockerfile + deploy wiring
- [ ] Registered in the DB-only section of `docs/ops-cron-scheduling.md`
- [ ] Deletes only rows older than 30 days; a test pins the boundary

## Context

Plan: `~/Downloads/wxyc-station-passcode-signup-plan.md`.

## Related

Blocked by the lifecycle module. Split from the review job because the failure consequences differ: a failed prune means a larger table, a failed review means a silent queue and a security gap, and those want separate alerting.

## Blocked by

- WXYC/Backend-Service#2359

Issue #2362, opened 2026-09-05
Station signup: manager API for reveal, rotate, revoke, clear-cooldown, status, and approve


jakebromberg:
## Problem

Managers need to operate the passcode and clear the review queue. Without these endpoints the feature is unmanageable: no way to rotate the sticky-note code, no way to recover from a cooldown, no way to mark an account reviewed.

## Desired end state

Endpoints in `apps/auth/`, mounted beside the other admin routes:

- **Reveal** — returns the current plaintext code. Readable storage exists precisely so a manager can read the code to a stranded DJ by phone without rotating (rotating under the two-row cap can invalidate the note the rest of the room is using).
- **Rotate** — generates a new code, returns the plaintext, refuses beyond two active rows.
- **Revoke** — manual kill switch, sets `revoked_reason = 'manual'`.
- **Clear cooldown** — writes a `cooldown_cleared` row that acts as a floor on the failure window. **This is the escape hatch that keeps the cooldown from being a lockout**: a sustained attacker can otherwise hold the endpoint in cooldown with nothing in the product able to clear it. Cheap, same auth gate, and it turns "wait for a manager who is not on site" into one phone call.
- **Status** — active code state (expired / revoked with reason, `last_used_at`, `use_count` / `max_uses`), cooldown state, recent failure counts.
- **Approve account** — sets `self_signup_reviewed_at` / `_reviewed_by`.

**Every reveal writes a `passcode_revealed` row carrying `actor_user_id`.** Readable storage removes the structural guarantee show-once storage had for free — that only whoever rotated ever saw the code — so the audit log is what replaces it, and it makes "who could have seen this?" answerable after a suspected leak.

## Auth gate — state it precisely

Use the admin-flag check the sibling endpoint uses: `session.user.role !== 'admin'` -> 403 (`apps/auth/app.ts:253-256`). `requirePermissions` lives in the backend app, not here.

Be explicit that this is **not** "stationManager only": via `grantsAdminFlag` it also admits stray `admin`/`owner` membership rows. That is the right trade — `/auth/admin/provision-user`, which creates accounts at **any** role including `stationManager`, already sits behind exactly this gate, so passcode rotation is not a wider grant than what it protects. A true stationManager-only gate would need a membership-role read that nothing in `apps/auth` does today; if that is wanted it is its own change with its own test.

[...]

Issue #2360, opened 2026-09-05
Station signup: provisionUser options and the PII lint/registry surfaces the endpoint needs


jakebromberg:
## Problem

Two pieces of groundwork the station-signup endpoint needs, split out so the endpoint PR stays under the 1000-line target and so the PII-adjacent edits land small.

### 1. `provisionUser` gains three options

In `apps/auth/provision-user.ts`, all defaulting to today's behaviour:

- `sendSetupInvite` (default `true`)
- `hasCompletedOnboarding` (default `false`)
- `selfSignupAt` (default `null`)

**`selfSignupAt` is the only writer of the column the entire review surface is built on.** Omit it and the feature ships looking healthy with a permanently empty manager queue. It is set on the user row at creation, inside `provisionUser`'s existing atomicity, not as a follow-up update that could fail after the account is live.

`sendSetupInvite: false` matters for a specific reason: the invite mints a 30-day `reset-password:<token>` (see WXYC/Backend-Service#1969, WXYC/Backend-Service#2306), and sending one to a DJ who just chose their own password would leave a month-long live credential sitting in an inbox for nothing.

Note `provisionUser` already passes `emailVerified: true` to `internalAdapter.createUser` unconditionally, for every caller — **no fourth option is needed** for that.

Add a comment at the signature noting this function now serves two callers at different trust levels, one of which passes a caller-supplied password. That is where a careless future edit does the most damage.

### 2. PII lint and registry surfaces

`wxyc/restricted-real-name` is an **error** across `apps/**/*.ts`, `shared/**/*.ts`, `jobs/**/*.ts` (`eslint.config.mjs:133-141`). The station-signup endpoint destructures `realName` from the request body — exactly the flagged shape — so it will not lint clean without:

- `apps/auth/station-signup.ts` added to `ALLOW_LIST` in `eslint-rules/restricted-real-name.cjs`
- a matching `docs/pii.md` Enforcement entry — the rule's own diagnostic requires both changes together

Separately, **`docs/pii.md` has no IP-address row today.** The attempt log stores `ip_hash` (sha256 truncated to 16 chars, never a raw IP), which makes hashed IPs a stored class. Add the Registry row.

## Constraints

- The allow-list entry names a file that does not exist yet. That is fine and intentional — it lands ahead of the endpoint so the endpoint PR is clean on arrival. Add a brief note in the PR body so a reviewer does not flag it as dead config.
[...]

PR #2366, opened 2026-09-05
Station signup: schema for station_passcode, station_signup_attempt, and review columns

### LINKED ISSUE #2358: Station signup: schema for station_passcode, station_signup_attempt, and the review columns
## Problem

DJs cannot create their own accounts. During holiday breaks WXYC suspends the regular schedule and fills shows by signup; an inactive DJ who walks into the control room with no account cannot log their show at all, because account creation is admin-only end to end and no manager is on site.

The fix is a short passcode on a sticky note in the control room, gating a self-signup path. Accounts made that way are flagged for manager review. This issue lays the schema foundation; everything else in the epic depends on it.

## Desired end state

Two new tables and three new `auth_user` columns exist, and all four new user-facing fields are registered with better-auth so they round-trip.

`station_passcode` — **unprefixed deliberately**. In this schema `auth_` marks better-auth-managed tables (`auth_user`, `auth_session`, `auth_verification`, `auth_jwks`); hand-rolled auth-adjacent tables are unprefixed (`anonymous_devices` in 0024, `user_activity` in 0025). These tables are ours, and an `auth_` prefix would misrepresent ownership.

| column | type | note |
| --- | --- | --- |
| `id` | `varchar(255)` PK | |
| `code_encrypted` | `text NOT NULL` | AES-256-GCM, `iv:tag:ciphertext` base64 in one column |
| `created_at` | `timestamptz NOT NULL DEFAULT now()` | |
| `created_by` | `varchar(255) NULL` -> `auth_user.id` `ON DELETE SET NULL` | |
| `expires_at` | `timestamptz NOT NULL` | default TTL 14 days |
| `revoked_at` | `timestamptz NULL` | manual kill switch |
| `revoked_reason` | `text NULL` | `manual` / `use_cap` |
| `last_used_at` | `timestamptz NULL` | |
| `use_count` | `integer NOT NULL DEFAULT 0` | |
| `max_uses` | `integer NOT NULL DEFAULT 25` | |

`station_signup_attempt`:

| column | type | note |
| --- | --- | --- |
| `id` | `varchar(255)` PK | |
| `attempted_at` | `timestamptz NOT NULL DEFAULT now()` | |
| `ip_hash` | `varchar(16) NULL` | sha256 truncated to 16 hex chars, exactly as `apps/auth/rate-limit-key.ts:30` does it. Never a raw IP |
| `actor_user_id` | `varchar(255) NULL` -> `auth_user.id` `ON DELETE SET NULL` | set only on manager actions |
| `outcome` | `varchar(24) NOT NULL` | `passcode_ok` / `passcode_fail` / `cooldown_refused` / `cooldown_cleared` / `passcode_revealed` |

**Composite index on `(outcome, attempted_at)`** — not `attempted_at` alone. Every cooldown read filters on `outcome`: the in-window failure count, the `MAX(attempted_at) WHERE outcome = 'cooldown_cleared'` floor, and the once-per-window `cooldown_refused` check. An `attempted_at`-only index leaves all three scanning a table that grows *with the attack*. Adding it later is a second migration.

`auth_user` additions:

```
self_signup_at          timestamptz NULL
self_signup_reviewed_at timestamptz NULL
self_signup_reviewed_by varchar(255) NULL  -- -> auth_user.id ON DELETE SET NULL
```

Pending review = `self_signup_at IS NOT NULL AND self_signup_reviewed_at IS NULL`. **Do not add a `pending_review` boolean** — it would duplicate what two columns already answer.

## better-auth registration (the part that silently breaks everything if missed)

Register all three new columns in `user.additionalFields` in `shared/authentication/src/auth.definition.ts`, **each with `input: false`**. Two separate silent failures ride on this:

- **Unregistered, better-auth drops them.** It only round-trips non-core user fields declared there. Undeclared, `selfSignupAt` is discarded on write and absent from the `admin/list-users` payload the roster reads — a working endpoint with a permanently empty review queue.
- **Registered without `input: false`, a DJ can approve themselves.** The BS#2297 comment in that block records exactly this: without the flag, public `POST /update-user` lets any signed-in session rewrite an additional field directly. `self_signup_reviewed_at` / `_reviewed_by` are that shape and the account holder is signed in.

**Also add `input: false` to the existing `hasCompletedOnboarding`** in the same block. It is registered without the flag today, so any signed-in session can already flip it via public `/update-user` — the same hole BS#2297 exists to close. Every current writer goes through `internalAdapter` or a direct `db.update`, so the flag breaks nothing.

## Constraints

- **Never hand-write the migration.** `docs/migrations.md:13`: "Always run `drizzle:generate` to add a migration — never hand-create the journal entry, never hand-edit the SQL or snapshot." Schema.ts first, then `npm run drizzle:generate` (use `--custom` per `docs/migrations.md:158` for DDL the differ cannot produce). Each migration is three artifacts: the `.sql`, a `meta/_journal.json` entry, and `meta/<idx>_snapshot.json`. A hand-written one ships without the snapshot and rots the chain silently (`scripts/validate-migrations.mjs` Check 7 enforces this).
- Journal `idx` is the filename number **+1** from idx 47 onward (`docs/migrations.md:184`), so `0160_...` pairs with journal idx 161.
- Install the `.gitattributes` `merge=journal` driver before starting.
- `0160` was free as of `0159_digital-asset-watermark.sql`; confirm against `main` before generating.
- Every `auth_user` FK is `ON DELETE SET NULL`, and `created_by` must be **nullable** to take it. `migrations/0048_fix-fk-on-delete-set-null.sql` exists because four `auth_user` FKs were created `ON DELETE NO ACTION` when they should not have been.
- Migration filename is kebab-case after the number (`0160_station-signup.sql`) — unbroken since `0097`; `0096_cdc_oversized_fallback.sql` is the last snake_case one.

## Acceptance criteria

- [ ] `shared/database/src/schema.ts` carries both tables and the three `auth_user` columns
- [ ] Migration generated via `drizzle:generate`, with all three artifacts committed
- [ ] Composite `(outcome, attempted_at)` index present
- [ ] All `auth_user` FKs `ON DELETE SET NULL`; `created_by` nullable
- [ ] `additionalFields` registers the three new fields with `input: false`
- [ ] `hasCompletedOnboarding` gains `input: false`
- [ ] A test pins the `additionalFields` registration (so a later edit cannot silently drop it)

## Context

Plan: `~/Downloads/wxyc-station-passcode-signup-plan.md` (to move to `docs/plans/` once merged). Fresh clones at `~/Developer/WXYC-fresh/`.

Related prior work: WXYC/Backend-Service#2297 (the `input: false` precedent and its reasoning), WXYC/Backend-Service#2171 (`auth_user.role` vs `auth_member.role` drift — relevant to the downgrade job later in this epic).

## Related

Blocks every other issue in this epic.

jakebromberg:
## Summary

Schema foundation for station self-signup (the epic's first issue — everything else depends on it): during holiday-break shows, WXYC fills open slots by signup, and a DJ with no account currently cannot log a show at all, since account creation is admin-only end to end. This adds a short-passcode self-signup gate, with created accounts flagged for manager review.

- Adds `station_passcode` (encrypted passcode + TTL/use-cap/revocation bookkeeping) and `station_signup_attempt` (cooldown state + audit trail, composite `(outcome, attempted_at)` index) to `shared/database/src/schema.ts`. Both are unprefixed deliberately — `auth_` marks better-auth-managed tables, and these are ours, matching the `anonymous_devices`/`user_activity` precedent for hand-rolled auth-adjacent tables.
- Adds `self_signup_at` / `self_signup_reviewed_at` / `self_signup_reviewed_by` to `auth_user`. Pending review = `self_signup_at IS NOT NULL AND self_signup_reviewed_at IS NULL` — no separate `pending_review` boolean.
- Every new `auth_user` FK is `ON DELETE SET NULL` on a nullable column.
- Generated via `drizzle-kit generate` as migration `0160_station-signup-schema.sql`, with its journal entry and snapshot. Verified applying cleanly against a fresh `postgres:14` instance in one transaction (matching prod's migrator behavior).
- Registers all three new fields, and the previously-unlocked `hasCompletedOnboarding`, with `input: false` in better-auth's `additionalFields` (`shared/authentication/src/auth.definition.ts`) — otherwise the public `POST /update-user` route would let a signed-in DJ rewrite their own review columns directly, per the BS#2297 precedent.
- Extends `tests/unit/authentication/pii-additional-fields-input.test.ts` to pin the `input: false` flag on the three new fields and on `hasCompletedOnboarding`.

## Test plan

- [x] `npm run test:unit` — 9201 tests pass
- [x] `npm run typecheck` — clean across all workspaces
- [x] `npm run lint` — 0 errors
- [x] `npm run format:check` — clean
- [x] `npm run lint:migrations` — journal validation passes
- [x] `npm run check:auth-tables-doc` — passes (no `auth_*` tables added)
- [x] Full migration chain applied cleanly against a fresh `postgres:14` container in a single transaction

Closes #2358


jakebromberg:
## Review pass — security fixes, an approved schema addition, and a deny-by-default test

Pushed as `b93ea06d`. Migration `0160` was **regenerated** with `drizzle-kit generate` rather than hand-edited (it has never been applied to production, so amending it was free); `.sql`, `meta/_journal.json`, `meta/0160_snapshot.json` and `meta/applied-hashes.json` are mutually consistent and `lint:migrations` reports 158 entries, 0 errors, 0 warnings.

### 1. BLOCKING — `capabilities` was writable by any signed-in session

Verified the whole chain against source before fixing:

- `POST /auth/update-user` (`better-auth/dist/api/routes/update-user.mjs`) declares `use: [sessionMiddleware]` and types its body `z.record(z.string(), z.any())`.
- Its handler calls `parseUserInput(ctx.context.options, rest, 'update')`, and `parseInputData` (`db/schema.mjs:59`) copies through every field whose attributes lack `input: false`.
- `databaseHooks.user.update.before` only vetoes `name` (`deriveOrRejectUserNameOnUpdate`), so nothing downstream stops the write.
- `buildJwtPayload` (`jwt-payload.ts:41`) and `buildOidcUserInfoClaim` (`:73`) then emit the array verbatim on the JWT and OIDC `capabilities` claim — which `docs/authentication.md` records as gating `WXYC/wxyc-crm`'s donor-dedup admin tool (`crm_merge`) and dj-site's `editor`/`webmaster` surfaces.

So: one request from any signed-in DJ, self-granted privilege, emitted on the next token. Now `input: false`.

**Confirmed it does not break the documented operator grant path.** `authClient.admin.updateUser` → the admin plugin's `/admin/update-user`, whose handler ends at `internalAdapter.updateUser(ctx.body.userId, ctx.body.data)` (`plugins/admin/routes.mjs:302`) with no `parseUserInput` call anywhere in the route. Same bypass the `realName`/`djName` writeup already verified. `docs/authentication.md` updated, since its "round-trips unfiltered" tail is no longer accurate for the public route.

**`isAnonymous` — explicit decision: locked.** It turns out better-auth's own anonymous plugin already declares this field `input: false` (`plugins/anonymous/schema.mjs`), and today *that* declaration is what actually holds on the write path — but only by accident of merge order. `getFields()` spreads plugin schemas **after** `options.user.additionalFields`, so the plugin wins for `parseUserInput`; `getAuthTables()` spreads them in the **opposite** order, so our entry wins there. The lock was resting on which of two contradictory orders a given call site happens to use, which is not a property worth betting a bump on. Restating the flag removes the dependency, and no legitimate writer is affected: `/sign-in/anonymous` sets the column through `internalAdapter.createUser`, which never filters input. Reasoning is in the code comment, not just here.

### 2. BLOCKING — `ip_hash` derivation

The old comment claimed "sha256 truncated to 16 hex chars … Never a raw IP". An unsalted digest over IPv4's 2^32 keyspace is exhaustively invertible on a laptop, so that column would have been a raw IP with extra steps.

Now specified as **keyed HMAC-SHA256**, same `varchar(16)`, same 16-hex truncation — **no schema type change, only the documented derivation**. The comment names the key (`STATION_SIGNUP_IP_HMAC_KEY`, server-held alongside `STATION_PASSCODE_KEY`), states it must fail closed rather than fall back to an unkeyed digest, and records what the property actually buys (protects a dump or RDS snapshot; does not survive host compromise — the same boundary #2359's passcode encryption draws) plus what rotation costs (correlation with older rows, acceptable on a 30-day log). Written to be precise enough for #2359 to implement and #2360 to classify.

**Miscitation fixed.** The comment pointed at `apps/auth/rate-limit-key.ts` as precedent. `rateLimitKeyFromRequest` there deliberately returns a **raw** IP, and the only hasher in that file, `hashCredential`, is module-private and hashes session tokens and bearer credentials — high-entropy secrets, where an unkeyed digest is sound and the invertibility problem does not arise. The new comment says explicitly that it is *not* that recipe, and keeps only the one thing genuinely borrowed (the 16-hex truncation width).

### 3. Approved schema addition — `station_signup_attempt.passcode_id`

Nullable `varchar(255)` FK to `station_passcode.id`, `ON DELETE SET NULL`, matching the table's existing FK style. NULL for every outcome that matches no passcode row — `passcode_fail` (a wrong code matches nothing by construction, which is exactly why the cooldown counts attempt rows rather than incrementing a per-passcode counter), `cooldown_refused`, `cooldown_cleared` — and populated for `passcode_ok` and `passcode_revealed`. The column comment says exactly that.

### 4. Test inverted to deny-by-default

`tests/unit/authentication/pii-additional-fields-input.test.ts` was a hand-maintained allowlist of six field names, which structurally cannot catch a field nobody remembers to add — precisely how `hasCompletedOnboarding` and `capabilities` both stayed unlocked (the latter right through the first draft of the PR that closed the former). It now brace-matches the `additionalFields` block, enumerates **every** field in it, and requires `input: false` on each, except a commented writable allowlist — `appSkin` alone, a UI preference with no authority, no PII, and no claim. Adding a new unlocked field fails by default; widening the allowlist is a deliberate, reviewable edit. There is also a stale-entry check on the allowlist and a floor on the parsed-field count so a parser failure cannot pass vacuously.

Anchoring to the block also fixes the second half of the finding: `source.match()` returned the first match *anywhere* in the file, so a braced example inside a comment could satisfy it. Three scanner tests on synthetic sources prove the deny-by-default behaviour, the comment immunity, and that a nested object isn't mistaken for a field. Mutation-checked: removing `input: false` from `capabilities` and adding an unflagged field turns the suite red.

### 5. FK referencing columns indexed

Postgres indexes the **referenced** side of an FK automatically (it's the PK) and the **referencing** side never. All four referencing columns carry `ON DELETE SET NULL`, which makes every parent delete responsible for finding and nulling the children — a sequential scan of the attempt log, i.e. the table that grows *with an attack*. Added `station_signup_attempt_actor_user_id_idx`, `station_signup_attempt_passcode_id_idx`, `station_passcode_created_by_idx`, `auth_user_self_signup_reviewed_by_idx`. Free now; a second migration later.

### 6. New `tests/unit/database/schema.station-signup.test.ts`

Follows the `schema.digital-asset.test.ts` / `schema.anonymous-devices.test.ts` precedent. Pins the composite index's column **order** — `(outcome, attempted_at)` and `(attempted_at, outcome)` are both valid, both apply, and every query returns identical rows, so a "tidy" reorder currently passes everything silently while breaking all three cooldown reads' plans. Also pins `created_by` nullability (the FK action and the nullability are one decision), `ip_hash` width, `passcode_id` nullability, the absence of a `pending_review` boolean, and the unprefixed table names. FK `ON DELETE` actions are deliberately **not** re-asserted — `tests/integration/fk-on-delete-general-guard.spec.js` already enumerates every FK Drizzle declares against live `pg_constraint.confdeltype`, and a second hand-listed copy is the exact shape BS#2239 showed doesn't work.

### 7. `0160` header trimmed

~30 lines duplicated prose that already lives in `schema.ts`, inside a file whose SHA-256 is frozen in `applied-hashes.json` at apply time — so it could never be corrected once `schema.ts` moved on (the #705 wedge). Kept only what is genuinely about the file: the `@no-precondition-needed` annotation and the unprefixed-naming rationale. Per-column reasoning now lives once, at the column.

---

## Deliberately NOT changed — deferred

- **No `CHECK` constraint on `station_signup_attempt.outcome`.** The five values are settled in #2358, but the writer that produces them doesn't exist until #2359, and a `CHECK` locked in now would have to be dropped and recreated the first time the epic adds a sixth outcome. `varchar(24)` plus the module owning the vocabulary is the right split for a schema-only foundation; revisit once the writer is real.
- **No `self_signup_at` stamping invariant.** #2360 makes `provisionUser`'s `selfSignupAt` option the sole writer of the column the whole review surface depends on. Nothing enforces "self-signed-up accounts always get stamped" at this layer, and nothing can until there's a writer — that guarantee belongs with #2360's tests, not with a schema-source guard here.
- **`docs/migrations.md`'s journal-idx-vs-filename table still stops at `0099`.** It also states the invariant as "idx = filename + 1 from 47 onward", which no longer holds at the tail: `0159_*` is idx 159 and this migration is idx 160. Correcting that (and deciding whether the off-by-one was re-synced deliberately) is a docs change with no bearing on this PR, and it would collide with any other in-flight migration PR touching the same file.
- **`ip_hash` remains `varchar(16)`.** Widening it was considered and rejected: 16 hex chars is the right size for a bucketing key, collisions are irrelevant for that use, and a short value limits what a leak yields. The vulnerability was the missing key, not the width.
- **No `docs/pii.md` row for `ip_hash`.** That's #2360's acceptance criterion, and splitting it across two PRs would leave the registry half-written.

## Verification

`npm ci --include=dev` → `npm run lint` (0 errors) → `npm run format:check` (clean) → `npm run typecheck` (clean) → `NODE_ENV=test npm run test:unit` (**525 suites, 9223 tests, all passing**). Plus `npm run lint:migrations` (0 errors, 0 warnings), `npm run check:auth-tables-doc`, `scripts/check-bulk-update-analyze.mjs`, and a full-chain apply of all 158 migrations inside a single transaction against a throwaway `postgres:14` container — matching what `migrate()` actually does, and prod's PG major rather than the dev image's PG18.

jakebromberg:
## Second review pass — 9b0210ce

All four blocking findings are addressed. Local gates: `npm run lint` 0 errors (1066 pre-existing warnings), `npm run format:check` clean, `npm run typecheck` 0 errors, `NODE_ENV=test npm run test:unit` 525 suites / 9234 tests all passing, `npm run lint:migrations` passed (158 entries, 0 warnings). No migration regeneration was needed — every `schema.ts` edit is comment-only, so the 0160 SQL, `_journal.json`, `0160_snapshot.json`, and `applied-hashes.json` are untouched. The known `lru-cache` hoisting artifact did not reproduce in this environment, so those numbers are absolute, not a delta.

### 1. `ip_hash` named the wrong IP source (blocking, fixed)

The comment specified HMAC-SHA256 over "the client IP as the string sent to the socket." Behind this deployment's nginx, `req.socket.remoteAddress` is nginx's own upstream address, identical for every client, so a faithful implementation would have written one constant into every row — the audit column carrying zero information, silently, with no test failing and nothing logged. The client IP is the nginx-set `X-Real-IP` header, never XFF and never the bare socket, exactly as `advanced.ipAddress.ipAddressHeaders: ['x-real-ip']` and `apps/auth/rate-limit-key.ts:9-14` already have it (BS#774, BS#1048).

Two things the old comment left open, both of which made its own "equal IPs hash equal" claim false, are now pinned:

- **Key encoding.** `STATION_SIGNUP_IP_HMAC_KEY` is read as 64 hex characters decoded to 32 raw bytes (`Buffer.from(key, 'hex')`). One encoding, stated, so a hex key and a base64 key cannot both "work" and give one address two hash spaces. Anything that does not decode to exactly 32 bytes is a configuration error.
- **IP canonicalization.** Trim, lowercase, and reduce an IPv4-mapped IPv6 address to its dotted quad, so `::ffff:1.2.3.4` and `1.2.3.4` produce one hash. Hash the canonical string's UTF-8 bytes.

I also **split the failure modes**, which the old "fail closed" phrasing conflated. The *derivation* fails closed — never an unkeyed digest fallback, since a plain sha256 over IPv4's 2^32 keyspace is a raw IP with extra steps. The *request* does not: a missing key, absent header, or value `net.isIP()` rejects writes NULL (which is why the column is nullable) and the signup proceeds. Refusing a walk-in DJ over an audit-only column is precisely the "control room is locked out" outcome #2365 forbids; the misconfiguration belongs at startup, where #2359 should surface it the way `create-auto-dj-user.ts:44-49` surfaces its own missing var.

### 2. The cooldown-grouping contradiction (blocking, fixed)

The sentence added last pass — that the HMAC leaves "the cooldown's per-IP grouping" unaffected — was false and contradicted the epic. A per-IP limiter was considered and **rejected** during planning: every legitimate user shares one IP, the control-room computer, so three fumbled codes would lock the whole room out. The table header was the correct half; it justifies the composite `(outcome, attempted_at)` index on station-wide reads that never mention `ip_hash`, and there is no index on `ip_hash` at all.

The `ip_hash` comment now opens with "AUDIT ONLY — never a cooldown grouping key" and carries a `NOT A GROUPING KEY` paragraph stating that the cooldown is **station-global**, that nothing on the request path reads this column, that the absence of an index on it is deliberate, and that #2359 must not key the cooldown here. The header now says the reads are `STATION-WIDE` and cross-references the column. The true part of the old sentence survives with the right justification: the derivation is stable so equal IPs hash equal, which is what makes the column groupable in a *forensic* query written later.

Both claims are now regression-tested (`states that the cooldown is station-global and ip_hash is not its grouping key`), including the structural half — no `CREATE INDEX ... ("ip_hash")` in the DDL and no `.on(table.ipHash)` in `schema.ts`.

### 3. `returned: false` (blocking, fixed — with one deliberate scoping decision)

`parseUserOutput` (`better-auth/dist/db/schema.mjs:24` → `filterOutputFields`) filters on `returned`, not `input`, so all three review fields were emitted on every `/get-session` body, and `buildJwtPayload` / `buildOidcUserInfoClaim` (`jwt-payload.ts:47,55`) spread the whole user object into the JWT and the OIDC claim.

`selfSignupReviewedBy` now carries `returned: false`. It is the only one of the three that discloses a **third party** — the reviewing manager's `auth_user.id`, handed to the DJ they reviewed and to every relying party (wxyc-crm, wxyc-canary).

`selfSignupAt` and `selfSignupReviewedAt` deliberately do **not** get the flag, and this is the one place I did not follow the review note as written — because the premise behind it does not hold in this better-auth version, and I verified rather than assumed:

> **`returned` is global, not audience-scoped.** The admin plugin's roster route runs the same filter: `plugins/admin/routes.mjs:373` maps every row through `parseUserOutput`. There is no "comes from the admin/roster API instead of the session payload" split available — dj-site/#1353 reads these fields straight off `admin/list-users`, which is that same filtered route.

Flagging either timestamp therefore silently pins #1353's `isPendingManagerReview` (`self_signup_at IS NOT NULL AND self_signup_reviewed_at IS NULL`) to a constant: hide `selfSignupReviewedAt` and every reviewed account reads as pending forever; hide `selfSignupAt` and the queue is permanently empty — the exact "ships looking healthy with an empty manager queue" failure #2360 warns about. Nothing goes red either way. And the security gain is nil: both are timestamps on the account holder's **own** row, naming no one else.

So the trade is one-sided — the manager's id is withheld, the subject's own two timestamps are not — and it is now pinned by tests in both directions (one asserting `returned: false` on `selfSignupReviewedBy`, one asserting its absence on the other two, each carrying the reasoning).

**For #1353's implementer:** `selfSignupReviewedBy` will now arrive **absent** from `admin/list-users`. The acceptance criteria do not need it (the predicate, the chip, and the approve action are all unaffected; the approve *write* goes through the admin route, which never calls `parseUserInput` or `parseUserOutput` on the way in). But if a "reviewed by ⟨name⟩" display is ever wanted, it must come from an admin endpoint of ours — #2362's surface — and not from re-opening this flag.

### 4. The deny-by-default guard was defeatable by a block comment (fixed)

`stripLineComments` removed only `//`, so `/* input: false */` inside an otherwise-unlocked field body satisfied the lock assertion — the guard protecting the exact vulnerability this PR exists to close, reporting green while `capabilities` was genuinely writable.

Replaced with a single left-to-right scanner handling both comment forms and copying string literals verbatim. A regex pair cannot do this safely in **either** order: `auth.definition.ts` carries `https://…` and `/auth/*` inside `//` comments, so a block-comment regex run first opens a "comment" at the `/*` in `//*` and swallows everything to the next `*/` anywhere in the file, and run second it finds `/*` fragments the line stripper has already half-eaten. Block comments collapse to a space rather than vanishing, so a comment wedged mid-identifier (`inp/**/ut: false`) cannot fuse two tokens into a third.

Mutation-verified, all red where they should be and green on restore:

| mutation | result |
| --- | --- |
| new field with no `input: false` | red |
| new field masked by `/* input: false */` | **red** (was green) |
| `returned: false` removed from `selfSignupReviewedBy` | red |
| `inp/**/ut: false` token fusion | red |
| `//` or `/*` inside a string literal | still parsed, still locked |

### 5. Index assertions were DDL-only (fixed)

Migration 0160's SHA-256 is frozen in `applied-hashes.json`, so DDL-only assertions pinned a file nobody can change and left `schema.ts` — the file everyone edits — unguarded. Deleting or renaming any of the three non-composite indexes passed; so did adding an unmigrated column.

Indexes and columns are now asserted on both sides, via small parity helpers (`declaredColumns`/`ddlColumns`, `declaredIndexes`/`ddlIndexes`). `attempted_at` — half the composite index and previously unasserted — is pinned for its `NOT NULL` and its `now()` default on both sides. Mutation-verified: deleting the `actor_user_id` index, renaming the `passcode_id` index, deleting the `station_passcode` `created_by` index, adding an unmigrated `notes` column, and dropping `defaultNow()` each go red.

The FK `ON DELETE` deferral is left exactly as it was — `tests/utils/collect-declared-foreign-keys.js` enumerates every declared FK with no allowlist, so re-asserting from migration text would be a second, weaker copy.

---

## Deferred, deliberately

- **A CHECK constraint on `outcome`, and the missing vocabulary for "valid passcode but expired / revoked / exhausted"** — belongs with #2359, which owns the writer. **This is a real spec gap #2359 must close, not just a nicety.** `passcode_fail` is today the only failure token, so it has to cover both "matched nothing" (NULL `passcode_id`, correctly) and "matched a row that was expired, revoked, or at its use cap" — where a `passcode_id` **is** known and dropping it throws away the most useful fact in the audit trail. The previous comment asserted the wrong-code case held "by construction"; I have replaced that with an explicit `OPEN, for BS#2359` note so nobody reads "passcode_fail implies NULL" as an invariant. The vocabulary is incomplete; the nullability is not. Adding the distinguishing token and any CHECK over the resulting vocabulary belongs with the writer, not with a schema that has no producer yet.
- **`use_count` / `max_uses` CHECK constraints** — #2359, same reason: the claim semantics (`UPDATE ... WHERE use_count < max_uses`) land with the code that performs the claim.
- **The `self_signup_at` stamping invariant** — #2360, which introduces the only writer of that column (`provisionUser`'s `selfSignupAt` option).

## Two corrections to the review, and one thing #2359 must not miss

- **Retention is not unimplemented — it is #2363**, the 30-day `station_signup_attempt` prune job, already filed under this epic and blocked by #2359. The schema comment references it directly.
- **`STATION_SIGNUP_IP_HMAC_KEY` and `STATION_PASSCODE_KEY` are absent from `.github/workflows/set-ec2-env-var.yml`'s `case "$SECRET_NAME"` allowlist (~line 191).** That workflow hard-fails with `Unsupported secret_name` on anything not listed, so **neither key can be set on the EC2 host today** and #2359 cannot enable this in production without editing that workflow in the same PR. Worth stating plainly because it interacts with the fail-closed decision above: if the derivation refused the request on a missing key, the un-allowlisted variable would guarantee the feature never worked in production at all.

jakebromberg:
Addendum to the pass above — a2513230 adds one paragraph to `docs/authentication.md`, immediately after the Capabilities entry the first pass touched. This PR changes the session/JWT/OIDC payload contract, and that file is where that contract is recorded, so leaving the change to live only in a code comment and a PR thread would put the next reader of the canonical doc one step behind the code.

The paragraph states the thing the review found easy to miss: `input` and `returned` are independent gates answering different questions (`parseUserInput` filters on the first, `parseUserOutput` on the second), so a field can be locked against writes and still ride every response body; `input: false` is enforced deny-by-default by the guard test, while `returned: false` stays a per-field decision; `returned` is global rather than audience-scoped, so hiding a field from `/get-session` also hides it from `/admin/list-users` and therefore from dj-site's roster; and `self_signup_reviewed_by` is currently the only field carrying it, with the reasoning for why its two sibling timestamps deliberately do not.

Issue #454, opened 2026-04-25
ETL: Missing --default-character-set=utf8 in MirrorSQL causes Unicode garbling


jakebromberg:
## Summary

Unicode characters in flowsheet and library data are being garbled during ETL imports. Em-dashes (U+2014) and other non-ASCII characters are stored as replacement characters (`�`) in PostgreSQL because the `mysql` CLI client connection defaults to `latin1` instead of `utf8`.

## Real-world example

Flowsheet entry **#5198493** (artist: "Mutat And Ilyas From Ulyap") has `album_title` stored as:

```
Music from the Caucasus � The Archive of ORED Recordings, 2013�2023
```

The correct value should be:

```
Music from the Caucasus — The Archive of ORED Recordings, 2013–2023
```

The em-dashes and en-dashes were replaced with the Unicode replacement character. This breaks metadata lookups because the garbled album title cannot match anything on Discogs.

## Root cause

`MirrorSQL.makeSqlCommand()` in `shared/database/src/legacy/sql.mirror.ts` builds a `mysql` CLI command without specifying `--default-character-set=utf8`. The production database is MySQL 4.1 (Kattare shared hosting). MySQL 4.1's `mysql` CLI defaults to `latin1` for the client connection charset. After the tubafrenzy database was migrated from latin1 to utf8 (migration V002), the data is stored as valid UTF-8 in the database. But when the `mysql` CLI reads it with a `latin1` connection charset, MySQL performs a server-side conversion from `utf8` to `latin1`. Characters outside the latin1 range cannot be represented in latin1, so they get replaced.

For comparison, `discogs-etl/scripts/sync-library.sh` correctly specifies `--default-character-set=utf8` when connecting to the same MySQL database.

## Affected data paths

All three ETL jobs that use `MirrorSQL` are affected:

1. **Library ETL** (`jobs/library-etl/job.ts`) — imports artist names, album titles, and labels
2. **Flowsheet ETL** (`jobs/flowsheet-etl/job.ts`) — imports entry data (artist names, album titles, track titles, labels)
3. **Rotation ETL** (`jobs/rotation-etl/job.ts`) — imports rotation release data

**Not affected:**
- **Tubafrenzy webhook** (`POST /internal/flowsheet-webhook`): Data arrives as JSON from tubafrenzy's Java app using JDBC with `characterEncoding=UTF-8&useUnicode=true`
- **Direct POST via flowsheet controller**: Data comes from dj-site via JSON

[...]

PR #2393, opened 2026-09-08
Capture the 14 CTA ground-truth values from in-database twins (#2152)

### LINKED ISSUE #2152: Repair the 14 U+FFFD-substituted compilation_track_artist values
## Problem

14 values in `wxyc_schema.compilation_track_artist` hold a literal **U+FFFD REPLACEMENT CHARACTER** where tubafrenzy's MySQL holds the real character:

| Table.column | Rows |
|---|---:|
| `wxyc_schema.compilation_track_artist.track_title` | 11 |
| `wxyc_schema.compilation_track_artist.artist_name` | 3 |

Verified at the byte level, not inferred from a terminal — `client_encoding` and `server_encoding` are both UTF8, so this is stored data, not a display artifact. Found by the first end-to-end catalog-parity run against the live pair on 2026-08-13 ([WXYC/discogs-etl#346 step-8a measurement](https://github.com/WXYC/discogs-etl/issues/346#issuecomment-5287728924)).

**CTA has never been swept for this corruption class.** [#863](https://github.com/WXYC/Backend-Service/issues/863)'s table set was `rotation` / `library` / `flowsheet`; [#2114](https://github.com/WXYC/Backend-Service/issues/2114)'s was `library` + `artists`. The Phase 4 script's own informational postlude reads CTA and expected `0` against the 2026-08-12 clone — the live count is not 0, so that expectation was wrong and this is the follow-up it asked for.

## Scope: this is the uncovered half of a 28-value sweep

The 2026-08-13 sweep found 28 U+FFFD values catalog-wide. Half are already covered; this ticket is the other half.

| Values | Covered by |
|---|---|
| `library.artist_name` 10 + `library.album_title` 1 + `artists.artist_name` 1 (id 656) | [#2114](https://github.com/WXYC/Backend-Service/issues/2114) → [PR #2121](https://github.com/WXYC/Backend-Service/pull/2121), merged 2026-08-12. **Merged ≠ applied** — see [#2153](https://github.com/WXYC/Backend-Service/issues/2153) |
| `artists.artist_name` 2 (ids 22025, 23162) + `artists.alphabetical_name` 1 (id 22025) | [#2124](https://github.com/WXYC/Backend-Service/issues/2124) |
| **`compilation_track_artist` 11 + 3 = 14** | **this ticket** |

`rotation` and `flowsheet` need no sweep here — the Phase 4 script already measured them: **flowsheet 0**, **rotation 5** (ids 10789, 13703, 16683, 21149, 21335), and all five are in Phase 3.5's deliberately-unrecovered bucket (empty curated-canonical column in `audit/bs_replacement_char_phase35.csv`). Do not write an acceptance criterion asserting rotation reaches 0; that bucket is intentional.

## Why this is not #1996

[#1996](https://github.com/WXYC/Backend-Service/issues/1996) is **double-encoding** in the same table — the original bytes survive, so it is mechanically reversible, and the damage is upstream in MySQL. This is **substitution**: the source byte is gone (`ef bf bd` is the UTF-8 encoding of U+FFFD itself). No Backend-side transform recovers it, and unlike #1996 the corruption is Backend-only — tubafrenzy holds the true character.

The two overlap on one table and must not fight each other. #1996's fix is overwhelmingly *delete the mangled twin*; this one is *rewrite the substituted character*. Sequencing note under Constraints.

## This is frozen residue, not an active corruption source

The lossy read was fixed on 2026-04-24 by commit `1eab07f5` ("Fix Unicode garbling in MirrorSQL by setting utf8 client charset", [#454](https://github.com/WXYC/Backend-Service/issues/454)) — `MirrorSQL.makeSqlCommand` now passes `--default-character-set=utf8`, so the live ETL reads the true bytes. #2114 independently confirmed the per-row (not per-column) damage pattern: `La Forêt` survives with the same `0xEA` byte that was destroyed in `La Bête`.

Two consequences, both load-bearing:

- **No treadmill, and no blocked-by on WXYC/wiki#89 step 3.** This is the opposite of #1996, which *is* blocked on stopping `library-etl` because its corruption is live upstream. Do not inherit that block by analogy.
- **The repair is durable.** `importCompilationTracks` (`jobs/library-etl/job.ts:765`) inserts CTA rows with `.onConflictDoNothing()` on `cta_unique_idx`. Post-#454 the ETL derives the *correct* string, so once a row is repaired the ETL's next insert conflicts with it and no-ops.

## The trap: `cta_unique_idx` makes a blanket UPDATE unsafe

`cta_unique_idx` is UNIQUE on `(library_id, artist_name, track_title)` (`shared/database/src/schema.ts:731`, and per #801 D7 it is permanent — do not touch the index). If a corrupt row's correctly-spelled twin already exists in the same compilation, `UPDATE`-ing the corrupt row to the true value raises a unique violation instead of repairing anything.

This is not hypothetical for this table: #1996 measured that **98.5% of its damaged CTA rows already have a correctly-encoded twin at the same track position**, because the insert-only writer kept both copies whenever the strings differed. The same double-ingest shape can apply here.

So the repair is **twin-aware**, per row:

- twin exists → `DELETE` the U+FFFD row (the clean row already carries the truth)
- no twin → `UPDATE` in place

Also note `compilation_track_artist.artist_name` is **free text, not an FK to `artists`** — so neither `fold_artist_name` (migration 0134) nor migration 0060's `cascade_library_artist_name` trigger reaches these rows. There is no cascade to lean on the way PR #2121 did for `library.artist_name`; every CTA row must be written directly.

## The corrupt side needs a live prod read — the local clone cannot see it

**`dev_env/seed-clone.sql` contains no `compilation_track_artist` rows.** The table appears exactly once in that file, in the TRUNCATE list at line 39; there is no `COPY wxyc_schema.compilation_track_artist … FROM stdin` block. The populated tables are `artists`, `format`, `genre_artist_crossreference`, `library`, and `rotation`.

That explains the discrepancy in the Phase 4 script's postlude, and it matters for anyone reusing that script's method: its "0 in the 2026-08-12 clone" for CTA was a **false negative measured against an empty table**, not evidence of a clean table. The live count is 14. Do not treat the clone as covering CTA.

Consequences for implementation:

- **Corrupt side (which 14 rows)** — requires a read against prod PG, or another snapshot that actually carries CTA rows. This is the one step that cannot be done from the repo.
- **Ground truth (what the strings should be)** — read Kattare `wxycmusic.COMPILATION_TRACK_ARTIST` directly. `--default-character-set=utf8` is mandatory (without it the server returns latin1 and reintroduces the corruption), and it needs a **MariaDB** client — the Homebrew MySQL 9.x client segfaults against the 5.1 server.

  **A `library.db` snapshot will not serve as ground truth here**, contrary to an earlier revision of this ticket. Every local snapshot checked (`lml-cutover-snapshots/prod-20260719`, `lml-cutover-snapshots/staging-20260718`, `library-metadata-lookup/library.db`, `discogs-etl/library.db`) contains only `library`, `streaming_links`, and the FTS shadow tables — one flat row per release, no per-track or per-compilation-credit data at all. If a *current* daily build turns out to carry CTA rows, it is usable; do not assume it from these.

So the script and its test can be built and verified in full against synthetic fixtures in a throwaway schema; the 14 `(row, true value)` pairs are a separate capture step. **Do not invent row ids or values to fill that gap** — an unresolved pair is a blocker to report, not a guess to make.

Enumeration predicate:

```sql
SELECT cta.library_id, l.legacy_release_id, cta.track_position, cta.artist_name, cta.track_title
FROM wxyc_schema.compilation_track_artist cta
JOIN wxyc_schema.library l ON l.id = cta.library_id
WHERE cta.artist_name LIKE E'%�%' OR cta.track_title LIKE E'%�%'
ORDER BY l.legacy_release_id, cta.track_position;
```

`cta.library_id` is a Backend `library.id`; tubafrenzy keys on `LIBRARY_RELEASE.ID`, so every join to ground truth routes through `library.legacy_release_id`.

## Approach

Follow the Phase 1/2, Phase 3.5, and Phase 4 shape — a hand-applied operator script under `scripts/audit/`, not a Drizzle migration (`docs/migrations.md` keeps migrations DDL-only; this is DML), plus an integration spec that executes the script's real statements against a throwaway schema.

1. Enumerate the 14 rows with their `legacy_release_id` and `track_position`.
2. Resolve each true string from ground truth. **Read it; never reconstruct it** — these are exactly the diacritic-bearing names where a plausible guess is wrong (`Csillagrablók`, `Bête`, `µ-Ziq`). Record the codepoint: PR #2121's `µ` is U+00B5 MICRO SIGN, *not* U+03BC GREEK SMALL LETTER MU.
3. Classify each row twin / no-twin against `cta_unique_idx`.
4. Human-review the 14 pairs before writing the script literals.
5. Script with an audit prelude and postlude, matching `bs_replacement_char_phase4.sql`.
6. `ANALYZE wxyc_schema.compilation_track_artist` outside any transaction ([#934](https://github.com/WXYC/Backend-Service/issues/934) — omitting it after #863's migration regressed `/flowsheet/suggest/*` to 5s timeouts).

## Constraints

- **Capturing the 14 true strings is the only deadline-bound step.** Once they are recorded in the script literals, the rest of this ticket is deadline-free and the `deadline:tubafrenzy-turndown` label can come off. Note the deadline is softer than it reads elsewhere: 2026-09-07 is a self-imposed target, Kattare hosting does not end on that date, and a MySQL dump capture is planned under the data-finalization umbrella. Capture the values early anyway — it costs an hour and removes the question entirely.
- **Sequencing against #1996.** Both rewrite CTA rows. This repair is small, frozen, and unblocked; #1996 is large and blocked on `library-etl` stopping. Land this first, and record the 14 repaired rows so #1996's twin analysis re-derives after, not before.
- **Do not change `cta_unique_idx`** (#801 D7).
- Run a `SELECT` with the same `WHERE` clause before any `UPDATE`/`DELETE`, per the org data-safety convention.
- Per WXYC/wiki#90's Phase 1 freeze, the repair belongs in Backend; do not add a tubafrenzy write path.

## Acceptance criteria

- [ ] The 14 rows enumerated with `legacy_release_id`, `track_position`, and column, recorded in the script's audit prelude.
- [ ] Each replacement traced to its tubafrenzy-sourced string, with the codepoint named.
- [ ] Each row classified twin (`DELETE`) or no-twin (`UPDATE`); no statement can raise a `cta_unique_idx` violation.
- [ ] `ANALYZE wxyc_schema.compilation_track_artist`, outside any transaction.
- [ ] Integration spec executes the script's real statements against a throwaway schema.
- [ ] Post-run: both CTA columns return 0 rows for the U+FFFD predicate.
- [ ] The script is **executed against prod** and its postlude output recorded on this issue before it closes — merging the PR is not the fix (see [#2153](https://github.com/WXYC/Backend-Service/issues/2153)).

## Related

- [WXYC/Backend-Service#2114](https://github.com/WXYC/Backend-Service/issues/2114) — the `library`/`artists` half, repaired by [PR #2121](https://github.com/WXYC/Backend-Service/pull/2121)
- [WXYC/Backend-Service#2153](https://github.com/WXYC/Backend-Service/issues/2153) — that script is merged but not applied
- [WXYC/Backend-Service#2124](https://github.com/WXYC/Backend-Service/issues/2124) — the two remaining `artists` rows
- [WXYC/Backend-Service#1996](https://github.com/WXYC/Backend-Service/issues/1996) — double-encoded CP1252 in the same table; different mechanism, coordinate ordering
- [WXYC/Backend-Service#863](https://github.com/WXYC/Backend-Service/issues/863) — the original U+FFFD recovery, and [#934](https://github.com/WXYC/Backend-Service/issues/934) — the missing-`ANALYZE` regression it caused
- [WXYC/Backend-Service#454](https://github.com/WXYC/Backend-Service/issues/454) — the ETL charset bug that produced this corruption, fixed 2026-04-24
- [WXYC/discogs-etl#346](https://github.com/WXYC/discogs-etl/issues/346) — the parity run where this surfaced

---

_Milestone 1 moved from 2026-08-31 to **2026-09-07** on 2026-08-28 — 24 scheduled DJs still lack a usable Backend account and the classic flowsheet UI is their only surface. Engineering readiness was not the gate. See [WXYC/wiki#125](https://github.com/WXYC/wiki/issues/125)._

jakebromberg:
Fills `pending_cta_repair` with the 14 real rows, turning `scripts/audit/bs_replacement_char_cta.sql` from the verified no-op it shipped as into a live repair. **Already applied to prod** — see the run record on #2152.

## Where the values came from

Not from any of the three channels the script's capture procedure lists. All three read tubafrenzy, and tubafrenzy had already retired when the capture was attempted; channel (b)'s escape hatch never materialised either, since no `library.db` snapshot ever gained per-track data. The procedure is preserved as the record of what was intended, annotated as unrunnable.

The values were still in the table. The script's own "Sequencing vs #1996" note records that post-#454 the ETL derives the correct string, and `importCompilationTracks` inserts via a bare, untargeted `onConflictDoNothing()` — so a correct row does not conflict with a corrupt one, because their `(library_id, artist_name, track_title)` tuples differ, and it lands beside it rather than being suppressed. That is the same 98.5% double-ingest shape #1996 measured, and the same fact this script's twin detection already depends on. What had not been noticed is that it makes the twin a source of *ground truth*, not merely a unique-index hazard to route around.

Release 8844 carries three generations of one credit:

```
110864   Wanda Sá    Só Danço Samba = Jazz 'N' Samba     <- clean original
1228848  Wanda SÃ¡   SÃ³ DanÃ§o Samba = Jazz 'N' Samba    <- #1996 double-encoding
3185320  Wanda SÃ¡   SÃ³ DanÃ��o Samba = Jazz 'N' Samba   <- this ticket's row
```

## Method

For each corrupt row the *undamaged* column is the join key against clean rows in the same release; the *damaged* column is then matched with its U+FFFD runs as bounded wildcards, against each candidate in both stored and re-encoded form. 13 of the 14 resolved to a value present byte-exact as a live row in the same release — located, not reconstructed.

Two decoding subtleties mattered:

- **CP1252's undefined bytes.** The corrupting client passed through the five bytes CP1252 leaves undefined (`0x81`, `0x8D`, `0x8F`, `0x90`, `0x9D`). Python's codec refuses them, which left six values looking irreversibly double-encoded until the codec was corrected.
- **Truncation.** Some twins are cut mid-character at `varchar(255)`, so the reversal must decode tolerantly at the tail only — never mid-string.

**One row needs the closest look at review:** `legacy_release_id` 12988. Its only pattern-matching twin is itself double-encoded *and* truncated, so reversing it drops the final `章`. The captured value instead comes from id 36019, which holds the untruncated string. This is the single row whose value comes from a sibling rather than a direct correspondent.

## Test changes

The spec's placeholder gate named its own succession plan — it asserted the block carried no string literals and said it "goes red the moment real data lands, forcing whoever adds it to add the codepoint assertion in the SAME change." That assertion is here rather than the test being deleted: byte pins for all 14 true values. This matters more than it sounds — release 59194 contains both `·` (U+00B7 MIDDLE DOT) and `・` (U+30FB KATAKANA MIDDLE DOT) in different titles, and an editor normalising one into the other would be invisible in review and would break the downstream byte-exact parity comparison.

The no-op test becomes its inverse: the block declares 14 rows, splits 11 `track_title` / 3 `artist_name` as the header promises, survives the placeholder scrub, and clears every operator-error guard.

The DB-backed tests were not run locally — the integration harness wants a `test-user` role absent from every running container, and the only bootstrap is `npm run db:stop`/`db:start`, where `db:stop` is `down -v` against volumes this change did not create. Left to CI. The static assertions were verified directly against the file: 14/14 byte pins, exactly 14 `VALUES` rows, no dollar-quoting.

## Prod run

Dry-run first, via the script's documented `sed 's/^COMMIT;$/ROLLBACK;/'`, which executes the real guards and the real DELETE/UPDATE inside a real transaction so a would-be `cta_unique_idx` violation surfaces exactly as it would for real. It reported `pending_declared 14 | pending_matched 14 | pending_unmatched 0` — every captured `current_*` matched a live row byte-exactly — then `UPDATE 10 / DELETE 10 / UPDATE 4` and residual 0, and discarded everything.

The real run produced identical counts and committed. Independent verification after the fact: `compilation_track_artist` U+FFFD residual is **0 / 0**; ten of the fourteen original ids are gone, four repaired in place.

10 rows took the DELETE-twin branch and 4 took UPDATE-no-twin, because the twin join requires *both* columns to match post-fix — repairing only the damaged column leaves the other in whatever form it had, so a fully-clean row is not a twin. Those 4 now carry a clean value in the repaired column and a still-double-encoded value in the other. That is #1996's scope, not this ticket's, and it is why the header says to land this first so #1996's twin analysis re-derives after these writes.

Closes #2152

### DIFF
--- a/plans/mu-ziq-replacement-char-remediation.md
+++ b/plans/mu-ziq-replacement-char-remediation.md
@@ -71,7 +71,7 @@ No authoring. Run the existing script per its operator instructions, including t
 
 **Artifact: `scripts/audit/bs_replacement_char_flowsheet.sql`, with a mirrored `tests/integration/bs-replacement-char-flowsheet.spec.js`.** A hand-applied `psql -f` operator script under `scripts/audit/`. **Do not author a Drizzle migration** (r1–r2 proposed this; wrong) — but cite the authority accurately, because `docs/migrations.md`'s `ddl-only` rule (`:62-64`) does not actually prescribe this artifact. It establishes that migrations are DDL-only (so: not a migration, which is the half that holds) and then prescribes a **one-shot backfill job under `jobs/<name>-backfill/`** for bulk DML — scoped to "rewrites of more than ~10k rows," which these repairs are nowhere near. The real authority for the operator-script posture is the predecessor headers: `bs_replacement_char_phase4.sql:3-7` states it verbatim ("Hand-applied operator script, NOT a Drizzle-tracked migration — same posture as … `bs_replacement_char_recovery.sql` … and `bs_replacement_char_phase35.sql`"). Both new scripts should carry that same sentence. The name is load-bearing in three other places below — the spec's `SCRIPT_PATH` constant (the shape at `bs-replacement-char-phase4.spec.js:43`), the CI paths-filter entry, and the acceptance criteria — so it is fixed here rather than left to the implementer. It follows the family convention already set by `recovery` / `phase35` / `phase4` / `cta`.
 
-There is a direct precedent to extend rather than invent: `bs_replacement_char_recovery.sql:110-112` already contains exact-match `flowsheet.artist_name` repairs of precisely this family (`Csillagrablók`, `Sonido Dueñez`, `Eydie Gorme`). Those three lines only — `:113` is `record_label` and `:114-118` are `track_title`. Match its conventions, including **embedding both the corrupt and correct characters raw** rather than as `U&'\FFFD'` escapes — every predecessor does, and `bs-replacement-char-phase4.spec.js:166-176` extracts values by regex from the *script file text*, so a mirrored spec inherits that assumption.
+There is a direct precedent to extend rather than invent: `bs_replacement_char_recovery.sql:110-112` already contains exact-match `flowsheet.artist_name` repairs of precisely this family (`Csillagrablók`, `Sonido Dueñez`, `Eydie Gorme`). Those three lines only — `:113` is `record_label` and `:114-118` are `track_title`. Match its conventions, including **embedding both the corrupt and correct characters raw** rather than as `U&'\FFFD'` escapes — every predecessor does, and `bs-replacement-char-phase4.spec.js:166-176` extracts values by regex from the _script file text_, so a mirrored spec inherits that assumption.
 
 Use the full skeleton all three predecessors share — header block → pre-amble counts → `BEGIN; SET LOCAL statement_timeout = …;` → UPDATEs → `COMMIT;` → post-amble residual verify → `ANALYZE` outside the transaction. Open with the family's `V_BS_FFFD_*` tag, which every sibling carries and which appears in operator-visible output: `bs_replacement_char_phase4.sql:1` is `V_BS_FFFD_P4`, `bs_replacement_char_cta.sql:1` is `V_BS_FFFD_CTA`, and `phase35` reuses its tag in a post-amble section label at `:56`. **Pin `V_BS_FFFD_FS` for §2 and `V_BS_FFFD_RES` for §3** rather than leaving them to the implementer.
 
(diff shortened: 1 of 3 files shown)


jakebromberg:
Note for review: this branch carries a third commit, `70ce0b10`, that reformats `plans/mu-ziq-replacement-char-remediation.md` — 90 lines, entirely unrelated to the capture.

It is pre-existing breakage on `main`, not something this change introduced. That file reached `main` unformatted because `lint-and-typecheck` **skipped** on its own PR (#2385): the change was docs-only, and the workflow's paths-filter routes `src` — not markdown — to the lint job. So `npm run format:check` never saw it, and it now fails on every branch that touches a prettier-checked path.

It is included here only because it blocks CI. Worth its own follow-up: the plan in that very file lists the paths-filter gap as a finding to fix, and this is the gap demonstrating itself on the plan document that documents it.

CI is green with it in: `lint-and-typecheck`, `unit-tests`, and `Integration-Tests` all pass. The three new spec assertions ran against a real Postgres — byte pins for all 14 values, the 14-row / 11-3 split, and the operator-error guards.

